PR #1825, opened 2021-09-19
Withdraw to dank

jplevyak:
Add support for withdrawing cycles during canister delete to a dank account.

Note: we need to set freeze_threshold to zero because the number of cycles we can send is discount by freeze_threshold_cycles() in  rs/cycles_account_manager/src/lib.rs withdraw_with_threshold().  This means that we can't recover as many cycles unless we set freeze_threshold to zero.


jplevyak:
OK, dropping this in favor of the one in this repo proper.

PR #1838, opened 2021-09-27
feat: Add support for withdrawing cycles during canister delete to a dank account.

jplevyak:
Note: we need to set freeze_threshold to zero because the number of cycles we can send is discount by freeze_threshold_cycles() in rs/cycles_account_manager/src/lib.rs withdraw_with_threshold(). This means that we can't recover as many cycles unless we set freeze_threshold to zero.

Note: this was https://github.com/dfinity/sdk/pull/1825


ericswanson-dfinity (approved):

ericswanson-dfinity (changes requested):
Would you please add an entry to CHANGELOG.adoc ?

jplevyak:
Changelog done.  PTAL

ericswanson-dfinity (approved):

PR #1839, opened 2021-09-27
fix: remove extraneous log output wrt stopping icx-proxy

ericswanson-dfinity:
When dfx starts, it logs "Stopping icx-proxy..." "Stopped" when it didn't actually do that.

### DIFF
--- a/src/dfx/src/actors/icx_proxy.rs
+++ b/src/dfx/src/actors/icx_proxy.rs
@@ -101,15 +101,17 @@ impl IcxProxy {
     }
 
     fn stop_icx_proxy(&mut self) {
-        info!(self.logger, "Stopping icx-proxy...");
-        if let Some(sender) = self.stop_sender.take() {
-            let _ = sender.send(());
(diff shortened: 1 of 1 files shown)


krpeacock (approved):

jwiegley:
Would this perhaps be any clearer?
```
    fn stop_icx_proxy(&mut self) {
        if let Some(sender) = self.stop_sender.take() {
            info!(self.logger, "Stopping icx-proxy...");
            sender.send(()).unwrap();
            info!(self.logger, "Stopped.");
        } else if let Some(join) = self.thread_join.take() {
            info!(self.logger, "Stopping icx-proxy...");
            join.join().unwrap();
            info!(self.logger, "Stopped.");
        }
    }
```

ericswanson-dfinity:
No, the normal case is that both are Some, and the stop sequence is: 
1) signal the handler thread to stop (sender.send())
2) wait for handler thread to exit (join.join())

PR #1841, opened 2021-09-30
feat: clarify the default webpack config

nmattia:
This simplifies the logic for finding canister IDs, as well as extends
the logic for arbitrary networks. This also adds some output, clarifying
the heuristics to the user.

The code is also a tad more robust because we never end up in a situation where e.g. both `localCanisters` and `prodCanisters` are undefined (previously the two seemingly exclusive catch-cases occurred together).

NOTE: **I HAVE NOT TESTED THIS!!** 

ok, I _did_ test this, but only for my use case (custom network name).

### DIFF
--- a/src/dfx/assets/new_project_node_files/webpack.config.js
+++ b/src/dfx/assets/new_project_node_files/webpack.config.js
@@ -4,25 +4,29 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
 const TerserPlugin = require("terser-webpack-plugin");
 const CopyPlugin = require("copy-webpack-plugin");
 
-let localCanisters, prodCanisters, canisters;
+let canisters;
 
 function initCanisterIds() {
-  try {
-    localCanisters = require(path.resolve(".dfx", "local", "canister_ids.json"));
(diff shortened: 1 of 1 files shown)


src/dfx/assets/new_project_node_files/webpack.config.js
```diff
@@ -4,25 +4,29 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
...
+    try {
+      return require(path);
+    } catch (error) {
+      console.log(`No canister_ids.json found for network ${network} (${path}), try a different network..`);
+      throw error;
+    }
+  }
 
-  canisters = network === "local" ? localCanisters : prodCanisters;
+  canisters = network === "ic" ?
```
krpeacock, line 27: @nmattia  Don't custom networks end up in the root `canister_ids.json`?

nmattia: not in my case, I'm running `dfx deploy --network foo` and they seem to end up in `.dfx/foo/canister_ids.json`! but this was trial and error

PR #1844, opened 2021-09-30
feat: add support for "rust" canister type

lwshang:
This is the initial work for supporting "rust" canister. `RustCanisterInfo` and `RustBuilder` are implemented.


lwshang:
The removed e2e test could be run locally. But it is invali, because the nix-shell for e2e-tests doesn't include `cargo`. We need to find a proper way to run this kind of e2e tests which run `cargo build --target wasm32-unknown-unknown` during the tests process.

jwiegley (approved):

ericswanson-dfinity (approved):

PR #1871, opened 2021-10-14
fix: Patch interpreter in replica

nmattia:
This makes sure that the system interpreter can be loaded, as opposed to
the one used during the build (which is most likely not present after
the replica was taken out of the nix store).


ericswanson-dfinity:
Closing because master branch now uses the fixed replica and ic-starter binaries.

PR #1926, opened 2021-11-17
feat: check upgrade safety in dfx canister install

chenyan-dfinity:
Newly deployed Motoko canisters now embed the Candid interface and Motoko stable signatures in the Wasm module.
`dfx deploy` and `dfx canister install` will automatically check

	1) the backward compatible of Candid interface in both upgrade and reinstall mode;
	2) the type safety of Motoko stable variable type in upgrade mode to avoid accidentally lossing data;


chenyan-dfinity:
@crusso This branch works if you replace `ic-ref` and `moc` with the latest master in the dfx cache.

src/dfx/src/lib/builders/motoko.rs
```diff
@@ -146,6 +145,19 @@ impl CanisterBuilder for MotokoBuilder {
         };
         motoko_compile(&self.logger, cache.as_ref(), &params)?;
 
+        // Generate stable types
```
crusso, line 148: Do we still need a separate build step for this (and --idl). It should all get produced with the main compilation to binary, if you add the flags.

chenyan-dfinity: Right, I haven't tried the new binary yet. We can fix this when SDK bumps the new moc release.

src/dfx/src/lib/builders/motoko.rs
```diff
@@ -206,16 +178,13 @@ struct MotokoParams<'a> {
...
             BuildTarget::Release => cmd.args(&["-c", "--release"]),
             BuildTarget::Debug => cmd.args(&["-c", "--debug"]),
-            BuildTarget::Idl => cmd.arg("--idl"),
-            BuildTarget::StableTypes => cmd.arg("--stable-types"),
-        };
-        if matches!(self.build_target, BuildTarget::Release | BuildTarget::Debug) {
-            cmd.arg("--public-metadata").arg("candid:service");
         };
+        cmd.arg("--idl").arg("--stable-types");
+        cmd.arg("--public-metadata").arg("candid:service");
```
crusso, line 187: I wonder if we should at least have a setting in dfx.json for that... Making something implicitly public by default seems wrong to me.

src/dfx/src/lib/builders/motoko.rs
```diff
@@ -206,16 +178,13 @@ struct MotokoParams<'a> {
...
         match self.build_target {
             BuildTarget::Release => cmd.args(&["-c", "--release"]),
             BuildTarget::Debug => cmd.args(&["-c", "--debug"]),
-            BuildTarget::Idl => cmd.arg("--idl"),
-            BuildTarget::StableTypes => cmd.arg("--stable-types"),
-        };
-        if matches!(self.build_target, BuildTarget::Release | BuildTarget::Debug) {
-            cmd.arg("--public-metadata").arg("candid:service");
         };
+        cmd.arg("--idl").arg("--stable-types");
```
crusso, line 186: Note that --idl still produce the full idl, including actor class args - will didc check do the sensible thing with that and only look at the actor interface on install, not the class?

src/dfx/src/lib/operations/canister/install_canister.rs
```diff
@@ -62,6 +55,49 @@ pub async fn install_canister(
...
+            std::fs::write(&deployed_path, candid)?;
+            let (mut env, opt_new) = check_candid_file(&candid_path)?;
+            let new_type = opt_new.expect("Generated did file should contain service interface");
+            let (env2, opt_old) = check_candid_file(&deployed_path)?;
+            let old_type = opt_old.expect("Deployed did file should contain service interface");
+            let mut gamma = HashSet::new();
+            let old_type = env.merge_type(env2, old_type);
+            let result = candid::types::subtype::subtype(&mut gamma, &env, &new_type, &old_type);
+            if let Err(err) = result {
+                let msg = format!("Candid interface compatibility check failed for canister '{}'.\nYou are making a BREAKING change. Other canisters or frontend relying on your canister may stop working.\n\n", canister_info.get_name()) + &err.to_string();
```
crusso, line 75:
```suggestion
                let msg = format!("Candid interface compatibility check failed for canister '{}'.\nYou are making a BREAKING change. Other canisters or frontend clients relying on your canister may stop working.\n\n", canister_info.get_name()) + &err.to_string();
```

src/dfx/src/lib/builders/motoko.rs
chenyan-dfinity: Yes, didc will skip the init args.

jzxchiang1:
Will this be merged soon? I noticed that with dfx 0.9.0 `moc` can now generate stable signatures.

chenyan-dfinity:
I hope so. We are stilling waiting for the replica support.

src/dfx/src/lib/operations/canister/install_canister.rs
```diff
@@ -62,6 +55,49 @@ pub async fn install_canister(
...
+        let candid = read_module_metadata(agent, canister_id, "candid:service").await;
+        if let Some(candid) = candid {
+            use crate::util::check_candid_file;
+            let candid_path = canister_info
+                .get_output_idl_path()
+                .expect("Generated did file not found");
+            let deployed_path = candid_path.with_extension("old.did");
+            std::fs::write(&deployed_path, candid)?;
+            let (mut env, opt_new) = check_candid_file(&candid_path)?;
+            let new_type = opt_new.expect("Generated did file should contain service interface");
```
crusso, line 68:
```suggestion
            let new_type = opt_new.expect("Generated did file should contain some service interface");
```

src/dfx/src/lib/operations/canister/install_canister.rs
```diff
@@ -62,6 +55,49 @@ pub async fn install_canister(
...
+            let (env2, opt_old) = check_candid_file(&deployed_path)?;
+            let old_type = opt_old.expect("Deployed did file should contain service interface");
```
crusso, line 70:
```suggestion
            let old_type = opt_old.expect("Deployed did file should contain some service interface");
```

chenyan-dfinity:
@crusso Can you take a look at this test: https://github.com/dfinity/sdk/runs/5254981265?check_suite_focus=true#step:8:109

I think upgrade from `{name:Text}` to `{name:Text; age: ?Nat}` is allowed, but the stable check doesn't pass.

Another concern is the appearance of `Data/1`, which can be a problem for the parser.

crusso:
> @crusso Can you take a look at this test: https://github.com/dfinity/sdk/runs/5254981265?check_suite_focus=true#step:8:109
> 
> I think upgrade from `{name:Text}` to `{name:Text; age: ?Nat}` is allowed, but the stable check doesn't pass.



The upgrade from v1 to v2 is not allowed  because  stable compatibility is based on Motoko subtyping, not candid subtyping, and you can't add a record field when evolving to a super type. (v2 needs to be able to consume v1's stable variable stateb to restore its own state)

> Another concern is the appearance of `Data/1`, which can be a problem for the parser.
That is worrying but as long as the slash isn't appearing in the  '. Most' files we are OK. I think it's only appearing here in the error report. It would be nice if we actually showed the incompatible  type defs, but that's a long standing TODO. 

I'm sorely tempted to add/reintroduce a self test to the emission of the most files to check they parse and are well formed, crashing compilation if not (ie failing if stamps creep in)

crusso:
To make v1 compatible with v2 maybe add 'age : Null' to the v1's type

src/dfx/src/lib/builders/motoko.rs
chenyan-dfinity: Yeah, the hacky query method is already public, so the default should be public. Let me add a TODO to control this from dfx.json.

ericswanson-dfinity (approved):

PR #1933, opened 2021-11-19
chore: sync motoko updater with change to use prebuilt motoko binaries

ericswanson-dfinity:


crusso:
So how does this also get the appropriate motoko-base?

ericswanson-dfinity:
> So how does this also get the appropriate motoko-base?

It doesn't, and I'm not sure what to do about that.

ericswanson-dfinity:
This can't work with the new nix-less cargo build support, and doesn't work with motoko-base anyway, so we'll do this another way

PR #1940, opened 2021-11-24
fix: force RUST_MIN_STACK to 8MB for replica

ericswanson-dfinity:
For discussion.
- Is this hardcoded value ok?
- Should we only set it if RUST_MIN_STACK is not already set in the environment? (of course dfx it itself a rust application, so would be using the same value from the same environment variable)
- Should we add a new command-line argument to `dfx start`? (which would apply only when using the replica, not `--emulator`)


adambratschikaye:
I think we have to go with a hard coded value, at least for now, just because there's no easy way to get the value that's used in the mainnet. That value is stored in `ic-os/guestos/rootfs/etc/systemd/system/ic-replica.service` (in the main ic repo) and already has a comment that the nix environment needs to be updated when the value changes. So we could add an additional comment about dfx needing to be updated.

I'm not sure what we want to do if the user has set the `RUST_MIN_STACK` themselves. We're only changing the environment only for the replica process, so all other running rust code will have the stack size they've set and this seems ok to me.

I don't have anything against an additional argument, but I'm not sure if anyone would actually want to use it.

src/dfx/src/actors/replica.rs
```diff
@@ -254,6 +254,9 @@ fn replica_start_thread(
             // For our production network, we actually set them to 600ms.
             "600",
         ]);
+
+        cmd.env("RUST_MIN_STACK", "8192000");
```
adambratschikaye, line 258: Should we add a comment that this should agree with the value in https://gitlab.com/dfinity-lab/core/ic/-/blob/master/ic-os/guestos/rootfs/etc/systemd/system/ic-replica.service ?

adambratschikaye (approved):

src/dfx/src/actors/replica.rs
ericswanson-dfinity: Yes, will do

chenyan-dfinity:
Maybe worth adding the following code to our e2e test. When the replica increase the stack size in the future, this test case will fail.
```
(module
  (export "canister_init" (func $init))
  (func $init
    (call $init)
  )
)
```

PR #1943, opened 2021-12-01
fix: Stop using full network URI as a file path

pie-flavor:
This should make DFX only use one directory for a network. Fixes SDK-159


adamspofford-dfinity:
Wrong account

PR #1944, opened 2021-12-01
fix: Stop using full network URI as a file path

adamspofford-dfinity:
This should make DFX only use one directory for a network. Fixes `SDK-159`.


ericswanson-dfinity (changes requested):
Can we add a test (either unit test, or e2e test) for the new behavior?

src/dfx/assets/new_project_node_files/package.json
```diff
@@ -8,7 +8,7 @@
     "prebuild": "npm run copy:types",
     "start": "webpack serve --mode development --env development",
     "prestart": "npm run copy:types",
-    "copy:types": "rsync -avr .dfx/$(echo ${DFX_NETWORK:-'**'})/canisters/** --exclude='assets/' --exclude='idl/' --exclude='*.wasm' --delete src/declarations"
+    "copy:types": "rsync -avr .dfx/$(echo $DFX_NETWORK | sed -e 's/[^0-9a-zA-Z_]/_/g' -e 's/^$/**/')/canisters/** --exclude='assets/' --exclude='idl/' --exclude='*.wasm' --delete src/declarations"
```
ericswanson-dfinity, line 11: How about we make this transformation in dfx, where we set DFX_NETWORK?

https://github.com/dfinity/sdk/blob/master/src/dfx/src/lib/builders/assets.rs#L277

e2e/tests-dfx/provider.bash
```diff
@@ -57,3 +57,12 @@ teardown() {
     dfx canister --network "http://127.0.0.1:$webserver_port" create --all
     [ -d ".dfx/http___127_0_0_1_$webserver_port" ]
 }
+
+@test "network as URL does not create unexpected names" {
+    dfx_start
+    webserver_port=$(cat .dfx/webserver-port)
+    dfx canister --network "http://127.0.0.1:$webserver_port" create --all
+    dfx build --network "http://127.0.0.1:$webserver_port" create --all
```
ericswanson-dfinity, line 65:
```suggestion
    dfx build --network "http://127.0.0.1:$webserver_port" --all
```

src/dfx/assets/new_project_node_files/package.json
adamspofford-dfinity: Good catch - should I set it all the way up here? https://github.com/dfinity/sdk/blob/ce91cf8ab0a3306950ec2425850511da6a7e3db0/src/dfx/src/lib/builders/mod.rs#L235

ericswanson-dfinity: Yeah, that sounds even better - but please check usages of that struct member to make sure.

ericswanson-dfinity (approved):

PR #1949, opened 2021-12-06
fix: Rust canister resolve dependency

lwshang:


e2e/tests-dfx/rust.bash
```diff
@@ -20,10 +20,30 @@ teardown() {
...
     assert_command dfx canister --no-wallet call hello greet dfinity
     assert_match '("Hello, dfinity!")'
 }
+
+@test "rust canister can resolve dependencies" {
+    dfx_new_rust rust_deps
+    install_asset rust_deps
+
+    # TODO: cargo install may takes too much time
+    cargo install ic-cdk-optimizer
```
ericswanson-dfinity, line 37: I think this is failing on CI because both tests try to install ic-cdk-optimizer.

ericswanson-dfinity (approved):

.github/workflows/e2e.yml
```diff
@@ -22,13 +22,16 @@ jobs:
...
+            ~/.cargo/bin
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-${{ matrix.rust }}-1
       - name: Install Rust
         run: |
           rustup update ${{ matrix.rust }} --no-self-update
           rustup default ${{ matrix.rust }}
           rustup component add rustfmt
+      - name: Install ic-cdk-optimizer
+        run: cargo install ic-cdk-optimizer
```
ericswanson-dfinity, line 34: This would fit better in the provisioning scripts, and then only for the rust tests.  Example: https://github.com/dfinity/sdk/blob/master/scripts/workflows/provision-linux.sh#L21
If installed here, ic-cdk-optimizer won't be available during the e2e tests, because those run in a later job on a different runner.

lwshang: Thanks!

scripts/workflows/provision-darwin.sh
```diff
@@ -34,3 +34,5 @@ echo "BATSLIB=${BATS_SUPPORT}" >> "$GITHUB_ENV"
 
 # Exit temporary directory.
 popd
+
+cargo install ic-cdk-optimizer
```
ericswanson-dfinity, line 38: This takes a really long time.  Can we install this only for the rust tests?

ericswanson-dfinity (changes requested):

scripts/workflows/provision-darwin.sh
lwshang: I intended to have the `ic-cdk-optimizer` binary be cached. Can we?

ericswanson-dfinity: Sure, we can give that a shot.  We just need to be sure we aren't adding 12 minutes to the build time of each matrix job.  A place to start: right now, the `test` job, which runs this provisioning and the e2e tests themselves, does not have a cargo cache step.

https://github.com/dfinity/sdk/blob/master/.github/workflows/e2e.yml#L116

It also hasn't configured rust to any particular version, as is done in the `build` job here: https://github.com/dfinity/sdk/blob/master/.github/workflows/e2e.yml#L30

lwshang: I'm moving this `cargo install ic-cdk-optimizer` mainly for overcoming the 1200s timeout on macOS.
https://github.com/dfinity/sdk/runs/4462587938?check_suite_focus=true
May I increase this limit?

ericswanson-dfinity (approved):

PR #1950, opened 2021-12-06
feat: Allow adding and removing controllers

adamspofford-dfinity:
Fixes SDK-207


ericswanson-dfinity (changes requested):
Please add an entry to the changelog regarding the wallet.wasm version change (see changelog for other examples)

Also, we should probably also increment the wallet patch version in the cycles-wallet repo.

CHANGELOG.adoc
```diff
@@ -25,6 +25,11 @@ Updated Motoko from 0.6.11 to 0.6.14.
 * Improve performance of bignum equality
 * Stable signatures: frontend, metadata, command-line args
 
+== Cycles wallet
+
+Module hash: 983acc59e65d470e4e3ba97512864cdc9a2b560caaca1c89afc9f2fc78aa2643
+https://github.com/dfinity/cycles-wallet/commit/0f4e6c9f243f2a19caef7f7c552c9adc6ad43140
```
ericswanson-dfinity, line 31: Now that we have a new commit in the cycles-wallet repo main branch (57e53fcb679d1ea33cc713d2c0c24fc5848a9759), let's regenerate the wallet.wasm from that and update these hashes.

ericswanson-dfinity (approved):

ericswanson-dfinity (approved):
Please add a note to the changelog about this.  Approving under the assumption that this is done.

PR #1951, opened 2021-12-07
feat: Remove automatic wallet pass-through

adamspofford-dfinity:
Implements SDK-209. DFX currently forwards all canister interactions through the cycles wallet obtained via `dfx identity get-wallet`. This PR removes this behavior, as well as the `--no-wallet` flag that previously disabled it - canister operations will now always come straight from the user. The old way of doing things can continue to be used via `--wallet $(dfx identity get-wallet)`.


e2e/tests-dfx/certified_info.bash
```diff
@@ -18,7 +18,7 @@ teardown() {
     dfx_start
     dfx canister create hello
     assert_command dfx canister info "$(dfx canister id hello)"
-    assert_match "Controllers: $(dfx identity get-wallet) Module hash: None"
+    assert_match "Controllers: $(dfx identity get-principal) Module hash: None"
```
ericswanson-dfinity, line 21: Shouldn't both the wallet and the identity's principal be controllers?

ericswanson-dfinity:
Part of this work is going to be updating the docs at https://github.com/dfinity/docs .  I recommend creating that PR now (it will be merged in as part of the next release) and linking to it here.

e2e/tests-dfx/update_settings.bash
```diff
@@ -25,40 +25,40 @@ teardown() {
     ALICE_WALLET=$(dfx --identity alice identity get-wallet)
     BOB_WALLET=$(dfx --identity bob identity get-wallet)
 
-    dfx canister create hello
+    dfx canister --wallet "${ALICE_WALLET}" create hello
     dfx build hello
-    dfx canister install hello
```
ericswanson-dfinity, line 30: Shouldn't this work without passing `--wallet` (therefore using the default identity)?

Can we leave this test along (removing `--no-wallet` as needed), and instead add another that adapts this test for the case when `--wallet` is passed to canister create?

src/dfx/src/lib/identity/identity_utils.rs
```diff
@@ -1,42 +1,33 @@
...
+    //         let network = env
+    //             .get_network_descriptor()
+    //             .expect("No network descriptor.");
+    //         let identity_name = env.get_selected_identity().expect("No selected identity.");
+    //         let wallet =
+    //             Identity::get_or_create_wallet_canister(env, network, identity_name, true).await?;
+    //         CallSender::SelectedIdWallet(*wallet.canister_id_())
+    //     } else {
+    //         CallSender::SelectedId
+    //     }
```
ericswanson-dfinity, line 26: please delete before merging

e2e/tests-dfx/update_settings.bash
```diff
@@ -75,40 +75,40 @@ teardown() {
...
+    dfx canister --wallet "${ALICE_WALLET}" install hello
```
ericswanson-dfinity, line 80: Same as note for previous test

ericswanson-dfinity (changes requested):
Please describe in CHANGELOG.md the changes and what a developer should do about them.

e2e/tests-dfx/certified_info.bash
adamspofford-dfinity: No, because --wallet was not specified.

ericswanson-dfinity: Ok, let's discuss then, because I think there's a disconnect here.  After creating a canister without passing any particular parameters (as is done here), I would expect both the wallet and the dfx identity to be controllers.

adamspofford-dfinity: @krpeacock thoughts?

CHANGELOG.adoc
```diff
@@ -2,6 +2,12 @@
 
 == DFX
 
+=== feat!: Remove the wallet proxy and the --no-wallet flag
+
+Breaking change: Canister commands, except for `dfx canister create`, will make the call directly, rather than via the user's wallet. The `--no-wallet` flag is thus removed from `dfx canister` as its behavior is the default.
+
+When upgrading existing canisters, use the `--wallet` flag in conjunction with `dfx identity get-wallet` in order to restore the old behavior.
+
```
ericswanson-dfinity, line 10: We need to spell out in detail what commands developers need to execute in order to update the controllers of their existing canisters.

e2e/tests-dfx/create.bash
```diff
@@ -83,33 +102,33 @@ teardown() {
...
 
-
 @test "create accepts --controller <controller> named parameter, with controller by identity name" {
     dfx_start
     dfx identity new alice
     ALICE_PRINCIPAL=$(dfx --identity alice identity get-principal)
+    ALICE_WALLET=$(dfx --identity alice identity get-wallet)
 
-    assert_command dfx canister create --all --controller alice
+    assert_command dfx canister create --all --controller "${ALICE_PRINCIPAL}"
```
ericswanson-dfinity, line 111: Since this test verifies that `create accepts --controller <controller> named parameter, with controller by identity name`, let's leave this line as is (`--controller alice`), and instead add a new test that passes in the principal by ID.

That way this test will still be testing what it says it does (that `--controller` works with an identity name).

e2e/tests-dfx/create.bash
```diff
@@ -83,33 +102,33 @@ teardown() {
...
 }
 
 @test "create accepts --controller <controller> named parameter, with controller by principal" {
     dfx_start
     dfx identity new alice
-    ALICE_PRINCIPAL=$(dfx --identity alice identity get-principal)
+    ALICE_WALLET=$(dfx --identity alice identity get-wallet)
 
-    assert_command dfx canister create --all --controller "${ALICE_PRINCIPAL}"
+    assert_command dfx canister create --all --controller "${ALICE_WALLET}"
```
ericswanson-dfinity, line 125: Again if we need to test passing a wallet ID, let's add a new, separate test, and leave this test (which tests passing a controller by principal).

e2e/tests-dfx/update_settings.bash
```diff
@@ -25,40 +73,40 @@ teardown() {
...
```
ericswanson-dfinity, line 78: Let's leave this test as is with respect to these two ` canister create|install` lines, and if necessary, make a second test that creates/installs using `--wallet`.

e2e/tests-dfx/update_settings.bash
```diff
@@ -75,40 +123,40 @@ teardown() {
     ALICE_WALLET=$(dfx --identity alice identity get-wallet)
     BOB_WALLET=$(dfx --identity bob identity get-wallet)
 
-    dfx canister create hello
+    dfx canister --wallet "${ALICE_WALLET}" create hello
```
ericswanson-dfinity, line 126: same note as previous test: Let's leave this test as is with respect to these two ` canister create|install` lines, and if necessary, make a second test that creates/installs using `--wallet`.

e2e/tests-dfx/wallet.bash
```diff
@@ -135,39 +133,41 @@ teardown() {
...
 
-    CALL_RES=$(dfx canister --no-wallet call e2e_project fromCall)
+    CALL_RES=$(dfx canister --wallet "$WALLET" call e2e_project fromCall)
     CALLER=$(echo "${CALL_RES}" | cut -d'"' -f 2)
-    ID=$(dfx identity get-principal)
-    assert_eq "$CALLER" "$ID"
+    assert_eq "$CALLER" "$WALLET"
 
-    assert_command dfx canister --no-wallet call e2e_project amInitializer
-    assert_eq '(true)'
```
ericswanson-dfinity, line 157: Shouldn't we still have a test for this functionality? (create and deploy a canister with no wallet; call it directly)

e2e/tests-dfx/wallet.bash
```diff
@@ -135,39 +133,41 @@ teardown() {
...
+    WALLET=$(dfx identity get-wallet)
+    assert_command dfx deploy --wallet "$WALLET"
+    CALL_RES=$(dfx canister --wallet "$WALLET" call e2e_project fromCall)
     CALLER=$(echo "${CALL_RES}" | cut -d'"' -f 2)
-    ID=$(dfx identity get-principal)
-    assert_eq "$CALLER" "$ID"
+    assert_eq "$CALLER" "$WALLET"
 
-    assert_command dfx canister --no-wallet call e2e_project amInitializer
-    assert_eq '(true)'
```
ericswanson-dfinity, line 172: Can we retain this test (`dfx deploy` vs `dfx deploy --wallet "$WALLET"`, then the rest)

src/dfx/src/commands/deploy.rs
```diff
@@ -54,9 +54,9 @@ pub struct DeployOpts {
     #[clap(long)]
     wallet: Option<String>,
 
-    /// Performs the call with the user Identity as the Sender of messages.
+    /// Performs the create call with the user Identity as the Sender of messages.
     /// Bypasses the Wallet canister.
-    #[clap(long, conflicts_with("wallet"))]
+    #[clap(long)]
```
ericswanson-dfinity, line 59: shouldn't `--no-wallet` still conflict with `--wallet` ?

ericswanson-dfinity (changes requested):

e2e/tests-dfx/update_settings.bash
adamspofford-dfinity: I already did - see https://github.com/dfinity/sdk/blob/1b0a74e4c714c51885c7fa911eab8a15ad3d09bf/e2e/tests-dfx/update_settings.bash#L17

e2e/tests-dfx/update_settings.bash
adamspofford-dfinity: Does there need to be a second test here? I set it up so that for each of the actions 'set controller' and 'set multiple controllers', there's three versions: without wallet, with wallet, and with wallet version 0.7.2.

e2e/tests-dfx/wallet.bash
adamspofford-dfinity: Should we? The only thing that still retains wallet passthrough by default is `create`, which `amInitializer` wouldn't measure; AFAICT the old tests are now indistinguishable from the ones in identity.bash.

e2e/tests-dfx/update_settings.bash
ericswanson-dfinity: I was thinking it would be good to have a version of the test that uses the most-default parameters (`dfx canister create hello`), with wallet 0.7.2, which is what the version of the test to the left does.  What do you think?

adamspofford-dfinity: There can't be a 'without wallet, but with wallet 0.7.2' test because wallet 0.7.2 doesn't set the caller as another controller. Unless update-settings is used, but then all you're testing is that update-settings successfully nullifies the effect of the wallet on canister creation.

e2e/tests-dfx/wallet.bash
adamspofford-dfinity: As in, not changing the --wallet flag in the `canister call` call?

CHANGELOG.adoc
```diff
@@ -2,6 +2,16 @@
 
 == DFX
 
+=== feat!: Remove the wallet proxy and the --no-wallet flag
+
+Breaking change: Canister commands, except for `dfx canister create`, will make the call directly, rather than via the user's wallet. The `--no-wallet` flag is thus removed from `dfx canister` as its behavior is the default.
+
+When working with existing canisters, use the `--wallet` flag in conjunction with `dfx identity get-wallet` in order to restore the old behavior. To upgrade an existing canister to work seamlessly with the new system, the following command should be used (substituting `canister_name` for your own):
+[source, bash]
```
ericswanson-dfinity, line 10: What does "restore the old behavior" mean?

```suggestion
When working with existing canisters, use the `--wallet` flag in conjunction with `dfx identity get-wallet` in order to restore the old behavior.

You will need to upgrade each of your existing canisters to work with the new system.  To do so, execute the following command for each of your canisters (substituting `canister_name` for your own):
[source, bash]
```

CHANGELOG.adoc
```diff
@@ -2,6 +2,16 @@
...
+----
+dfx canister --wallet "$(dfx identity get-wallet)" update-settings <canister_name> --add-controller "$(dfx identity get-principal)"
```
ericswanson-dfinity, line 12: It seems like a lot to ask every developer to do this individually for each of their canisters.

Also, there's a little more to it for canisters that are on the ic network (since the local network is the default):
```
dfx canister --network ic --wallet "$(dfx identity --network ic get-wallet)"
 update-settings <canister_name> --add-controller "$(dfx identity get-principal)"
```
And even more if they use a different identity on the ic that they do for local development:
```
dfx --identity icident canister --network ic
 --wallet "$(dfx --identity icident identity --network ic get-wallet)"
 update-settings <canister_name>
 --add-controller "$(dfx --identity icident identity get-principal)"
```

Can you think of a way to simplify the upgrade process?

e2e/tests-dfx/wallet.bash
ericswanson-dfinity: If we aren't removing test coverage, then it's fine.

CHANGELOG.adoc
adamspofford-dfinity: Should I include both of those commands? That's a lot to edit in a paste, and they presumably know how the flags work already. Re: simplification, I'm not sure if you're asking for a shell solution or a DFX solution.

ericswanson-dfinity: Just to follow up after our team meeting: let's make sure to include the specific command to run for mainnet canisters:
```
dfx canister --network ic --wallet "$(dfx identity --network ic get-wallet)"
 update-settings <canister_name> --add-controller "$(dfx identity get-principal)"
```
Also, I think we should call out that `dfx --network ic wallet upgrade` is necessary.

ericswanson-dfinity: Actually we can streamline it a little bit already, using the `--all` parameter rather than calling once with each canister name:
```
dfx canister --network ic --wallet "$(dfx identity --network ic get-wallet)"
 update-settings --all --add-controller "$(dfx identity get-principal)"
```

adamspofford-dfinity: Should that be for the first command too?

ericswanson-dfinity: Yes I think we should specify `--all` in the instructions for both.

CHANGELOG.adoc
```diff
@@ -2,6 +2,23 @@
...
+You will need to upgrade each of your existing canisters to work with the new system.  To do so, execute the following command for each of your canisters:
+[source, bash]
+----
+dfx canister --wallet "$(dfx identity get-wallet)" update-settings --all --add-controller "$(dfx identity get-principal)"
+----
+Or, if you're upgrading a canister on the IC:
+[source, bash]
+----
+dfx canister --network ic --wallet "$(dfx identity --network ic get-wallet)" update-settings --all --add-controller "$(dfx identity get-principal)"
+----
```
ericswanson-dfinity, line 20:
```suggestion
You will need to upgrade your wallet and each of your existing canisters to work with the new system.  To do so, execute the following in each of your dfx projects:
[source, bash]
----
dfx wallet upgrade
dfx canister --wallet "$(dfx identity get-wallet)" update-settings --all --add-controller "$(dfx identity get-principal)"
----
To upgrade projects that you have deployed to the IC mainnet, execute the following:
[source, bash]
----
dfx wallet --network ic upgrade
dfx canister --network ic --wallet "$(dfx identity --network ic get-wallet)" update-settings --all --add-controller "$(dfx identity get-principal)"
----
```

ericswanson-dfinity (approved):

PR #1973, opened 2022-01-06
fix: retry ledger transfers

ericswanson-dfinity:
Retry ledger transfers for the following commands:
- dfx ledger create-canister
- dfx ledger top-up
- dfx ledger transfer

This PR accomplishes this through the following:
1. Set `created_at_time` to the same value for all attempts, so that if an earlier attempt was in fact processed but an error occurred while delivering the response, the ledger will detect subsequent calls as duplicates.
2. Use the ledger `transfer` method, which returns structured error information, rather than `send_dfx`.
3. Use the `transfer` method in an almost-idempotent way, by transforming `TransferError::TxDuplicate` errors into `Ok(block_height)`.

Addresses https://dfinity.atlassian.net/browse/SDK-211.

Also moved most of the types from nns_types/mod.rs to ledger_types/mod.rs to better match the location to which these types correspond.


src/dfx/src/commands/ledger/mod.rs
```diff
@@ -95,49 +98,92 @@ fn get_icpts_from_args(
...
+                    bail!(agent_err);
+                }
+            }
+        }
+    };
+
+    Ok(block_height)
+}
+
+async fn transfer_and_notify(
```
krpeacock, line 160: In what context is `transfer_and_notify` still being used, now that we have `transfer`?

ericswanson-dfinity: `ledger create-canister` and `ledger top-up` both call `transfer_and_notify`, which in turn calls `transfer`.

`ledger transfer` calls `transfer` directly, since it doesn't need to follow up with notify.

krpeacock (approved):

roman-kashitsyn:
Should we update the changelog?

ericswanson-dfinity:
> Should we update the changelog?

Yes, I'll follow up!

PR #1979, opened 2022-01-13
fix: update-settings --all inconsistency

adamspofford-dfinity:


e2e/tests-dfx/update_settings.bash
```diff
@@ -350,3 +350,21 @@ teardown() {
...
+    dfx identity use alice
+    dfx_start
+
+    dfx canister create --all
+    dfx build --all
+    dfx canister install --all
+
+    # make bob a controller
+    assert_command dfx canister update-settings --all --add-controller bob
+    # check that bob has the authority to make someone else a controller
```
ericswanson-dfinity, line 368: Can we call `dfx canister info` and verify that the list of controllers is as expected?  Both here and after adding charlie.

Otherwise we are using `dfx canister update-settings --all` to verify the functionality of `dfx canister update-settings --all`, and it seems pretty easy to imagine ways that that could miss something.

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #1993, opened 2022-01-23
Allow users to force reinstall a canister without prompting

jzxchiang1:
Useful for scripting


ggreif:
I have seen the `yes` tool to be used in such situations. Does your patch offer something of more value?

jzxchiang1:
> I have seen the `yes` tool to be used in such situations. Does your patch offer something of more value?

Oh interesting, didn't know there was a `yes` command! No need for this then.

PR #2000, opened 2022-01-31
chore: bump rustc toolchain to 1.58

chenyan-dfinity:
and fix clippy

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.55.0"
+channel = "1.58.0"
 components = ["rustfmt", "clippy"]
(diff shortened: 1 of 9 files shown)


ericswanson-dfinity (commented):
Please note that for the time being, our releases are built through the nix build, which due to the [dependency on common](https://github.com/dfinity-lab/common/blob/master/pkgs/overlays/rust/1_55.nix) uses toolchain 1.55.

chenyan-dfinity:
I see. Some crates we depend on doesn't seem to compile after `cargo update`. I will see what I can do.

chenyan-dfinity:
Resolved. We cannot run `cargo update`

PR #2003, opened 2022-02-01
fix: dfx generate support motoko

lwshang:
With current limitation of `moc`, `dfx generate` Motoko canisters actually build the canisters beforehand.


src/dfx/src/commands/generate.rs
```diff
@@ -35,12 +35,25 @@ pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
 
     // This is just to display an error if trying to generate before creating the canister.
     let store = CanisterIdStore::for_env(&env)?;
+    // If generate for motoko canister, build first
+    let mut build_before_generate = false;
     for canister in canister_pool.get_canister_list() {
-        store.get(canister.get_name())?;
+        let canister_id = store.get(canister.get_name())?;
+        if let Some(info) = canister_pool.get_canister_info(&canister_id) {
+            build_before_generate = info.get_type() == "motoko";
```
ericswanson-dfinity, line 43: What happens if a dfx project has both motoko canisters and rust canisters, and the rust canisters are defined after the motoko ones?
```suggestion
            if info.get_type() == "motoko" {
                build_before_generate = true;
            }
```

ericswanson-dfinity (changes requested):

src/dfx/src/commands/generate.rs
lwshang: `dfx generate` works similar to `dfx build`. The canisters dependencies graph is built first, and the bindings are generated in the proper order.

ericswanson-dfinity (approved):

PR #2004, opened 2022-02-01
chore: Update replica to 936bf9cc

ericswanson-dfinity:
The included 128-bit cycles support is needed in order to upgrade to Motoko 0.6.21.

This isn't the latest replica, because darwin binaries weren't generated for the latest replica (see https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/2914?commit_id=d45fbb038ab9cb993fce2e84e46d86b0b0a1ece7 for work in progress to fix).

Also removed a recursive chmod from `install_asset`, because in tests that restarted replica, replica noticed the permissions on the crypto directory granted too much access.


lwshang (approved):

ericswanson-dfinity:
The e2e/start.bash tests are failing after killing the replica process.  When the replica process tries to restart, it gets this error:
```
# thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: 
"Crypto state directory /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/dfx-e2e-XXXXXXXX.vFNSKcWh/working-dir/hello/.dfx/state/replicated_state/node-100/crypto 
has permissions 0o40772, allowing general access"', replica/src/setup.rs:331:75
```

PR #2005, opened 2022-02-01
chore: update replica to commit 50ead5a8 (linux only; darwin not available)

ericswanson-dfinity:


ericswanson-dfinity:
the problem was a `chmod -R a+w .`, since removed in another branch.

PR #2007, opened 2022-02-03
chore: respect provider when determining whether on IC or not

sesi200:
Attempting a PR from a different repo to test #2001 

Jira: [SDK-279](https://dfinity.atlassian.net/browse/SDK-279)

Previously: dfx only allowed to access the main IC net with two reserved strings for `--network`.
New: main IC net can also be accessed by an alias that looks like this: `"staging": { "providers": [ "https://ic0.app" ], "type": "persistent" }`


ericswanson-dfinity (approved):

sesi200:
Closed in favour of #2022

PR #2010, opened 2022-02-07
feat: remote canister support

ericswanson-dfinity:
Adds a `remote` section to canisters defined in dfx.json, which indicates that a canister already exists on certain networks and is managed by some other project.

The `remote` section can specify a canister ID per network, meaning dfx will use that canister ID when building other canisters, and will not create or install those canisters.

The `remote` section can also specify a `candid` file to use when building for remote networks, rather than the candid file either generated by `moc` or specified elsewhere for rust and custom canisters.

This facilitates writing mocks for remote canisters.  Additional public methods can be added for setting up data or responses for tests.  When building for a remote network, the `remote.candid` field is used in order to report inter-canister calls to those methods as errors.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -60,10 +60,13 @@ The canister will not be created or installed on these remote networks.
 For other networks, the canister will be created and installed as usual.
 
 The "candid" field, if set within the remote object, specifies the candid file to build against when
-building other canisters on a network for which the canister is remote.
+building other canisters on a network for which the canister is remote.  This definition can differ
+from the candid definitions for local builds.
 
 For example, if have an installation of the ICP Ledger (see https://github.com/dfinity/ic/tree/master/rs/rosetta-api/ledger_canister#deploying-locally[Ledger Installation Guide])
-in your dfx.json, you could configure the canister ID of the Ledger canister on the ic network like so:
+in your dfx.json, you could configure the canister ID of the Ledger canister on the ic network as below.  In this case,
+the private interfaces would be available for local builds, but only the public interfaces would be available
+when building for `--network ic`.
 [source, json]
 ----
 {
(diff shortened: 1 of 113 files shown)


CHANGELOG.adoc
```diff
@@ -30,6 +31,95 @@ Caused by:
...
 `dfx identity import` will now correctly identify these as EC keys, rather than Ed25519.
 
+=== feat: Remote canister support
+
+It's now possible to specify that a canister in dfx.json references a "remote" canister on a specific network,
+that is, a canister that already exists on that network and is managed by some other project.
+
+Motoko, Rust, and custom canisters may be configured in this way.
+
+This is the general format of the configuration in dfx.json:
```
oggy-dfin, line 41: Is this a complete minimal configuration? I.e., I don't need to set any additional properties to get something dfx likes?

CHANGELOG.adoc
```diff
@@ -30,6 +31,95 @@ Caused by:
...
+    }
+  }
+}
+----
+
+The "id" field, if set for a given network, specifies the canister ID for the canister on that network.
+The canister will not be created or installed on these remote networks.
+For other networks, the canister will be created and installed as usual.
+
+The "candid" field, if set within the remote object, specifies the candid file to build against when
```
oggy-dfin, line 62: Can I override a candid setting (e.g., in your ledger example, could I use a different candid file for the local and remote canisters?)

oggy-dfin: Also, could it be useful (not necessarily in this PR) to support arbitrary URLs for the candid file location?

CHANGELOG.adoc
ericswanson-dfinity: No, this is only the complete `remote` field, not a complete `dfx.json`.  A Motoko canister would need a `src` field, for example. And a `defaults`, `networks`, and `version` would also be required.

oggy-dfin: Sorry, I wasn't very clear - I meant if this is a minimal complete `canister` section?

ericswanson-dfinity: No, this shows the form of the `remote` section.  A complete `canister` section would be different for a Motoko, Rust, or custom canister.  Below are examples of a complete custom or Motoko section.

sesi200 (approved):

Issue #1648, opened 2021-05-07
[bug:] dfx new --no-frontend generates an unused <proj-name>_assets folder and canister.


crusso:
Nit: `dfx new --no-frontend` generates an unused <proj-name>_assets folder and canister.

```
crusso@vm:~/random_maze$ dfx --version
dfx 0.7.0-beta.6
crusso@vm:~$ dfx new --no-frontend random_maze
Fetching manifest https://sdk.dfinity.org/manifest.json
Creating new project "random_maze"...
crusso@vm:~$ tree random_maze/
random_maze/
├── dfx.json
├── README.md
└── src
    ├── random_maze
    │   └── main.mo
    └── random_maze_assets
        └── assets
            └── sample-asset.txt
4 directories, 4 files
```

PR #2020, opened 2022-02-11
fix: dfx new --no-frontend should not create an assets container

rikonor:
See https://github.com/dfinity/sdk/issues/1648

This change introduces the use of Handlebars templates to control whether dfx.json contains an assets container via the `--no-frontend` flag.

There is still the issue that the resulting project has a `src/<name>_assets` directory with a dummy text file.

Another approach would be to instead of dynamically generating a project directory, allow using pre-made example projects (See [Next.js example](https://nextjs.org/docs/api-reference/create-next-app#options)).
This would deprecate `--type [motoko, rust]` and `--frontend` / `--no-frontend`.
Instead, provide a default project (e.g motoko-based) vendored with `dfx` but also allow specifying an `--example` arg.


krpeacock (approved):

e2e/tests-dfx/bootstrap.bash
```diff
@@ -40,18 +40,3 @@ teardown() {
     assert_command_fail dfx bootstrap --port 8000
     assert_match "Cannot forward API calls to the same bootstrap server"
 }
-
-@test "bootstrap supports http requests" {
```
ericswanson-dfinity, line 44: Can we use `dfx_new_frontend` here, and keep this test?  Same for the other removed tests in this PR.

ericswanson-dfinity (changes requested):

.github/workflows/e2e.yml
```diff
@@ -91,7 +91,6 @@ jobs:
           fi
           time dfx deploy
           time dfx canister call smoke_backend greet '("fire")'
-          time curl --fail http://localhost:8000/sample-asset.txt?canisterId=$(dfx canister id smoke_frontend)
```
ericswanson-dfinity, line 94: Can we create this project with a frontend?  Ensuring that a basic project can serve assets is an important part of the "smoke" test.

PR #2033, opened 2022-02-17
feat: add remote generate-binding command

sesi200:
Jira: [SDK-138](https://dfinity.atlassian.net/browse/SDK-138)

Enhancement to #2010: Add ability to run `dfx remote generate-binding` to facilitate integrating with remote canisters.

The command creates bindings in the canister's `main` file if they have a `remote.candid` field in their dfx.json declaration.


e2e/tests-dfx/remote_generate_binding.bash
```diff
@@ -0,0 +1,34 @@
...
+
+    assert_command dfx remote generate-binding --all
+
+    if [[ ! -f remote.mo ]]; then
+        echo "remote.mo not created"
+        exit 1
+    fi
+}
+
+@test "generate-remote-binding --overwrite behaves as expected" {
```
ericswanson-dfinity, line 24: So far the tests only check behavior with `--all` for a remote Motoko canister.  Let's fill out these tests for different combinations.  Here are a few ideas to get you started.  Please see what others can you think of!
- generate-binding <specific Motoko canister> (main file does not exist)
- generate-binding <specific Rust canister> (main file does not exist)
- generate-binding <specific canister> (main file already exists)
- generate-binding <specific canister> (multiple remote canisters exist; none have their main file already.  what happens?)
- generate-binding (no arguments; multiple remote canisters exist; none have their main file already.  what happens?)
- generate-binding --overwrite <specific canister> (main file already exists)
- generate-binding --all (multiple canisters: one already has its main file, another does not)

ericswanson-dfinity (changes requested):

e2e/tests-dfx/remote_generate_binding.bash
sesi200: Added tests for all those cases, plus what I can think of:
* `generate-binding --overwrite --all` behaves as expected

e2e/tests-dfx/remote_generate_binding.bash
```diff
@@ -0,0 +1,205 @@
...
+    assert_command dfx remote generate-binding remote-rust
+
+    if [[ ! -f remote.rs ]]; then 
+        echo "remote.rs not created"
+        exit 1
+    fi
+
+    if [[ -f remote.mo ]]; then 
+        echo "remote.mo created when it shouldn't"
+        exit 1
```
ericswanson-dfinity, line 97: bats has specific functions for reporting failures, for example


```suggestion
        batslib_decorate "remote.mo created when it shouldn't"
        fail
```

sesi200: Good suggestion to refactor the tests. But you're actually supposed to pipe into these commands, otherwise the tests hang indefinitely. Correct use example:
```
echo "This is the error message" \
| batslib_decorate "This is the context similar to what we use in rust with .context()" \
| fail
```

I added proper assertions using these two functions. Now the tests are actually readable

ericswanson-dfinity (approved):

PR #2048, opened 2022-03-01
chore: Update replica to b2081fe

ericswanson-dfinity:
This is the latest replica on mainnet as seen on https://dashboard.mercury.dfinity.systems/release


ericswanson-dfinity:
Many failures of the form
```
thread 'main' panicked at 'No sandbox_launcher binary found'
```

PR #2049, opened 2022-03-02
fix: High Sierra, Mojave no longer supported

ericswanson-dfinity:
This is so that we display an error about the minimum supported OSX version being Catalina, rather than an error message about an expired cert.


sesi200 (approved):

ggreif (approved):
Wise move!

PR #2053, opened 2022-03-02
chore: update icx-proxy to 0230bff

ericswanson-dfinity:
This updates to a version for which the icx-proxy repo also produced release tarballs.  A later PR will use those instead.


ericswanson-dfinity:
Closing because this would require a feature flag; not obvious how to set this up under Nix.

PR #2055, opened 2022-03-03
test: make sure /_/candid interface is available with dfx start

ericswanson-dfinity:
`dfx start` and `dfx bootstrap` both provide access to the candid interface through the `/_/candid` web endpoint, but we only have an e2e test for `dfx bootstrap`.  This adds a test for the same behavior with `dfx start`.


chenyan-dfinity (approved):
Note that the `_/candid` endpoint will eventually be deprecated, once all canisters embed the candid interface in the Wasm module.

ericswanson-dfinity:
> Note that the `_/candid` endpoint will eventually be deprecated, once all canisters embed the candid interface in the Wasm module.

I will be very happy to remove the built-in webserver entirely once that happens!

PR #2064, opened 2022-03-08
chore: Update agent-rs to 0.13.0

ericswanson-dfinity:
Note that canister creation either goes through the management canister (128bit cycles) or the wallet (64bit cycles)


src/dfx/src/lib/operations/canister/create_canister.rs
```diff
@@ -93,7 +93,7 @@ pub async fn create_canister(
                 }
                 CallSender::Wallet(wallet_id) => {
                     let wallet = Identity::build_wallet_canister(*wallet_id, env)?;
-                    // amount has been validated by cycle_amount_validator
+                    // amount has been validated by cycle_amount_validator128 (!)
```
ericswanson-dfinity, line 96: We need to figure out what to do here; the string has been validated with the 128-bit validator.

PR #2072, opened 2022-03-10
feat: encrypt pem files 

sesi200:


src/dfx/src/lib/identity/identity_manager.rs
```diff
@@ -33,6 +30,38 @@ fn default_identity() -> String {
...
+    pub file_nonce: Vec<u8>,
+}
+
+impl EncryptionConfiguration {
+    /// Generates a random salt and nonce. Use this for every new identity
+    pub fn new() -> DfxResult<Self> {
+        let mut nonce: [u8; 12] = [0; 12];
+        let mut salt: [u8; 32] = [0; 32];
+        let sr = rand::SystemRandom::new();
+        sr.fill(&mut nonce)?;
```
randombit, line 54: As I understand it, this `EncryptionConfiguration` is going to be used to encrypt multiple different identities, at least potentially. If that's the case this is bad - if a nonce is reused with GCM, confidentiality is lost. A nonce/key pair must only be used a single time.

I'm not totally familiar with how this system works but some things to consider

* It's possible to generate both the AES key and the GCM nonce using Argon2
* Argon2 supports auxiliary inputs such as "associated data" (`ad` in the API). If you use Argon2 with the same password but different associated datas, you will generate new effectively unique keys for each `ad`. If you set the `ad` to the serialization of the identity name then you will get a unique key/nonce per file.

This *still* has a problem if you re-encrypt a *different* data under the same key/nonce. AIUI this doesn't happen here - the identity key is generated, and then later decrypted many times, but is never modified/rewritten.

sesi200: Thanks for the comments! `EncryptionConfiguration` is separate per identity. See example:
image

Using the identity name to derive anything (or using it as associated data) won't work since identities can be renamed at will (at least at the moment).

Using Argon2 to generate key and nonce is certainly possible. But what would you use as the input? Using `SystemRandom` is certainly possible, but I don't quite see how that would be a randomness upgrade (applying a deterministic function to an input shouldn't change it's entropy, right?). Using the password to derive them will lead to key/nonce reuse .

ericswanson-dfinity (commented):
What would you think about using a different filename or extension for the encrypted pem files?  It could even be `identity.pem.encrypted` or `identity.pem.argon2`.

```
$ cat identity.pem
?S??y??#???,??_??6?>??B(??Lp?~	O?^?=XRT? ...
```

ericswanson-dfinity (approved):
Looks good.  As discussed, let's follow up next week with what to do about the default identity, since for the time being it's still (automatically) created and stored unencrypted.

sesi200:
Waiting on @randombit to approve on behalf of security before I merge this.

PR #2084, opened 2022-03-18
chore: reactivate mitmproxy test on Linux

sesi200:


sesi200:
Problem was that the apt-get package for mitmproxy is not an official package, and is abandoned. Now the workflow directly downloads mitmproxy on linux from the official releases.

ericswanson-dfinity (approved):

PR #2090, opened 2022-03-22
feat: JS language bindings by default configure the HttpAgent to generate a Nonce for updates.

ericswanson-dfinity:
Must run `dfx generate` for this change to take effect in a project.

See https://forum.dfinity.org/t/calling-a-counter-multiple-times-asynchronously-with-promise-all-gives-the-same-value-each-time/11618/2

Fixes https://dfinity.atlassian.net/browse/SDK-390

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -1,6 +1,45 @@
 = dfx changelog
 :doctype: book
 
+= 0.9.4
+
+== DFX
+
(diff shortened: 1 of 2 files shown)


sesi200 (approved):

ericswanson-dfinity:
Let's do this instead in the HttpAgent.

ericswanson-dfinity:
We'll do this in the agent instead

PR #2099, opened 2022-03-24
fix: identities existing now is handled consistently

sesi200:
# Description

Steps to reproduce:
`dfx identity new alice`, cancel process during password prompt (or something goes wrong during command execution)
`dfx identity list` shows alice
`dfx identity use alice` claims alice does not exist
`dfx identity remove alice` says alice does not exist

Cause: Folder dfx/identities/alice is created, but without content. Folders are considered as identities to list, but dfx identity remove and dfx identity use require that the identity file within that folder exists.

Fixes Jira [SDK-397](https://dfinity.atlassian.net/browse/SDK-397)

# How Has This Been Tested?

Manual testing, do steps above.

# Checklist:

- [x] I have made corresponding changes to the documentation. No changes needed


src/dfx/src/lib/identity/mod.rs
```diff
@@ -40,6 +40,7 @@ pub const ANONYMOUS_IDENTITY_NAME: &str = "anonymous";
 pub const IDENTITY_PEM: &str = "identity.pem";
 pub const IDENTITY_PEM_ENCRYPTED: &str = "identity.pem.encrypted";
 pub const IDENTITY_JSON: &str = "identity.json";
+pub const TEMP_IDENTITY_NAME: &str = "___temp_identity_name";
```
ericswanson-dfinity, line 43: This is close!  Sometimes people run a number of operations concurrently, with a unique identity for each operation.

This came up recently: someone was running concurrently something like `dfx identity new something$n && dfx identity use something$n && dfx <do something>` for many n concurrently.  This caused problems since `dfx identity use` alters shared data.  The fix was `dfx identity new something$n && dfx --identity something$n <do something>`.

I'd suggest two changes:
- for the temporary identity directory name, use `___temp___<<new identity name>>`
- make `dfx identity list` ignore any directory that begins with `___temp___`

Here is a script that demonstrates the issue.  It works without the `&`.

```
#!/usr/bin/env bash

set -e

export DFX_CONFIG_ROOT="$(mktemp -d)"
trap "echo files were created in $DFX_CONFIG_ROOT"  EXIT

for a in $(seq 1 10); do
    echo $a
    dfx identity new --disable-encryption many$a &
done

wait

for a in $(seq 1 10); do
    dfx identity use many$a
done
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #2109, opened 2022-03-29
fix: Error if nonzero cycles are passed without a wallet proxy

adamspofford-dfinity:
`--with-cycles <n!=0>` is only valid in a call if it is proxying through the wallet.

### DIFF
--- a/src/dfx/src/commands/canister/call.rs
+++ b/src/dfx/src/commands/canister/call.rs
@@ -224,7 +224,7 @@ pub async fn exec(
         .map_or(0_u64, |amount| amount.parse::<u64>().unwrap());
 
     if call_sender == &CallSender::SelectedId && cycles != 0 {
-        bail!("Cannot provide cycles without proxying through the wallet (did you mean to use `canister --wallet`?)");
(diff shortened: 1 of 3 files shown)


src/dfx/src/commands/canister/call.rs
```diff
@@ -223,6 +223,10 @@ pub async fn exec(
         .as_deref()
         .map_or(0_u64, |amount| amount.parse::<u64>().unwrap());
 
+    if call_sender == &CallSender::SelectedId && cycles != 0 {
+        bail!("Cannot provide cycles without proxying through the wallet (did you mean to use `canister --wallet`?)");
```
ericswanson-dfinity, line 227:
```suggestion
        bail!("Cannot provide cycles without proxying through the wallet (did you mean to use `canister call --wallet`?)");
```

ericswanson-dfinity (changes requested):
Can we have the following?
- an e2e test (demonstrates the error message with `canister call --with-cycles`, and that `canister --wallet <wallet> call --with-cycles` succeeds)
- a changelog entry

ericswanson-dfinity (approved):

PR #2113, opened 2022-03-31
feat: dfx start and dfx replica: launch ic-btc-adapter when so configured

ericswanson-dfinity:
# Description

`dfx start` and `dfx replica` now launch the `ic-btc-adapter` binary, if either the `--btc-adapter-config <path>` parameter is passed, or the `.defaults.bitcoin.btc_adapter_config` field is set in dfx.json.

If a unix domain socket path is configured (`.incoming_source.Path` in the btc adapter config file), dfx will:
- delete this file before launching ic-btc-adapter,
- wait to launch replica until after this file exists (meaning ic-btc-adapter is listening on the unix domain socket).

If the ic-btc-adapter process exits (crashes), dfx will relaunch it, in the same way that it relaunches icx-proxy if the replica restarts.

Fixes https://dfinity.atlassian.net/browse/SDK-352

Gathering the docs changes here, but not opening a PR until the ic-starter work is done: https://github.com/dfinity/docs/tree/ericswanson/sdk-349-dfx-btc-integration

# How Has This Been Tested?

Through e2e tests, and by running `dfx start` and `dfx replica` with the new argument manually.

# Checklist:

- [x] (See above) I have made corresponding changes to the documentation.

### DIFF
--- a/src/dfx/src/commands/replica.rs
+++ b/src/dfx/src/commands/replica.rs
@@ -79,7 +79,7 @@ fn get_port(config: &ConfigDefaultsReplica, port: Option<String>) -> DfxResult<u
 
 /// Start the Internet Computer locally. Spawns a proxy to forward and
 /// manage browser requests. Responsible for running the network (one
-/// replica at the moment) and the proxy.
+/// replica at the moment), the proxy, and (if configured) the bitcoin adapter.
 pub fn exec(env: &dyn Environment, opts: ReplicaOpts) -> DfxResult {
     let system = actix::System::new();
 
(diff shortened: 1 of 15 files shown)


src/dfx/src/commands/replica.rs
```diff
@@ -63,27 +77,59 @@ fn get_port(config: &ConfigDefaultsReplica, port: Option<String>) -> DfxResult<u
...
-    shutdown_controller: Addr<ShutdownController>,
-) -> DfxResult {
-    let replica_config = get_config(env, opts)?;
-    start_replica_actor(env, replica_config, shutdown_controller)?;
-    Ok(())
-}
-
 /// Start the Internet Computer locally. Spawns a proxy to forward and
 /// manage browser requests. Responsible for running the network (one
 /// replica at the moment) and the proxy.
```
sesi200, line 82:
```suggestion
/// replica at the moment), the proxy, and (if configured) the bitcoin adapter.
```
AFAIU the adapter only gets started if it is configured either through CLI option or in dfx.json. Is it possible to make that a little more explicit? Right now that condition is hidden in line 113, where you may notice it or not. An `if let Some` would make that much more readable for me personally.

sesi200 (changes requested):

src/dfx/src/commands/replica.rs
ericswanson-dfinity: Great point.  Thanks!

sesi200 (approved):

PR #2118, opened 2022-04-06
feat: switch to 128-bit cycle counts

adamspofford-dfinity:


src/dfx/src/commands/canister/call.rs
```diff
@@ -73,12 +73,12 @@ struct CallIn {
...
     args: Vec<u8>,
-    cycles: u64,
+    cycles: u128,
 }
 
-async fn do_wallet_call(wallet: &Canister<'_, Wallet>, args: &CallIn) -> DfxResult<Vec<u8>> {
+async fn do_wallet_call(wallet: &WalletCanister<'_>, args: &CallIn) -> DfxResult<Vec<u8>> {
     let (result,): (Result<CallResult, String>,) = wallet
-        .update_("wallet_call")
+        .update_("wallet_call128")
```
ericswanson-dfinity, line 81: Should this go through CallForwarder so that the agent can call either `wallet_call` or `wallet_call128` depending on the wallet version?

src/dfx/src/commands/wallet/send.rs
```diff
@@ -24,13 +24,13 @@ pub async fn exec(env: &dyn Environment, opts: SendOpts) -> DfxResult {
...
-        amount: u64,
+        amount: u128,
     }
     let canister = Principal::from_text(opts.destination.clone())?;
     // amount has been validated by cycle_amount_validator
-    let amount = opts.amount.parse::<u64>().unwrap();
+    let amount = opts.amount.parse::<u128>().unwrap();
     let (res,): (Result<(), String>,) =
-        wallet_update(env, "wallet_send", In { canister, amount }).await?;
+        wallet_update(env, "wallet_send128", In { canister, amount }).await?;
```
ericswanson-dfinity, line 33: Should this go through the agent's wallet_send so that it can route to either `wallet_send` or `wallet_send128` ?

ericswanson-dfinity (changes requested):
Maybe you've already planned this, but please see if we can make use of the updated agent functions that route either to the 64-bit or 128-bit wallet functions depending on the wallet's version.

src/distributed/wallet.did
```diff
@@ -7,6 +7,7 @@ type EventKind = variant {
   CyclesReceived: record {
     from: principal;
     amount: nat64;
+    memo: opt text;
```
ericswanson-dfinity, line 10: Please add this .did file and the new wallet.wasm here: https://github.com/dfinity/sdk/tree/master/e2e/archive/wallet

src/dfx/src/commands/canister/call.rs
```diff
@@ -73,12 +73,17 @@ struct CallIn {
...
-    cycles: u64,
+    cycles: u128,
 }
 
-async fn do_wallet_call(wallet: &Canister<'_, Wallet>, args: &CallIn) -> DfxResult<Vec<u8>> {
-    let (result,): (Result<CallResult, String>,) = wallet
-        .update_("wallet_call")
+async fn do_wallet_call(wallet: &WalletCanister<'_>, args: &CallIn) -> DfxResult<Vec<u8>> {
+    // todo change to wallet.call when IDLValue implements ArgumentDecoder
+    let builder = if wallet.version_supports_u128_cycles() {
```
ericswanson-dfinity, line 81: Would you please add some e2e tests that exercise the new behavior?  There are examples in https://github.com/dfinity/sdk/blob/master/e2e/tests-dfx/wallet.bash that test using different wallet versions.

For example, for this piece, tests that show:
- succeeds calling previous wallet version with cycles <64 bits
- fails (reports error) calling previous wallet version, with cycles >64 bits
- succeeds calling new wallet version with cycles >64 bits

ericswanson-dfinity (changes requested):
Looks great!  Just needs some tests.

e2e/archive/wallet/0.9.0/wallet.did
```diff
@@ -0,0 +1,282 @@
+type EventKind = variant {
```
ericswanson-dfinity, line 1: This should be in wallet/0.10.0/, since that is the version in which this will be released.

ericswanson-dfinity (changes requested):
Looks good!  Please describe what's changed in the CHANGELOG, including notice to upgrade the wallet, and including the wasm hash and associated cycles wallet repo commit (as seen in the changelog when we've updated the wallet before)

PR #2120, opened 2022-04-07
docs: contributing: cargo just works

ericswanson-dfinity:
# Description

We don't need nix in order to build dfx anymore, and `cargo build` just works now.

Updated the contributing docs to say how to build dfx and how to run end to end tests without nix.


sesi200 (approved):

sesi200:
Merged it already since someone just wanted to build locally: https://forum.dfinity.org/t/where-is-the-source-code-of-dfx-command/11997

PR #2129, opened 2022-04-14
feat: Update notify operation for ledger changes

adamspofford-dfinity:
The ledger.notify function has been deprecated and replaced with the cmc.notify_top_up and cmc.notify_create_canister functions. This implements the dfx ledger operations in terms of those two new functions.

As cited in #2239 the CMC cannot be tested in an e2e test currently. This functionality has been tested manually.

Addresses SDK-399.


src/dfx/src/lib/ledger_types/mod.rs
```diff
@@ -74,16 +73,11 @@ impl fmt::Display for TransferError {
...
-pub enum CyclesResponse {
-    CanisterCreated(Principal),
-    ToppedUp(()),
-    Refunded(String, Option<BlockHeight>),
-}
-
 /// Position of a block in the chain. The first block has position 0.
 pub type BlockHeight = u64;
 
+pub type BlockIndex = u64;
```
ericswanson-dfinity, line 79: What's the difference between a `BlockHeight` and a `BlockIndex` ?

ericswanson-dfinity (approved):

src/dfx/src/lib/ledger_types/mod.rs
adamspofford-dfinity: I used the exact names from the Candid declaration for clarity. There is no actual difference, even semantically.

PR #2135, opened 2022-04-19
feat: add fabricate-cycles command

sesi200:
# Description

Fixes: [SDK-426](https://dfinity.atlassian.net/browse/SDK-426)

# How Has This Been Tested?

Added e2e tests that test...
- canister id and canister name
- `--all` instead of canister name
- default amount and specific number of cycles
- command cannot be run on real IC

# Checklist:

Documentation updates will be made once the name for the command is definitive.


src/dfx/src/commands/ledger/fabricate_cycles.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[clap(long, validator(cycle_amount_validator), conflicts_with("t"))]
+    amount: Option<String>,
+
+    /// Specifies the amount of trillion cycles to fabricate. Defaults to 10T cycles.
+    #[clap(
+        long,
+        validator(trillion_cycle_amount_validator),
+        conflicts_with("amount")
+    )]
+    t: Option<String>,
```
ericswanson-dfinity, line 30: Can we use `icp` and `e8s` arguments instead of `t` here, to be consistent with `dfx ledger top-up`, which is the equivalent mainnet command?

https://github.com/dfinity/sdk/blob/master/src/dfx/src/commands/ledger/top_up.rs#L31

ericswanson-dfinity (changes requested):

src/dfx/src/commands/ledger/fabricate_cycles.rs
sesi200: I do not think this is possible this easily. `provisional_top_up_canister` expects an amount in cycles, but `--e8s/--icp` would be an amount in ICP. We don't have an exchange rate set, therefore there's no ICP price to convert to cycles.

I'm trying to figure out how we can query the exchange rate, but it's not looking great

ericswanson-dfinity: Ah of course, you're right.  Nevermind!

ericswanson-dfinity (approved):

PR #2139, opened 2022-04-19
feat: Add --wasm flag to dfx-canister-install

adamspofford-dfinity:
This adds a `--wasm` flag to `dfx canister install`, so you can install an arbitrary wasm module if you already have one instead of going through the dfx.json process.


e2e/tests-dfx/install.bash
```diff
@@ -66,3 +66,15 @@ teardown() {
 
     assert_match "ComputeNetworkNotFound.*nosuch"
 }
+
+@test "install succeeds with arbitrary wasm" {
+    dfx_start
+    dfx canister create --all
+    assert_command dfx canister install e2e_project --wasm "${archive:?}/wallet/0.10.0/wallet.wasm"
```
ericswanson-dfinity, line 73: How about calling `dfx canister info` and verifying that the module hash matches the expected value?

src/dfx/src/commands/canister/install.rs
```diff
@@ -22,7 +24,7 @@ pub struct CanisterInstallOpts {
     canister: Option<String>,
 
     /// Deploys all canisters configured in the project dfx.json files.
-    #[clap(long, required_unless_present("canister"))]
+    #[clap(long, short, required_unless_present("canister"))]
```
ericswanson-dfinity, line 27: What is the rationale for adding `short` here?  We have very few command-line options that provide a short version, and even some of those are dubious.

src/dfx/src/commands/canister/install.rs
```diff
@@ -78,28 +84,44 @@ pub async fn exec(
...
+            install_canister_wasm(
+                env,
+                agent,
+                canister_id,
+                canister_info.as_ref().map(|info| info.get_name()).ok(),
+                &install_args,
+                mode,
+                timeout,
+                call_sender,
+                fs::read(wasm_path)?,
```
ericswanson-dfinity, line 102: Please add a .with_context() here
```suggestion
                fs::read(wasm_path).with_context(||format!("Unable to read {}", wasm_path.to_string_lossy()))?,
```

ericswanson-dfinity (changes requested):

src/dfx/src/commands/canister/install.rs
adamspofford-dfinity: Oh, that was left over from testing. If you do not want to keep it I can just revert it.

ericswanson-dfinity: Yeah, let's revert it please

ericswanson-dfinity (approved):

PR #2152, opened 2022-04-22
build: make release script work without nix-shell

ericswanson-dfinity:
# Description

Make it so the build script no longer requires nix to be installed.

At one point, I had a problem: how to update the first `version =` line in Cargo.toml.  "I know, I'll use sed," I thought.  Well, whichever version of sed nix bundles worked, but sed on OSX had other ideas.  Anyway, you know the rest of the story.  This PR makes the script use `awk` instead.

# How Has This Been Tested?

```
DRY_RUN=1 ./scripts/release.sh 9.8.7-beta.1
```

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -154,38 +154,35 @@ validate_default_project() {
 build_release_branch() {
 
     announce "Building branch $BRANCH for release $NEW_DFX_VERSION"
-    NIX_COMMAND=$(envsubst <<"EOF"
-        set -e
 
-        echo "Cleaning up cargo build files..."
-        $DRY_RUN_ECHO cargo clean --release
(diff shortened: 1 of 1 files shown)


sesi200 (approved):

nmattia:
Quick drive-by comment (not really important nor interesting):

> whichever version of sed nix bundles this, but sed on OSX had other ideas.

Most likely you got GNU sed from Nix, and I'm assuming BSD (macOS) sed choked on the "only replace first match" bit (`0,/...`)

ericswanson-dfinity:
> Most likely you got GNU sed from Nix, and I'm assuming BSD (macOS) sed choked on the "only replace first match" bit (`0,/...`)

Yes, that was the problem!

PR #2161, opened 2022-04-27
fix: Update built wallet wasm

adamspofford-dfinity:
The wallet wasm file that was included in 75bc046 likely introduced a UI regression because NPM was not being run properly locally. This updated binary was built correctly (and should be included in the 0.10 release).


ericswanson-dfinity (dismissed):

ericswanson-dfinity (changes requested):
Please also update https://github.com/dfinity/sdk/blob/master/e2e/archive/wallet/0.10.0/wallet.wasm

ericswanson-dfinity (approved):

PR #2163, opened 2022-04-28
chore: add a lot of context

sesi200:
# Description

Lots of errors don't add enough context, so it's possible to do something like `dfx canister call ...` and get `File not found`. Not helpful enough. This PR adds context to almost all places where errors are returned.

Fixes [SDK-418](https://dfinity.atlassian.net/browse/SDK-418)

# How Has This Been Tested?

No functional code changes, only adding contexts. One small exception: `fabricate_cycles.rs` mirrors the fix in #2155. CI should do the job just fine.

# Checklist:


src/dfx/src/config/cache.rs
```diff
@@ -144,44 +151,63 @@ pub fn install_version(v: &str, force: bool) -> DfxResult<PathBuf> {
...
             }
-            file.unpack_in(temp_p.as_path())?;
+            file.unpack_in(temp_p.as_path())
+                .context("Failed to unpack archive asset.")?;
 
-            let full_path = temp_p.join(file.path()?);
-            let mut perms = std::fs::metadata(full_path.as_path())?.permissions();
+            let full_path = temp_p.join(file.path().context("Failed to get file path.")?);
+            let mut perms = std::fs::metadata(full_path.as_path())
+                .context("Failed to get file metadata.")?
```
ericswanson-dfinity, line 178: Please include `full_path` in the context message here

src/dfx/src/config/cache.rs
```diff
@@ -194,36 +220,50 @@ pub fn install_version(v: &str, force: bool) -> DfxResult<PathBuf> {
...
 }
 
 pub fn list_versions() -> DfxResult<Vec<Version>> {
-    let root = get_bin_cache_root()?;
+    let root = get_bin_cache_root().context("Failed to get bin cache root.")?;
     let mut result: Vec<Version> = Vec::new();
 
-    for entry in std::fs::read_dir(root)? {
-        let entry = entry?;
+    for entry in std::fs::read_dir(root).context("Failed to read bin cache root content.")? {
```
ericswanson-dfinity, line 256: Please include `root` in the context message

src/dfx/src/config/dfinity.rs
```diff
@@ -379,7 +381,13 @@ impl ConfigInterface {
     }
 
     pub fn is_remote_canister(&self, canister: &str, network: &str) -> DfxResult<bool> {
-        Ok(self.get_remote_canister_id(canister, network)?.is_some())
+        Ok(self
+            .get_remote_canister_id(canister, network)
+            .context(format!(
+                "Failed to get remote canister id for {}.",
```
ericswanson-dfinity, line 387: Please include `network` here as well

src/dfx/src/config/dfinity.rs
```diff
@@ -469,7 +478,9 @@ pub struct Config {
 #[allow(dead_code)]
 impl Config {
     fn resolve_config_path(working_dir: &Path) -> DfxResult<Option<PathBuf>> {
-        let mut curr = PathBuf::from(working_dir).canonicalize()?;
+        let mut curr = PathBuf::from(working_dir)
+            .canonicalize()
+            .context("Failed to canonicalize working dir path.")?;
```
ericswanson-dfinity, line 483: Please include `working_dir` in the context message

ericswanson-dfinity (changes requested):
Wow, this is a very thorough approach!  I flagged a few instances where we ought to include canister names or use .with_context instead of .context as examples, but please apply these sorts of changes throughout.

src/dfx/src/actors/btc_adapter.rs
```diff
@@ -90,7 +90,10 @@ impl BtcAdapter {
 
         let (sender, receiver) = unbounded();
 
-        let handle = btc_adapter_start_thread(logger, self.config.clone(), addr, receiver)?;
+        let handle = anyhow::Context::context(
```
ericswanson-dfinity, line 93: I'm sure there's a good reason, but why does this the form `anyhow::Context::context(do_something(), "context")?`; instead of the form `do_something().context("context")?` used elsewhere in this PR?

src/dfx/src/actors/mod.rs
```diff
@@ -63,7 +70,12 @@ pub fn start_emulator_actor(
     // handing it over to ic-ref. If we read the file and it has
     // contents we shall assume it is due to our spawned ic-ref
     // process.
-    std::fs::write(&emulator_port_path, "")?;
+    std::fs::write(&emulator_port_path, "").with_context(|| {
+        format!(
+            "Failed to write/clear emulator port file {:?}.",
+            &emulator_port_path
```
ericswanson-dfinity, line 76: Is it more common/accepted to use the debug format, or to use .to_string_lossy()?
```suggestion
            "Failed to write/clear emulator port file {}.",
            emulator_port_path.to_string_lossy()
```

src/dfx/src/actors/mod.rs
```diff
@@ -35,7 +36,10 @@ pub fn start_btc_adapter_actor(
     shutdown_controller: Addr<ShutdownController>,
     btc_adapter_pid_file_path: PathBuf,
 ) -> DfxResult<Addr<BtcAdapter>> {
-    let btc_adapter_path = env.get_cache().get_binary_command_path("ic-btc-adapter")?;
+    let btc_adapter_path = env
+        .get_cache()
+        .get_binary_command_path("ic-btc-adapter")
+        .context("Failed to fetch 'ic-btc-adapter' binary.")?;
```
ericswanson-dfinity, line 42: Suggestion: use the verb "fetch" when downloading something, or otherwise obtaining something from a remote source:
```suggestion
        .context("Failed to determine path of 'ic-btc-adapter' binary.")?;
```

src/dfx/src/actors/mod.rs
```diff
@@ -54,7 +58,10 @@ pub fn start_emulator_actor(
     env: &dyn Environment,
     shutdown_controller: Addr<ShutdownController>,
 ) -> DfxResult<Addr<Emulator>> {
-    let ic_ref_path = env.get_cache().get_binary_command_path("ic-ref")?;
+    let ic_ref_path = env
+        .get_cache()
+        .get_binary_command_path("ic-ref")
+        .context("Failed to fetch 'ic-ref' binary.")?;
```
ericswanson-dfinity, line 64: Would you please take a look at `impl Cache for DiskBasedCache {
` in src/config/cache.rs?  It might provide a single place to add this particular type of context.  WDYT?

src/dfx/src/commands/bootstrap.rs
```diff
@@ -48,14 +48,17 @@ pub fn exec(env: &dyn Environment, opts: BootstrapOpts) -> DfxResult {
     let config = env.get_config_or_anyhow()?;
     let config_defaults = get_config_defaults_from_file(env);
     let base_config_bootstrap = config_defaults.get_bootstrap().to_owned();
-    let config_bootstrap = apply_arguments(&base_config_bootstrap, env, opts.clone())?;
+    let config_bootstrap = apply_arguments(&base_config_bootstrap, env, opts.clone())
+        .context("Failed to fetch bootstrap server configuration.")?;
```
ericswanson-dfinity, line 52:
```suggestion
        .context("Failed to apply arguments to bootstrap server configuration.")?;
```

src/dfx/src/commands/canister/call.rs
```diff
@@ -105,7 +105,8 @@ async fn do_wallet_call(wallet: &WalletCanister<'_>, args: &CallIn) -> DfxResult
         .with_arg(args)
         .build()
         .call_and_wait(waiter_with_exponential_backoff())
-        .await?;
+        .await
+        .with_context(|| format!("Failed wallet call with args {:?}.", args))?;
```
ericswanson-dfinity, line 109: Including the call arguments might be going too far

src/dfx/src/commands/canister/call.rs
```diff
@@ -185,20 +189,35 @@ pub async fn exec(
...
 
     let (canister_id, maybe_candid_path) = match CanisterId::from_text(callee_canister) {
         Ok(id) => {
             if let Some(canister_name) = canister_id_store.get_name(callee_canister) {
-                get_local_cid_and_candid_path(env, canister_name, Some(id))?
+                get_local_cid_and_candid_path(env, canister_name, Some(id)).with_context(|| {
+                    format!(
+                        "Failed to get canister id or candid path for {}",
+                        canister_name
+                    )
```
ericswanson-dfinity, line 202:
```suggestion
                    format!(
                        "Failed to get canister id or candid path for canister '{}' with id {:?}",
                        canister_name,
                        id
                    )
```

src/dfx/src/commands/canister/call.rs
```diff
@@ -185,20 +189,35 @@ pub async fn exec(
...
-            get_local_cid_and_candid_path(env, callee_canister, Some(canister_id))?
+            let canister_id = canister_id_store
+                .get(callee_canister)
+                .with_context(|| format!("Failed to get canister id for {}", callee_canister))?;
+            get_local_cid_and_candid_path(env, callee_canister, Some(canister_id)).with_context(
+                || {
+                    format!(
+                        "Failed to get canister id or candid path for {}",
+                        callee_canister
+                    )
```
ericswanson-dfinity, line 218:
```suggestion
                    format!(
                        "Failed to get local canister id or candid path for canister '{}' with id {:?}",
                        callee_canister,
                        id
                    )
```

src/dfx/src/commands/canister/delete.rs
```diff
@@ -81,25 +81,35 @@ async fn delete_canister(
...
 
     // Get the canister to transfer the cycles to.
     let target_canister_id = if no_withdrawal {
         None
     } else if to_dank {
-        Some(Principal::from_text("aanaa-xaaaa-aaaah-aaeiq-cai")?)
+        Some(
+            Principal::from_text("aanaa-xaaaa-aaaah-aaeiq-cai")
+                .context("Failed to read principal aanaa-xaaaa-aaaah-aaeiq-cai.")?,
+        )
```
ericswanson-dfinity, line 99: Another option here would be to use this form (example from elsewhere) to construct the principal constant, so no error is possible:
```
pub const MAINNET_CYCLE_MINTER_CANISTER_ID: Principal =
    Principal::from_slice(&[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x01, 0x01]);
```

Otherwise, I'd suggest including the term "dank" in the context message.

src/dfx/src/commands/canister/delete.rs
```diff
@@ -147,34 +163,44 @@ async fn delete_canister(
...
+            update_settings(env, canister_id, settings, timeout, call_sender)
+                .await
+                .with_context(|| {
+                    format!("Failed to update canister settings for {}.", canister_id)
+                })?;
 
             // Install a temporary wallet wasm which will transfer the cycles out of the canister before it is deleted.
-            let wasm_module = wallet_wasm(env.get_logger())?;
+            let wasm_module =
+                wallet_wasm(env.get_logger()).context("Failed to fetch wallet wasm file.")?;
```
ericswanson-dfinity, line 174:
```suggestion
                wallet_wasm(env.get_logger()).context("Failed to load wallet wasm file.")?;
```

src/dfx/src/commands/canister/delete.rs
```diff
@@ -147,34 +163,44 @@ async fn delete_canister(
...
+                .context("Failed to create argument blob.")?;
             let mode = InstallMode::Reinstall;
             let install_builder = mgr
                 .install_code(&canister_id, &wasm_module)
                 .with_raw_arg(args.to_vec())
                 .with_mode(mode);
             let install_result = install_builder
-                .build()?
+                .build()
+                .context("Failed to create build install call.")?
```
ericswanson-dfinity, line 189:
```suggestion
                .context("Failed to build InstallCode call.")?
```

src/dfx/src/commands/canister/deposit_cycles.rs
```diff
@@ -38,13 +38,17 @@ async fn deposit_cycles(
...
 ) -> DfxResult {
     let log = env.get_logger();
-    let canister_id_store = CanisterIdStore::for_env(env)?;
-    let canister_id =
-        Principal::from_text(canister).or_else(|_| canister_id_store.get(canister))?;
+    let canister_id_store =
+        CanisterIdStore::for_env(env).context("Failed to load canister id store.")?;
+    let canister_id = Principal::from_text(canister)
+        .or_else(|_| canister_id_store.get(canister))
+        .with_context(|| format!("Failed to get id for {}.", canister))?;
```
ericswanson-dfinity, line 45:
```suggestion
        .with_context(|| format!("Failed to get canister id for {}.", canister))?;
```

src/dfx/src/commands/canister/info.rs
```diff
@@ -23,22 +23,29 @@ pub async fn exec(env: &dyn Environment, opts: InfoOpts) -> DfxResult {
...
 
-    fetch_root_key_if_needed(env).await?;
+    fetch_root_key_if_needed(env)
+        .await
+        .context("Failed to fetch root key.")?;
     let controller_blob = agent
         .read_state_canister_info(canister_id, "controllers", false)
-        .await?;
+        .await
+        .with_context(|| format!("Failed to read canister {}'s controllers.", canister_id))?;
```
ericswanson-dfinity, line 39: I'm not sure we use the possessive (apostrophe) form anywhere else.
```suggestion
        .with_context(|| format!("Failed to read controllers of canister {}.", canister_id))?;
```

src/dfx/src/commands/canister/info.rs
```diff
@@ -23,22 +23,29 @@ pub async fn exec(env: &dyn Environment, opts: InfoOpts) -> DfxResult {
...
     let cbor: Value = serde_cbor::from_slice(&controller_blob)
         .map_err(|_| anyhow!("Invalid cbor data in controllers canister info."))?;
     let controllers = if let Value::Array(vec) = cbor {
         vec.into_iter()
             .map(|elem: Value| {
                 if let Value::Bytes(bytes) = elem {
-                    Ok(Principal::try_from(&bytes)?.to_text())
+                    Ok(Principal::try_from(&bytes)
+                        .context("Failed to construct principal.")?
```
ericswanson-dfinity, line 47: Even better would be to 
```suggestion
                        .context("Failed to construct principal of controller from bytes ({}).", hex::encode(&bytes))?
```

src/dfx/src/commands/canister/sign.rs
```diff
@@ -74,20 +74,35 @@ pub async fn exec(
...
+        CanisterIdStore::for_env(env).context("Failed to load canister id store.")?;
 
     let (canister_id, maybe_candid_path) = match Principal::from_text(callee_canister) {
         Ok(id) => {
             if let Some(canister_name) = canister_id_store.get_name(callee_canister) {
-                get_local_cid_and_candid_path(env, canister_name, Some(id))?
+                get_local_cid_and_candid_path(env, canister_name, Some(id)).with_context(|| {
+                    format!(
+                        "Failed to get local canister id and candid path for {}.",
+                        canister_name
```
ericswanson-dfinity, line 86: Please include id

src/dfx/src/commands/canister/sign.rs
```diff
@@ -74,20 +74,35 @@ pub async fn exec(
...
-            let canister_id = canister_id_store.get(callee_canister)?;
-            get_local_cid_and_candid_path(env, callee_canister, Some(canister_id))?
+            let canister_id = canister_id_store
+                .get(callee_canister)
+                .with_context(|| format!("Failed to get canister id for {}.", callee_canister))?;
+            get_local_cid_and_candid_path(env, callee_canister, Some(canister_id)).with_context(
+                || {
+                    format!(
+                        "Failed to get local canister id and candid path for {}.",
+                        callee_canister
```
ericswanson-dfinity, line 102: Please include canister_id

src/dfx/src/commands/canister/update_settings.rs
```diff
@@ -71,51 +73,67 @@ pub async fn exec(
...
+                        format!("Failed to get canister status for {}.", canister_id)
+                    })?;
                 controllers.get_or_insert(status.settings.controllers)
             };
             let removed = removed
                 .iter()
                 .map(|r| controller_to_principal(env, r))
-                .collect::<DfxResult<Vec<_>>>()?;
+                .collect::<DfxResult<Vec<_>>>()
+                .context("Failed to convert controller to principal.")?;
```
ericswanson-dfinity, line 136: This is unlikely to every come up, but I am wondering for cases like this that collect and transpose, would it work to move the context into the mapping?

```
.map(|r| controller_to_principal(env, r)
     .with_context(||format!("Failed to convert controller {} to principal", r)))
.collect:: ...
.context("Failed to convert controllers to principals")
```

src/dfx/src/commands/deploy.rs
```diff
@@ -150,26 +161,44 @@ fn display_urls(env: &dyn Environment) -> DfxResult {
...
                 .get_config()
-                .is_remote_canister(canister_name, &network.name)?
-            {
+                .is_remote_canister(canister_name, &network.name)
+                .with_context(|| {
+                    format!(
+                        "Failed to determine if canister {} is remote.",
+                        canister_name
+                    )
+                })?;
```
ericswanson-dfinity, line 172: I notice this is repeated (as an example; CanisterIdStore::for_env is similar in this respect).

This is a good example of a place where we aren't actually adding context (at this level) that can't be added by the method being called.  Add to that, that we're actually missing something (the network name), that would be easier to get right if it were done in one place.
```suggestion
                .is_remote_canister(canister_name, &network.name)
                .with_context(|| {
                    format!(
                        "Failed to determine if canister {} is remote for network {}.",
                        canister_name,
                        &network.name
                    )
                })?;
```

Can you think of a good solution to reduce code duplication?

src/dfx/src/commands/deploy.rs
```diff
@@ -150,26 +161,44 @@ fn display_urls(env: &dyn Environment) -> DfxResult {
...
             };
             if let Some(canister_id) = canister_id {
-                let canister_info = CanisterInfo::load(&config, canister_name, Some(canister_id))?;
+                let canister_info = CanisterInfo::load(&config, canister_name, Some(canister_id))
+                    .with_context(|| {
+                    format!(
+                        "Failed to load canister info for canister {}.",
+                        canister_name
+                    )
+                })?;
```
ericswanson-dfinity, line 187: Again a place where we aren't adding context; rather reporting information that the callee could report.

src/dfx/src/commands/generate.rs
```diff
@@ -16,48 +17,66 @@ pub struct GenerateOpts {
 }
 
 pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
-    let env = create_agent_environment(env, None)?;
+    let env = create_agent_environment(env, None).context("Failed to create AgentEnvironment.")?;
```
ericswanson-dfinity, line 20: Another example where we aren't adding any useful context at this level; it would be better if `create_agent_environment` provided this.

src/dfx/src/commands/identity/export.rs
```diff
@@ -14,7 +15,10 @@ pub struct ExportOpts {
 pub fn exec(env: &dyn Environment, opts: ExportOpts) -> DfxResult {
     let name = opts.identity.as_str();
 
-    let pem = IdentityManager::new(env)?.export(name)?;
+    let pem = IdentityManager::new(env)
+        .context("Failed to load identity manager.")?
+        .export(name)
+        .with_context(|| format!("Failed to export {}.", name))?;
```
ericswanson-dfinity, line 21: Another example where we are adding, from a higher level, context that could be added at a lower level.

ericswanson-dfinity (changes requested):
This looks like it was a ton of work.  I appreciate that.  And I'm wondering if we can find an approach where we add context messages when we're actually adding context about the *caller*, but in cases where we're only actually adding context about the *callee*, we depend on the callee to add its own context as appropriate.

It's challenging because in many cases, at a given level a method may have a bunch of `?` expressions in it, while overall the whole function itself would benefit from an outer .with_context call.  But instead, we're making every caller do that.

Let's discuss on Monday.

src/dfx/src/lib/package_arguments.rs
```diff
@@ -12,7 +12,8 @@ pub type PackageArguments = Vec<String>;
 pub fn load(cache: &dyn Cache, packtool: &Option<String>) -> DfxResult<PackageArguments> {
     if packtool.is_none() {
         let stdlib_path = cache
-            .get_binary_command_path("base")?
+            .get_binary_command_path("base")
+            .context("Failed to get path to 'base' binary.")?
```
ericswanson-dfinity, line 16: This one is tricky: the `get_binary_command_path` method is being used to get the path to a directory (of the base library), not an executable.
```suggestion
            .context("Failed to get path to 'base' library in cache.")?
```

src/dfx/src/actors/btc_adapter.rs
sesi200: Namespace clashes. This file imports `actix::Context`, so I can't also import `anyhow::Context`. If there's a better solution/Rust feature to solve this I'd be happy to learn about it

src/dfx/src/actors/mod.rs
sesi200: Looks like we use `to_string_lossy()` more often. I'll switch to that in the messages

src/dfx/src/commands/deploy.rs
```diff
@@ -183,13 +212,27 @@ fn display_urls(env: &dyn Environment) -> DfxResult {
...
+                            Url::parse(&url).with_context(|| {
+                                format!(
+                                    "Failed to parse candid url {} for canister {}.",
+                                    &url, canister_name
+                                )
+                            })?,
+                        );
                     } else if let Some(ui_canister_id) = ui_canister_id {
-                        let mut url = Url::parse(&network.providers[0])?;
+                        let mut url = Url::parse(&network.providers[0]).with_context(|| {
```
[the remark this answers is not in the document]
sesi200: Yeah, I think it makes sense to extract such a block to a method, and I just did so in the latest commit. But I think this is a general problem we have: We accept very deeply nested things in our code and are reluctant/too lazy to just factor out some functions, even though it would increase readability by quite a lot.

src/dfx/src/lib/identity/mod.rs
```diff
@@ -156,10 +164,17 @@ impl Identity {
 
         // Everything is created. Now move from the temporary directory to the actual identity location.
         let identity_dir = manager.get_identity_dir_path(name);
-        std::fs::rename(temp_identity_dir, identity_dir)?;
+        std::fs::rename(&temp_identity_dir, &identity_dir).with_context(|| {
+            format!(
+                "Failed to move temporary directory {} to permanent identiy directory {}.",
```
ericswanson-dfinity, line 169: typo
```suggestion
                "Failed to move temporary directory {} to permanent identity directory {}.",
```

src/dfx/src/lib/models/canister.rs
```diff
@@ -432,12 +498,26 @@ fn build_canister_js(canister_id: &CanisterId, canister_info: &CanisterInfo) ->
                 &canister_info.get_name().to_uppercase(),
             );
 
-        match decode_path_to_str(&file.path()?)? {
+        match decode_path_to_str(&file.path().context("Path is not valid unicode.")?)
+            .context("Failed to decode path to str.")?
```
ericswanson-dfinity, line 502: would `.context("Failed to decode path to str.")` be better as a context directive on `decode_path_to_str`?

ericswanson-dfinity (changes requested):

src/dfx/src/commands/start.rs
```diff
@@ -237,6 +253,15 @@ pub fn exec(
...
 
+        let webserver_bind = get_reusable_socket_addr(address_and_port.ip(), 0)?;
+
+        run_webserver(
+            env.get_logger().clone(),
+            build_output_root.clone(),
+            network_descriptor.clone(),
+            webserver_bind,
+        )?;
+
```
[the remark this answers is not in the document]
sesi200: Yes, it is. Thanks for catching that!

src/dfx/src/lib/environment.rs
```diff
@@ -226,12 +235,13 @@ impl<'a> AgentEnvironment<'a> {
         let mut identity_manager = IdentityManager::new(backend)?;
         let identity = identity_manager.instantiate_selected_identity()?;
 
-        let agent_url = network_descriptor.providers.first().unwrap();
+        let default = String::from(DEFAULT_IC_GATEWAY);
+        let agent_url = network_descriptor.providers.first().unwrap_or(&default);
```
ericswanson-dfinity, line 239: It turns out that `get_network_descriptor()` ensures that there is at least one provider, so in practice, this should never happen.  That's probably why we were getting away with an `.unwrap()` here.

That said, I don't think we should default to `DEFAULT_IC_GATEWAY` here.  The network name could be "staging", or "testnet12", or something like that.  It's kind of like saying, "if there's no connect string, connect to the production database."  In the wrong circumstances, lots could go wrong.

Can we change this to return an error with a context something like `Network '{}' does not specify any network providers` if there are no providers?  Alternatively, please leave it as an `.unwrap()`, and we can improve the encapsulation of NetworkDescriptor as future work.

src/dfx/src/lib/models/canister.rs
```diff
@@ -432,12 +499,18 @@ fn build_canister_js(canister_id: &CanisterId, canister_info: &CanisterInfo) ->
                 &canister_info.get_name().to_uppercase(),
             );
 
-        match decode_path_to_str(&file.path()?)? {
+        match decode_path_to_str(&file.path()?).context("Failed to decode path to str.")? {
```
ericswanson-dfinity, line 502: I think this context is now reported by the `decode_path_to_str` itself
```suggestion
        match decode_path_to_str(&file.path()?)? {
```

ericswanson-dfinity (changes requested):

src/dfx/src/lib/environment.rs
sesi200: Oh, yes, that is really dangerous. Thanks for catching that! I'll just add context. Luckily that works on Options too.

PR #2165, opened 2022-04-28
feat: generate bitcoin adapter configuration file

ericswanson-dfinity:
The configuration options are simple enough that dfx can provide them

Fixes https://dfinity.atlassian.net/browse/SDK-457


src/dfx/src/commands/replica.rs
```diff
@@ -88,26 +89,32 @@ pub fn exec(env: &dyn Environment, opts: ReplicaOpts) -> DfxResult {
...
+
     std::fs::write(&btc_adapter_pid_file_path, "")?;
 
+    let config = env.get_config_or_anyhow()?;
+    let btc_adapter_config = write_btc_adapter_config_if_enabled(
+        config.get_config(),
+        &btc_adapter_config_path,
+        opts.enable_bitcoin,
+        opts.bitcoin_node.clone(),
+    )?;
```
sesi200, line 102:
```suggestion
    let btc_adapter_config = write_btc_adapter_config_if_enabled(
        config.get_config(),
        &btc_adapter_config_path,
        opts.enable_bitcoin,
        opts.bitcoin_node.clone(),
    ).context("Failed to write BTC adapter config.")?;
```
Could you add some more `context`s in the spirit of #2163 ? I think right now it could produce a simple `Premission denied` with no way to understand what's happening.

src/dfx/src/commands/start.rs
```diff
@@ -160,6 +160,7 @@ pub fn exec(
 
     std::fs::write(&pid_file_path, "")?; // make sure we can write to this file
     std::fs::write(&btc_adapter_pid_file_path, "")?;
+    std::fs::write(&btc_adapter_config_path, "")?;
     std::fs::write(&icx_proxy_pid_file_path, "")?;
     std::fs::write(&webserver_port_path, "")?;
```
sesi200, line 165: same here

sesi200 (changes requested):

src/dfx/src/commands/replica.rs
ericswanson-dfinity: Done, but at a lower level, since this context message here wouldn't actually be adding context at this level.

sesi200 (approved):

PR #2177, opened 2022-05-05
feat: `dfx fix`/`dfx diagnose`

### LINKED ISSUE #2123: Deploying new canisters is rejected
**TLDR:**
Newly created canisters generate `Error: The replica returned an HTTP Error: Http Error: status 403 Forbidden, content type "application/cbor", content: Requested canister rejected the message` at deployment.

**Prerequisites:**
- Install dfx and Node 16.x (not 17.x) as described [here](https://smartcontracts.org/docs/quickstart/1-quickstart.html)
- Acquire IC cycles as described in [this guide](https://smartcontracts.org/docs/quickstart/4-quickstart.html)

**Scenario:**
1. Create a new directory and add files (html, css, js, dfx.json) as described in [Hosting a Static Website on the Internet Computer](https://smartcontracts.org/docs/samples/host-a-website.html)
2. Run a command `dfx deploy --network ic`

**Expected result:**
The canister is deployed with a message:
```
Deploying all canisters.
All canisters have already been created.
Building canisters...
Installing canisters...
Installing code for canister website, with canister_id osjqh-4aaaa-aaaai-acgra-cai
Uploading assets to asset canister...
Starting batch.
Staging contents of new and changed assets:
  /index.html 1/1 (490 bytes)
  /page-2.html 1/1 (356 bytes)
  /script.js 1/1 (114 bytes)
  /style.css 1/1 (105 bytes)
  /index.html (gzip) 1/1 (316 bytes)
  /page-2.html (gzip) 1/1 (254 bytes)
Committing batch.
Deployed canisters.
```
**Observed result:**
```
Error: The replica returned an HTTP Error: Http Error: status 403 Forbidden, content type "application/cbor", content: Requested canister rejected the message
```

**Temporary fix:**
Run the command at the root of the project:
`% dfx canister --wallet "$(dfx identity --network ic get-wallet)" --network ic update-settings --all --add-controller "$(dfx identity get-principal)"`

Then you can attempt to deploy again with `dfx deploy --network ic`

adamspofford-dfinity:
Implements SDK-454.

Fixes #2123.

This adds two new commands, `dfx fix` and `dfx diagnose`. Both operate off of a list of known environment-specific problems when upgrading DFX, currently:

- the wallet is outdated
- canisters have the wallet but not the identity as a controller

`diagnose` will find all the cases and tell you what commands can be used to fix them; `fix` will automatically fix them for you.


src/dfx/src/lib/migrate.rs
```diff
@@ -0,0 +1,116 @@
...
+        .expect("Could not get agent from environment");
+    let mut mgr = IdentityManager::new(env)?;
+    let ident = mgr.instantiate_selected_identity()?;
+    let wallet = Identity::wallet_canister_id(env, network, ident.name())
+        .map_err(|_| anyhow!("No wallet found; nothing to do"))?;
+    let wallet = WalletCanister::create(agent, wallet)
+        .await
+        .context("No accessible wallet found; nothing to do")?;
+    let mgmt = ManagementCanister::create(agent);
+    if !wallet.version_supports_u128_cycles() {
```
ericswanson-dfinity, line 43: Right now the newest version is the only wallet version that supports 128-bit cycles.

After the next wallet version is released, would you expect `dfx fix` to upgrade it?

src/dfx/src/lib/migrate.rs
```diff
@@ -0,0 +1,116 @@
...
+            let wasm = wallet_wasm(env.get_logger())?;
+            mgmt.install_code(wallet.canister_id_(), &wasm)
+                .with_mode(InstallMode::Upgrade)
+                .call_and_wait(waiter_with_timeout(expiry_duration()))
+                .await
+                .context("Could not upgrade wallet")?;
+        } else {
+            println!("The wallet is outdated; run `dfx wallet upgrade`");
+        }
+    }
```
ericswanson-dfinity, line 55: One of the things that people have been running into is trying to run the 0.9.0 upgrade procedure with an identity that isn't a controller of the wallet.

Can we check the wallet's controllers (available through the equivalent of `dfx canister info`), and if the current identity isn't a controller of the wallet, list identities that are?

src/dfx/src/lib/migrate.rs
```diff
@@ -0,0 +1,116 @@
...
+    for name in store.ids.keys() {
+        if let Some(id) = store.find(name) {
+            if mgmt
+                .canister_status(&id)
+                .call_and_wait(waiter_with_timeout(expiry_duration()))
+                .await
+                .is_err()
+            {
+                if fix {
+                    println!(
```
ericswanson-dfinity, line 66: It looks like we're assuming that any failure of the `canister_status` call means that the current identity isn't a controller of the canister in question.

What would you think about instead using the equivalent of `dfx canister info` (callable by anyone) and explicitly checking the controller list in order to determine if we should add the current identity as a controller?

ericswanson-dfinity (changes requested):

src/dfx/src/lib/migrate.rs
sesi200: Generating that list is non-trivial for encrypted identities. It's not difficult to store their principals (and maybe even the principals of their wallets?) in unencrypted format so we have the list available, but it would make it obvious to an attacker which pem file they would want to attempt to crack.

ericswanson-dfinity: Great point!  Similar for HSM.

adamspofford-dfinity: So should this be abandoned?

src/dfx/src/lib/migrate.rs
adamspofford-dfinity: Currently we don't tie wallet bundling into a strong version, so all we would have to go on is module hash, and there's no way of knowing whether a given module hash is newer or older. The version we *do* publish is tied to the API, and API changes would probably come with ic-utils changes, so what I've got here seems like the best option for now. Though it can be changed later if it proves not to be.

src/dfx/src/lib/migrate.rs
adamspofford-dfinity: Makes sense, I had forgotten you could query it without the mgmt canister's method.

src/dfx/src/lib/migrate.rs
ericswanson-dfinity: > One of the things that people have been running into is trying to run the 0.9.0 upgrade procedure with an identity that isn't a controller of the wallet.

Can you make it so the tool checks the current identity against the wallet canister's controllers, and displays a nice message about what is wrong and what to do ("This identity isn't a controller of the wallet.  The controllers of the wallet are  [list of controllers].  You need to run this command with an identity that is one of those" or something like that)
 , instead of an error message, which would be more difficult to understand?

src/dfx/src/lib/migrate.rs
```diff
@@ -0,0 +1,105 @@
...
+                                    memory_allocation: None,
+                                },
+                            },)),
+                            0,
+                        )
+                        .call_and_wait(waiter_with_timeout(expiry_duration()))
+                        .await
+                        .context("Could not update canister settings")?;
+                } else {
+                    println!("Canister {name} is outdated; run `dfx canister update-settings` with the --add-controller flag")
```
ericswanson-dfinity, line 99: What would you think about saying something about running `dfx fix`, in addition to displaying the individual commands that one could run?

src/dfx/src/lib/migrate.rs
```diff
@@ -0,0 +1,105 @@
...
+use super::{
+    environment::Environment,
+    error::DfxResult,
+    identity::{Identity, IdentityManager},
+    models::canister_id_store::CanisterIdStore,
+    network::network_descriptor::NetworkDescriptor,
+    root_key::fetch_root_key_if_needed,
+};
+
+pub async fn migrate(env: &dyn Environment, network: &NetworkDescriptor, fix: bool) -> DfxResult {
```
ericswanson-dfinity, line 27: Could you refactor the contents of this method into smaller methods?

src/dfx/src/lib/migrate.rs
```diff
@@ -0,0 +1,137 @@
...
+            .read_state_canister_info(wallet, "controllers", false)
+            .await?;
+        let controllers: Vec<Principal> = serde_cbor::from_slice(&cbor)?;
+        bail!("This identity isn't a controller of the wallet. You need to be one of these principals to upgrade the wallet: {}", controllers.into_iter().join(", "))
+    };
+    did_migrate |= migrate_wallet(env, agent, &wallet, fix).await?;
+    let store = CanisterIdStore::for_env(env)?;
+    for name in store.ids.keys() {
+        if let Some(id) = store.find(name) {
+            did_migrate |= migrate_canister(agent, &wallet, id, name, &ident, fix).await?;
```
ericswanson-dfinity, line 50: This should skip remote canisters

adamspofford-dfinity: good catch!

src/dfx/src/lib/migrate.rs
```diff
@@ -0,0 +1,143 @@
...
+    let mgmt = ManagementCanister::create(agent);
+    if !wallet.version_supports_u128_cycles() {
+        if fix {
+            println!("Upgrading wallet... ");
+            let wasm = wallet_wasm(env.get_logger())?;
+            mgmt.install_code(wallet.canister_id_(), &wasm)
+                .with_mode(InstallMode::Upgrade)
+                .call_and_wait(waiter_with_timeout(expiry_duration()))
+                .await
+                .context("Could not upgrade wallet")?;
```
ericswanson-dfinity, line 82: There's a piece missing here: storing the new wallet wasm with the wallet itself so that the wallet's `wallet_create_wallet` call (which can be called through the wallet's ui) can create a wallet with the new wallet version.  This is taken from the 'dfx wallet upgrade' code: 
```
    wallet
        .wallet_store_wallet_wasm(wasm)
        .call_and_wait(waiter_with_timeout(expiry_duration()))
        .await?;
```

ericswanson-dfinity: For extra credit, refactor this and `dfx wallet upgrade` so that they share code.

ericswanson-dfinity (approved):

PR #2184, opened 2022-05-09
feat: print wallet balance in a human readable form

smallstepman:
# Description

Changes the default behaviour of `dfx wallet balance` - the command will print the balance as trillion cycles. Additionally, a new flag has been added: `dfx wallet balance --precise`, which will print unscaled balance (default behaviour pre this PR).

Closes [SDK-436](https://dfinity.atlassian.net/browse/SDK-436)

# How Has This Been Tested?
- `cargo run wallet balance`
- `cargo run wallet balance --precise`
- unit tests

# Checklist:


src/dfx/src/commands/wallet/balance.rs
```diff
@@ -3,13 +3,128 @@ use crate::lib::environment::Environment;
...
+
+fn pretty_thousand_separators(i: String) -> String {
+    // 1. walk backwards (reverse string) and return characters until comma is seen
+    // 2. once comma is seen, start counting chars and:
+    //   - every third character but not at the end of the string: return (char + separator)
+    //   - otherwise: return char
+    // 3. re-reverse the string
+    const SEPARATOR: char = ',';
+    let mut count: u32 = 0;
+    let mut seen_comma = false;
```
ericswanson-dfinity, line 60: I got a bit confused by this, because `,` is a comma (in English anyway).  I know 123.456,789 and 123,456.789 are both valid formats depending on the locality.
```suggestion
    const THOUSANDS_SEPARATOR: char = ',';
    const DECIMAL_POINT: char = '.';
    let mut count: u32 = 0;
    let mut seen_decimal_point = false;
```

ericswanson-dfinity (changes requested):
Great test coverage!

src/dfx/src/commands/wallet/balance.rs
```diff
@@ -3,13 +3,128 @@ use crate::lib::environment::Environment;
 use crate::lib::error::DfxResult;
 
 use clap::Parser;
+use rust_decimal::prelude::*;
```
sesi200, line 6: Maybe @ericswanson-dfinity: is is acceptable style to use everything in a module? I don't necessarily think using everything in a prelude is wrong, but it is much more clear if everything is imported explicitly.

sesi200, line 6: Maybe @ericswanson-dfinity: is is acceptable style to use everything in a module? I don't necessarily think using everything in a prelude is wrong, but it is much more clear if everything is imported explicitly.

src/dfx/src/commands/wallet/balance.rs
smallstepman: Thanks for pointing this out! Admittedly that was quite confusing. I replaced that with `DECIMAL_POINT` and `GROUP_DELIMITER`

src/dfx/src/commands/wallet/balance.rs
```diff
@@ -3,13 +3,128 @@ use crate::lib::environment::Environment;
...
+pub struct WalletBalanceOpts {
+    /// Get balance raw value (without upscaling to trillions of cycles).
+    #[clap(long)]
+    precise: bool,
+}
 
 pub async fn exec(env: &dyn Environment, _opts: WalletBalanceOpts) -> DfxResult {
     let balance = get_wallet(env).await?.wallet_balance().await?;
-    println!("{} cycles.", balance.amount);
+    if _opts.precise {
```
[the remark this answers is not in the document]
smallstepman: Thanks - was wondering if that's the convention here, but I failed to find other examples, tho I haven't looked too hard ^^. Fixed.

src/dfx/src/commands/wallet/balance.rs
smallstepman: nice catch, thank you so much! fixed :)

smallstepman:
Looks like lint is failing. Any idea why and how to fix it?

ericswanson-dfinity:
For the documentation piece, you can open a PR to this branch: https://github.com/dfinity/portal/tree/dfx-0.10.1

I'd suggest to start with these sections:
- https://github.com/dfinity/portal/blob/master/docs/developer-docs/build/project-setup/default-wallet.md#check-the-cycle-balance
- https://github.com/dfinity/portal/blob/master/docs/references/cli-reference/dfx-wallet.adoc#examples-2

ericswanson-dfinity (approved):

PR #2196, opened 2022-05-12
fix: identities with configured wallets are not broken anymore and re…

sesi200:
…moved only when using the --drop-wallets flag

# Description

```
WALLET="rwlgt-iiaaa-aaaaa-aaaaa-cai"
dfx identity new --disable-encryption alice
dfx identity use alice
dfx identity --network ic set-wallet --force "$WALLET"
dfx identity use default
dfx identity remove alice
```
This removes the identity pem file, but leaves the wallets.json config in the identity folder. Removing the id fails, but the id is not usable anymore.

Solution: `dfx identity remove alice` only removes alice iff it has no wallets configured. Otherwise, it will display the wallets and a message to include `--drop-wallets` in order to remove the identity. If `--drop-wallets` is added (`dfx identity remove alice --drop-wallets`), then the identity is removed.

Fixes [error report on the forum](https://forum.dfinity.org/t/cannot-remove-identity-directory/12910)

# How Has This Been Tested?

Added an e2e test for this scenario.

# Checklist:

### DIFF
--- a/e2e/tests-dfx/identity_command.bash
+++ b/e2e/tests-dfx/identity_command.bash
@@ -133,6 +133,7 @@ teardown() {
 }
 
 @test "identity remove: only remove identities with configured wallet if --drop-wallets is specified" {
+    # There's no replica running, and no real wallet.  This is just a valid principal.
     WALLET="rwlgt-iiaaa-aaaaa-aaaaa-cai"
     assert_command dfx identity new --disable-encryption alice
     assert_command dfx identity use alice
--- a/e2e/tests-dfx/identity_command.bash
+++ b/e2e/tests-dfx/identity_command.bash
@@ -132,6 +132,17 @@ teardown() {
     assert_command_fail dfx identity remove charlie
 }
 
+@test "identity remove: only remove identities with configured wallet if --drop-wallets is specified" {
+    WALLET="rwlgt-iiaaa-aaaaa-aaaaa-cai"
+    assert_command dfx identity new --disable-encryption alice
+    assert_command dfx identity use alice
+    assert_command dfx identity --network ic set-wallet --force "$WALLET"
+    assert_command dfx identity use default
+    assert_command_fail dfx identity remove alice
+    assert_match "$WALLET"
+    assert_command dfx identity remove alice --drop-wallets
+}
+
 @test "identity remove: cannot remove the non-default active identity" {
     assert_command dfx identity new --disable-encryption alice
     assert_command dfx identity use alice
--- a/e2e/tests-dfx/identity_command.bash
+++ b/e2e/tests-dfx/identity_command.bash
@@ -139,8 +139,9 @@ teardown() {
     assert_command dfx identity --network ic set-wallet --force "$WALLET"
     assert_command dfx identity use default
     assert_command_fail dfx identity remove alice
-    assert_match "$WALLET"
+    assert_match "identity 'alice' on network 'ic' has wallet $WALLET"
     assert_command dfx identity remove alice --drop-wallets
+    assert_match "identity 'alice' on network 'ic' has wallet $WALLET"
 }
 
 @test "identity remove: cannot remove the non-default active identity" {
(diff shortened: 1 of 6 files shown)


sesi200:
Open question: Is this a reasonable approach to removing identities?

src/dfx/src/lib/identity/identity_manager.rs
```diff
@@ -233,14 +235,33 @@ impl IdentityManager {
...
         self.require_identity_exists(name)?;
 
         if self.configuration.default == name {
             return Err(DfxError::new(IdentityError::CannotDeleteDefaultIdentity()));
         }
 
+        let wallet_config_file = self.get_persistent_wallet_config_file(name);
+        if wallet_config_file.exists() {
+            if drop_wallets {
+                remove_identity_file(&wallet_config_file)?;
```
ericswanson-dfinity, line 256: Would it make sense to display the linked wallets here too?

e2e/tests-dfx/identity_command.bash
```diff
@@ -132,6 +132,17 @@ teardown() {
...
+    WALLET="rwlgt-iiaaa-aaaaa-aaaaa-cai"
+    assert_command dfx identity new --disable-encryption alice
+    assert_command dfx identity use alice
+    assert_command dfx identity --network ic set-wallet --force "$WALLET"
+    assert_command dfx identity use default
+    assert_command_fail dfx identity remove alice
+    assert_match "$WALLET"
+    assert_command dfx identity remove alice --drop-wallets
+}
+
```
ericswanson-dfinity, line 145: Would you please add a test that demonstrates that `dfx identity remove` displays the wallet ids when removing a persistent identity?

You can configure the local network to be persistent for this case.  Here's a snippet from another test that does this:
```
    # shellcheck disable=SC2094
    cat <<<"$(jq .networks.local.type=\"persistent\" dfx.json)" >dfx.json
```

ericswanson-dfinity (approved):
Approved assuming an e2e test that demonstrates the display of the associated persistent wallet ids.

e2e/tests-dfx/identity_command.bash
sesi200: On line 139 `set-wallet --force` adds the wallet for a persistent network and line 142 checks that the wallet is listed there. I made it bit more clear what the test is actually checking for.

I'd like to avoid the `setup_actuallylocal_network` because that would make the test run for a much longer time. That's why I'm using the `--force` flag. Does this sufficiently test the functionality?

e2e/tests-dfx/identity_command.bash
```diff
@@ -132,6 +132,19 @@ teardown() {
     assert_command_fail dfx identity remove charlie
 }
 
+@test "identity remove: only remove identities with configured wallet if --drop-wallets is specified" {
+    WALLET="rwlgt-iiaaa-aaaaa-aaaaa-cai"
```
ericswanson-dfinity, line 136: Rather than hardcoding the wallet id, we can read it with `dfx identity get-wallet`

e2e/tests-dfx/identity_command.bash
ericswanson-dfinity: Yes, that makes sense.

e2e/tests-dfx/identity_command.bash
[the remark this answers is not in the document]
sesi200: This again requires a running replica which I'd like to avoid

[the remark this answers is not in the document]
ericswanson-dfinity: Oh, I see.  There's no actual wallet.  Ok.

ericswanson-dfinity, line 136:
```suggestion
    # There's no replica running, and no real wallet.  This is just a valid principal.
    WALLET="rwlgt-iiaaa-aaaaa-aaaaa-cai"
```

ericswanson-dfinity (approved):
I left a suggestion about describing what's going on in the test, because when I read it I thought the test was using a value that just happened to equal the actual wallet id.

e2e/tests-dfx/identity_command.bash
sesi200: I like this solution, thank you!

PR #2198, opened 2022-05-12
feat: publish manifest and install script to github

pikajude:
You can check the contents of the manifest here https://github.com/dfinity/sdk/suites/6485129500/artifacts/240458728

- [x] manifest.json matches
- [x] install.sh passes shellcheck
- [x] install.sh is formatted
- [x] install.sh contains the right git revision

Once this is approved I'll switch the artifact upload step to a "push to new branch" step using GHPagesUploadAction.

### DIFF
--- a/.github/workflows/publish-manifest.yml
+++ b/.github/workflows/publish-manifest.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Generate
         run: |
           shellcheck --shell=sh public/install/*.sh --exclude SC2039,SC2154,SC2034
-          shfmt -d -p -i 4 -ci -bn -s public/install/*.sh
+          ~/go/bin/shfmt -d -p -i 4 -ci -bn -s public/install/*.sh
           mkdir _out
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-manifest.yml
```diff
@@ -0,0 +1,35 @@
+name: Publish manifest
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
ericswanson-dfinity, line 7: I think we only want to run this on push to master
```suggestion
```

ericswanson-dfinity (changes requested):
The contents of the files look good.  Once this only runs on push to master it looks good to go.

.github/workflows/publish-manifest.yml
pikajude: Yeah I just had to test it

ericswanson-dfinity (approved):

PR #2227, opened 2022-05-24
fix: Remove legacy environment variable

adamspofford-dfinity:
`CANISTER_CANDID_{name}` was kept for compatibility in 0.9, but deprecated in favor of `CANISTER_CANDID_PATH_{name}`. Removing it was planned for 0.10.


ericswanson-dfinity (changes requested):
Please call this out in the changelog

adamspofford-dfinity:
Going to set this as draft until a CDK release has been made with the change.

ericswanson-dfinity (approved):

PR #2239, opened 2022-06-01
test: e2e tests ledger subcommands with local replica

lwshang:
# Description

Get pre-built nns canisters and using `ic-nns-tool` to install and initialize them on local system replica.
Then run e2e tests of `dfx ledger` subcommands.

`ic-nns-tool` only works with `replica` but `ic-ref`. So the `ic-ref` flavor of this test is skipped.

Note that, the subcommands (`create-canister`, `top-up`, `notify`) involved `cycles-minting-canister`(CMC) are not testable for now. Because `ic-nns-tool` cannot set `icp_xdr_conversion_rate` in CMC, which is a crucial parameter in cycles minting. In production, the `icp_xdr_conversion_rate` is set with proposal which is not approachable in dfx e2e environment.

# Checklist:

- [x] ~I have edited the CHANGELOG accordingly.~
- [x] ~I have made corresponding changes to the documentation.~


e2e/utils/setup_nns.bash
```diff
@@ -0,0 +1,66 @@
+#! /bin/bash
+
+# The commit hash was get from ic release note
+IC_COMMIT="3b5d893c0857c47715fc339112e5dd1dbfff77a8"
+
+if test -z "${NNS_ARTIFACTS}"
+then
+  NNS_ARTIFACTS=$(mktemp -d -t dfx-e2e-nns-artifacts-XXXX) # Linux mktemp requires Xs in suffix
```
ericswanson-dfinity, line 8: Can we clean these up after the tests run?  It looks like the intent here is to download the artifacts once per invocation of ledger.bash, rather than once per test (which is why the artifacts aren't stored under DFXE2E_TEMP_DIR, created in standard_setup and cleaned up in standard_teardown).

It's convenient that the test automatically downloads the artifacts that it needs, but this also means that every time someone runs `bats e2e/tests-dfx/ledger.bash`,  it downloads 10 wasm files an the ic-nns-init binary, and leaves them in a temp dir.

Overall I'd suggest calling an equivalent of this script from the provision scripts, and exporting E2E_NNS_ARTIFACTS with the location of the artifacts as a precondition for running these tests.

lwshang: Thx! I'll move them into provision scripts.

ericswanson-dfinity (approved):

PR #2249, opened 2022-06-08
fix: reuse unix domain socket paths

ericswanson-dfinity:
# Description

The unix domain socket paths end up in the replica configuration.  Since ic-starter currently does not overwrite a configuration file if one already exists, any changes would not take effect for a new replica.

This PR makes dfx reuse unix domain socket paths used by an earlier run, so the replica configuration does
not change.

Also, this PR makes `dfx stop` wait until the dfx process and its children actually exit.  This is because
`dfx start` deletes the unix domain sockets from the /tmp directory just before it exits.  An earlier
invocation of dfx start could delete the unix domain socket files out from under the new invocation.

Fixes https://dfinity.atlassian.net/browse/SDK-518

# How Has This Been Tested?

Tested manually, because the relevant error messages only show up in the replica logs, and we don't yet have any e2e tests that verify actual canister http or bitcoin functionality, just that the processes start and stop.

```
$ rm -rf .dfx ; dfx-wip start --enable-canister-http --enable-bitcoin --background && dfx-wip stop && dfx-wip start --enable-canister-http --enable-bitcoin
```

# Checklist:


src/dfx/src/commands/start.rs
```diff
@@ -436,23 +443,48 @@ pub fn configure_btc_adapter_if_enabled(
...
     let pid = sysinfo::get_current_pid()
-        .map_err(|s| anyhow!("Unable to obtain pid of current process: {}", s))
-        .context("Unable to construct btc adapter socket path")?;
+        .map_err(|s| anyhow!("Unable to obtain pid of current process: {}", s))?;
+    let timestamp_seconds = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_secs();
 
     // Unix socket domain names can only be so long.
```
ggreif, line 459: "socket domain" -> "domain socket"

sesi200 (approved):

PR #2250, opened 2022-06-09
feat: run cargo audit when building rust canisters

sesi200:
# Description

When `cargo-audit` is installed, dfx now checks for vulnerabilities when building rust canisters. To make warnings and errors more obvious, `warn!` and `error!` are now formatted with colored `ERROR` or `WARN` prefixes.

New message styling: 
image
image

Fixes [SDK-471](https://dfinity.atlassian.net/browse/SDK-471)

# How Has This Been Tested?

Added e2e tests.

# Checklist:


e2e/assets/vulnerable_rust_deps/Cargo.lock
```diff
@@ -0,0 +1,963 @@
+# This file is automatically @generated by Cargo.
```
lwshang, line 1: Should we include this Cargo.lock file?

sesi200: My assumption was that this would make cargo not auto-update to fixed versions, but given that Cargo.toml specifies the vulnerable version explicitly, including this is not necessary. Removed it.

lwshang (approved):

PR #2258, opened 2022-06-10
refactor: motoko-base-tarball -> motoko-base

ericswanson-dfinity:
In sources.json:
1. Remove motoko-base definition
2. Rename motoko-base-tarball to motoko-base

The motoko-base definition, which referenced the next-moc branch in the motoko-base repo, was no longer used.

No need for anything in the changelog because these are just internal references


ggreif (approved):
LGTM!

PR #2262, opened 2022-06-14
fix: canister call uses remote candid file

viviveevee:
# Description

If dfx is not able to determine the canister type, it just gives up at trying to determine the candid file.

Fixes [SDK-506](https://dfinity.atlassian.net/browse/SDK-506)

# How Has This Been Tested?

Added this situation to an existing e2e test.

# Checklist:

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -49,11 +49,7 @@ jobs:
       - name: Audit
         run: |
           cd sdk
-          # RUSTSEC-2020-0159 - direct dependency on chrono 0.4.19 "Potential segfault in `localtime_r` invocations"
-          # RUSTSEC-2020-0071 - indirect dependency on time 0.1.43 "Potential segfault in the time crate"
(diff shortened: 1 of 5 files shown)


ericswanson-dfinity (changes requested):
Would you please add an e2e test that demonstrates the problem this PR fixes?  The test should fail prior to this change and succeed after it.

Here is an e2e test that shows canister call making a call to a remote canister, using the remote canister's candid definition: https://github.com/dfinity/sdk/blob/master/e2e/tests-dfx/remote.bash#L57

Is it the case that this PR makes it so canister call works for remote canisters before dfx deploy?

sesi200:
I found a basic cause for the problem: If dfx was not able to determine the canister type, it just gave up at trying to determine the candid file. e2e now checks that it also works in that case.

Changes to `get_local_cid_and_candid_path` are reverted.

ericswanson-dfinity (approved):

PR #2266, opened 2022-06-15
feat: post-install actions

adamspofford-dfinity:
This adds the ability to specify post-installation tasks with a per-canister `post-install` key in dfx.json. Post-installation tasks receive the same environment variables that build tasks do.


src/dfx/src/config/dfinity.rs
```diff
@@ -62,6 +63,7 @@ pub const DEFAULT_IC_GATEWAY_TRAILING_SLASH: &str = "https://ic0.app/";
 /// It only contains a type; everything else should be infered using the
 /// CanisterInfo type.
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
```
ericswanson-dfinity, line 66: Is there a way to configure serde to accept both kebab case and the default?  We already have `subnet_type`.

adamspofford-dfinity: It doesn't affect anything except the named fields of this particular struct. Unless you were talking about overall naming convention?

ericswanson-dfinity: Yes, bringing it up for overall consistency.

src/dfx/src/lib/operations/canister/install_canister.rs
```diff
@@ -159,6 +164,52 @@ pub async fn install_canister(
         post_install_store_assets(canister_info, agent, timeout).await?;
     }
 
+    if !canister_info.get_post_install().is_empty() {
```
ericswanson-dfinity, line 167: A suggestion: move this all into a separate method, in part so that it can have its own `#[context]`, and in part because the rest of the parts of this method would be better refactored into smaller methods.

e2e/tests-dfx/install.bash
```diff
@@ -81,3 +81,19 @@ teardown() {
...
+
+@test "install runs post-install tasks" {
+    dfx_start
+    # shellcheck disable=SC2094
+    cat <<<"$(jq '.canisters.e2e_project."post-install"="sh -c \"echo hello\""' dfx.json)" >dfx.json
+
+    assert_command dfx canister create --all
+    assert_command dfx build
+
+    assert_command dfx canister install --all
```
ericswanson-dfinity, line 93: Can we have e2e tests demonstrating expected environment variables are provided for dependencies in the following cases?
- `dfx install --all`
- `dfx install <specific canister>`
- `dfx deploy` (all)
- `dfx deploy <specific canister>`

e2e/tests-dfx/install.bash
```diff
@@ -81,3 +81,19 @@ teardown() {
     dfx canister create --all
     assert_command_fail dfx canister install --all --wasm "${archive:?}/wallet/0.10.0/wallet.wasm"
 }
+
+@test "install runs post-install tasks" {
+    dfx_start
+    # shellcheck disable=SC2094
+    cat <<<"$(jq '.canisters.e2e_project."post-install"="sh -c \"echo hello\""' dfx.json)" >dfx.json
```
ericswanson-dfinity, line 88: It looks like SerdeVec is meant to support the post-install field being either an array or a string.  Can we have an e2e test that demonstrates expected behavior when the `post-install` field is an array?

ericswanson-dfinity (changes requested):

src/dfx/src/config/dfinity.rs
adamspofford-dfinity: Then it makes more sense not to include kebab at all.

e2e/tests-dfx/install.bash
```diff
@@ -81,3 +81,40 @@ teardown() {
...
+    # shellcheck disable=SC2094
+    cat <<<"$(jq '.canisters.e2e_project.post_install=["sh -c \"echo hello\"", "sh -c \"return 1\""]' dfx.json)" >dfx.json
+    assert_command_fail dfx canister install --all --mode upgrade
+    assert_match hello
+}
+
+@test "post-install tasks receive environment variables" {
+    dfx_start
+    # shellcheck disable=SC2094
+    cat <<<"$(jq '.canisters.e2e_project.post_install="sh -c \"echo hello $CANISTER_ID\""' dfx.json)" >dfx.json
```
ericswanson-dfinity, line 105: This PR goes to some trouble to make sure it includes environment variables for dependencies.  Could we add tests that demonstrate that those environment variables are present when expected?

e2e/tests-dfx/install.bash
```diff
@@ -81,3 +81,53 @@ teardown() {
     dfx canister create --all
     assert_command_fail dfx canister install --all --wasm "${archive:?}/wallet/0.10.0/wallet.wasm"
 }
+
+@test "install runs post-install tasks" {
+    dfx_start
+    # shellcheck disable=SC2094
+    cat <<<"$(jq '.canisters.e2e_project.post_install="sh -c \"echo hello\""' dfx.json)" >dfx.json
```
ericswanson-dfinity, line 88: Should these have the same kind of handling that the custom build script field does?  All of these post-install scripts are running `sh -c "something"`, but if the post-install script was "post-install.sh", would you expect that to work without a leading `./`?

e2e/tests-dfx/install.bash
```diff
@@ -81,3 +81,54 @@ teardown() {
...
+    install_asset post_install
+    dfx_start
+    echo "echo \$CANISTER_ID" >> postinstall.sh
+
+    assert_command dfx canister create --all
+    assert_command dfx build
+    id=$(dfx canister id postinstall_script)
+
+    assert_command dfx canister install --all
+    assert_match "hello $id"
```
ericswanson-dfinity, line 113: Can you make the output for the two canisters different?  Here we expect both post-install scripts to execute, but can't tell the difference.

e2e/tests-dfx/install.bash
```diff
@@ -81,3 +81,54 @@ teardown() {
...
+
+    assert_command dfx canister install postinstall
+    assert_match hello
+
+    assert_command dfx canister install postinstall_script
+    assert_match hello
+    
+    echo 'return 1' >> postinstall.sh
+    assert_command_fail dfx canister install postinstall_script --mode upgrade
+    assert_match hello
```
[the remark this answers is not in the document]
adamspofford-dfinity: It's expected because stdout is still printed; it's tested for to ensure that the script is why the command failed.

e2e/tests-dfx/install.bash
```diff
@@ -81,3 +81,54 @@ teardown() {
...
+@test "post-install tasks discover dependencies" {
+    install_asset post_install
+    dfx_start
+    echo "echo hello \$CANISTER_ID_postinstall" >> postinstall.sh
+
+    assert_command dfx canister create --all
+    assert_command dfx build
+    id=$(dfx canister id postinstall)
+    
+    assert_command dfx canister install --all
```
ericswanson-dfinity, line 132: how about `dfx canister install postinstall_script` here?  That might make it more clear that `the `postinstall_script` canister has access to the `postinstall` (a dependency) canister ID.

ericswanson-dfinity (approved):

Issue #2199, opened 2022-05-12
Post-build step for asset canister fails because directory is considered to be outside the workspace root


paulyoung:
My `dfx.json` looks like this:

```json
{
  "version": 1,
  "dfx": "0.9.2",
  "canisters": {
    "myproject": {
      "type": "custom",
      "build": "nix build",
      "candid": "backend/src/myproject/src/myproject.did",
      "wasm": "result/target/wasm32-unknown-unknown/release/myproject.wasm"
    },
    "myproject_assets": {
      "type": "assets",
      "dependencies": [
        "myproject"
      ],
      "frontend": {
        "entrypoint": "result/dist/index.html"
      },
      "source": [
        "result/dist/"
      ]
    }
  },
  "networks": {
    "local": {
      "bind": "127.0.0.1:8000",
      "type": "ephemeral"
    }
  }
}
```

There's apparently some issue with using `"result/dist/"` as the source directory because when I run `dfx deploy myproject_assets` I see `Error: The post-build step failed for canister 'ryjl3-tyaaa-aaaaa-aaaba-cai' with an embedded error: Directory at '/Users/py/projects/paulyoung/myproject/main/result/dist/' is outside the workspace root.`

I suspect this is because `result` is actually a symlink to `/nix/store/<hash>-myproject`. I'm working around this for now but it might be nice if people could opt out of this check.

Also, is there a way to specify a build command for asset canisters? The above relies on the dependency of `myproject` to put the assets in the right place. It would help with my workaround if I could decouple the two.

paulyoung:
My workaround is as follows.

`"build": "bash -c \"nix build && rm -rf result-assets && mkdir result-assets && cp -R result/dist result-assets && chmod 775 -R result-assets\"",`

I'm using `bash -c` because otherwise is treated `&&` as a command and I get an error complaining that it doesn't exist.

It would be great if things like `foo && bar` just worked.

sesi200:
Hi Paul, I feel your pain! We have reworking the custom build command on our backlog and know the current state is mediocre at best. Thanks for giving good examples that we can test against!

paulyoung:
I think I saw some code that looked like it might try to write to the provided directory as well. If that's true then that would have been another problem I would have run into.

paulyoung:
@sesi200 I saw some PRs mentioning this so I wondered if I had missed anything that might allow me to work around this now.

paulyoung:
[...]

PR #2267, opened 2022-06-15
fix: Support direct custom build script

adamspofford-dfinity:
Addresses SDK-476. Currently, `type:custom`'s `build` flag obeys shell rules, i.e. `"script.sh"` does not work and requires `"./script.sh"`. This changes the behavior to support the specific case of `"script.sh"`, but *not* `"script.sh --args"`, for the same reason as the shell disallows it in the first place.

### DIFF
--- a/src/dfx/src/lib/builders/custom.rs
+++ b/src/dfx/src/lib/builders/custom.rs
@@ -190,7 +190,7 @@ impl CanisterBuilder for CustomBuilder {
 fn run_command(args: Vec<String>, vars: &[super::Env<'_>]) -> DfxResult<()> {
     let (command_name, arguments) = args.split_first().unwrap();
     let command_path = Path::new(command_name);
-    let mut cmd = if arguments.len() == 0
+    let mut cmd = if arguments.is_empty()
         && command_path.components().count() == 1
         && command_path.exists()
     {
--- a/src/dfx/src/lib/builders/custom.rs
+++ b/src/dfx/src/lib/builders/custom.rs
@@ -13,8 +13,8 @@ use ic_types::principal::Principal as CanisterId;
 use serde::Deserialize;
 use slog::info;
 use slog::Logger;
-use std::path::PathBuf;
-use std::process::Stdio;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
 
 /// Set of extras that can be specified in the dfx.json.
 struct CustomBuilderExtra {
@@ -189,8 +189,16 @@ impl CanisterBuilder for CustomBuilder {
 
 fn run_command(args: Vec<String>, vars: &[super::Env<'_>]) -> DfxResult<()> {
     let (command_name, arguments) = args.split_first().unwrap();
-
-    let mut cmd = std::process::Command::new(command_name);
+    let command_path = Path::new(command_name);
+    let mut cmd = if arguments.len() == 0
+        && command_path.components().count() == 1
+        && command_path.exists()
+    {
+        // handle "command.sh" with no ./
+        Command::new(format!("./{command_name}"))
+    } else {
+        Command::new(command_name)
+    };
 
     cmd.args(arguments)
         .stdout(Stdio::inherit())
--- a/e2e/assets/custom_canister/build.sh
+++ b/e2e/assets/custom_canister/build.sh
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+echo CUSTOM_CANISTER2_BUILD_DONE
--- a/e2e/assets/custom_canister/dfx.json
+++ b/e2e/assets/custom_canister/dfx.json
@@ -6,6 +6,12 @@
       "candid": "main.did",
       "wasm": "main.wasm",
       "build": "echo CUSTOM_CANISTER_BUILD_DONE"
+    },
+    "custom2": {
+      "type": "custom",
+      "candid": "main.did",
+      "wasm": "main.wasm",
+      "build": "build.sh"
     }
   },
   "defaults": {
--- a/e2e/tests-dfx/build.bash
+++ b/e2e/tests-dfx/build.bash
@@ -128,11 +128,14 @@ teardown() {
   install_asset custom_canister
   dfx_start
   dfx canister create --all
-  assert_command dfx build
+  assert_command dfx build custom
   assert_match "CUSTOM_CANISTER_BUILD_DONE"
+  assert_command dfx build custom2
+  assert_match "CUSTOM_CANISTER2_BUILD_DONE"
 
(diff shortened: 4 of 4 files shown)


src/dfx/src/lib/builders/custom.rs
```diff
@@ -189,8 +189,16 @@ impl CanisterBuilder for CustomBuilder {
...
+    let command_path = Path::new(command_name);
+    let mut cmd = if arguments.is_empty()
+        && command_path.components().count() == 1
+        && command_path.exists()
+    {
+        // handle "command.sh" with no ./
+        Command::new(format!("./{command_name}"))
+    } else {
+        Command::new(command_name)
+    };
```
sesi200, line 201: In case the command does not really work (e.g. the mentioned `script.sh --arg` or maybe `curl a && curl b`), can you please also add an explanation to the error message (or somewhere else in our docs) explaining a) that this does not work intentionally and b) what to do instead?

I've had people run into this and simply getting hung up on this a few times already.

sesi200: Also see: https://github.com/dfinity/sdk/issues/2199#issuecomment-1125520224

src/dfx/src/lib/builders/custom.rs
```diff
@@ -189,8 +189,16 @@ impl CanisterBuilder for CustomBuilder {
 
 fn run_command(args: Vec<String>, vars: &[super::Env<'_>]) -> DfxResult<()> {
     let (command_name, arguments) = args.split_first().unwrap();
-
-    let mut cmd = std::process::Command::new(command_name);
+    let command_path = Path::new(command_name);
```
ericswanson-dfinity, line 192: There are other places where this is broken, but dfx tries to let you execute it from anywhere within a project's directory tree.

See resolve_config_path 
 (near https://github.com/dfinity/sdk/blob/master/src/dfx/src/config/dfinity.rs#L491), which works its way up the directory tree until it finds dfx.json.

See also `Config::get_project_root()`, which returns the directory that contains dfx.json.  Using this would be in alignment with the ticket description
> just having "script.sh" as the custom builder command should run script.sh which lies just next to dfx.json.

Overall I'd suggest building an absolute path and trying to execute that, rather than prepending `./`.  What would be wrong with something like this?
```
    let command_path = if command_path.is_relative() {
        config.get_project_root().join(command_path)
    } else {
        command_path
    };
```

src/dfx/src/lib/builders/custom.rs
```diff
@@ -189,8 +189,16 @@ impl CanisterBuilder for CustomBuilder {
...
+    let mut cmd = if arguments.is_empty()
+        && command_path.components().count() == 1
```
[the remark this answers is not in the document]
adamspofford-dfinity: Yes. Only in the specific case of the path having zero slashes in it does this kick in.

src/dfx/src/lib/builders/custom.rs
adamspofford-dfinity: I don't think this interacts with parent directories at all. As for an absolute path, that's fine if you say it's fine, I was just trying to change as little about the expected behavior as possible

ericswanson-dfinity: 1. If we're going to assume that a filename refers to a file in a particular directory, check that it exists, and then execute it if so, then I think we should verify its existence and execute it by absolute path, rather than by prepending `./` to it. 

2. dfx is meant to work if run within any subdirectory of a project.  Though there are bugs that stop this from working, we don't have to add more today.  If we use `Config::get_project_root()` here, we don't introduce a new one.  This would also be consistent with this request from the ticket:

> just having "script.sh" as the custom builder command should run script.sh which lies just next to dfx.json.

src/dfx/src/lib/builders/custom.rs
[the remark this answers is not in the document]
ericswanson-dfinity: Okay.  Is it expected that both of the above declarations would successfully call the script in question?

src/dfx/src/lib/builders/custom.rs
adamspofford-dfinity: Oh, I see what you mean. So then should *any* relative script path be canonicalized against get_project_root before being passed to Command::new?

adamspofford-dfinity: I see what you mean, it's just weird to change the precise meaning of the path in the single use case. Anyway, if it's next to dfx.json then the original problem doesn't trigger from a subdirectory; if it is to be executable from a subdirectory then we would need to call Command::current_dir for the script, which requires canonicalizing *all* commands, not just in this special case. Should we do this?

ericswanson-dfinity: Yeah, I think we should canonicalize all commands as relative paths from the project root, and specify Command::current_dir as the project root.  In the common case (where dfx is being run from the project root), that will be like canonicalizing from the current directory, and using the current directory for execution.  What do you think?

adamspofford-dfinity: I think that's a great idea.

src/dfx/src/lib/builders/custom.rs
```diff
@@ -190,12 +192,16 @@ impl CanisterBuilder for CustomBuilder {
     }
 }
 
-fn run_command(args: Vec<String>, vars: &[super::Env<'_>]) -> DfxResult<()> {
+fn run_command(args: Vec<String>, vars: &[super::Env<'_>], cwd: &Path) -> DfxResult<()> {
     let (command_name, arguments) = args.split_first().unwrap();
-
-    let mut cmd = std::process::Command::new(command_name);
+    let canonicalized = which::which_in(command_name, env::var_os("PATH"), cwd)
```
ericswanson-dfinity, line 197: If a project's dfx.json has `"build": "build.sh"`, and there is a `build.sh` in that project, and the user has `$HOME/bin` in their path, and `$HOME/bin/build.sh` exists, which `build.sh` should we expect to be called?

ericswanson-dfinity: And, whichever is the answer, let's have tests that demonstrate the behavior.

adamspofford-dfinity: We would expect the local command to be preferred for reproducibility across machines.

ericswanson-dfinity: (I ask because it looks like the code would first search $PATH)

ericswanson-dfinity (changes requested):
Looks good.  Would you please add a test that verifies the expected behavior with respect to custom build commands found both in the project and on the path?

PR #2269, opened 2022-06-16
chore: Replace chrono with time

adamspofford-dfinity:
Chrono's dependency on time@0.1 invokes RUSTSEC-2020-0071. We have an ignore rule on the cargo-audit step because of the vulnerability's ubiquity and near-impossibility to remove while keeping local offset functionality, but fortunately we don't actually use the local offset anywhere and deal exclusively in UTC. This  PR replaces all usages of chrono with time 0.3, which resolves the rustsec advisory in addition to having a generally better API.


ericswanson-dfinity (approved):
Fantastic!

PR #2281, opened 2022-06-20
chore: Remove `extra` field from config

adamspofford-dfinity:
As it makes everything easier, I have also taken the opportunity to replace the truly ancient crate humanize_rs with humantime and byte-unit, replace a few more Strings with PathBufs, and remove some of the more object-oriented design in the canister info/builder system.


src/dfx/src/lib/canister_info.rs
```diff
@@ -273,30 +228,46 @@ impl CanisterInfo {
...
+                .map(|x| x.get_output_idl_path().to_path_buf()),
+            CanisterTypeProperties::Assets { .. } => self
+                .as_info::<AssetsCanisterInfo>()
+                .map(|x| x.get_output_idl_path().to_path_buf()),
+            CanisterTypeProperties::Rust { .. } => self
+                .as_info::<RustCanisterInfo>()
+                .map(|x| x.get_output_idl_path().to_path_buf()),
         }
+        .ok()
+        .or_else(|| self.remote_candid.clone())
```
ericswanson-dfinity, line 246: Looks like we lost the equivalent of the `self.get_extra_optional("candid")` lookup before falling back to the `remote_candid` field.  This was added recently to handle the case where the type lookup failed.  Maybe @sesi200 will have more insight.

ericswanson-dfinity (changes requested):

src/dfx/src/config/dfinity.rs
```diff
@@ -61,18 +64,73 @@ pub const DEFAULT_IC_GATEWAY_TRAILING_SLASH: &str = "https://ic0.app/";
...
+    pub args: Option<String>,
+
+    #[serde(default)]
+    pub initialization_values: InitializationValues,
+
+    #[serde(default)]
+    pub dependencies: Vec<String>,
+
+    #[serde(default)]
+    pub frontend: bool,
```
sesi200, line 84:
```suggestion
    #[serde(default)]
    pub is_frontend: bool,
```
That or a comment explaining what it does seems appropriate

src/dfx/src/commands/remote/generate_binding.rs
```diff
@@ -56,25 +53,25 @@ pub fn exec(env: &dyn Environment, opts: GenerateBindingOpts) -> DfxResult {
...
                     continue;
                 }
-                if main_path.exists() {
+                if main.exists() {
                     if opts.overwrite {
                         info!(
                             log,
                             "Overwriting main file {} of canister {}.",
-                            main,
+                            main.display(),
```
sesi200, line 61: I think we default to `to_string_lossy` in this codebase (for no major reason I can remember).

src/dfx/src/lib/canister_info.rs
sesi200: This was added as a fallback when no canister type can be inferred ([Jira here](https://dfinity.atlassian.net/browse/SDK-506), PR #2262). Since we now handle this with exhaustive pattern matching and an option will always be chosen it should be fine to remove this branch.

Thinking a bit more I feel that we should not even rely on `get_output_idl_path` as the only way to determine the candid to use in `commands/canister/call.rs`. Using this as a fallback for OUTPUT idl sounds like a bug waiting to happen. If we refactor this to use proper logic (`candid` or `remote/candid` if available, otherwise `output_idl_path`) in `call.rs`, `get_output_idl_path` could get rid of the `.or_else` branch.

src/dfx/src/config/dfinity.rs
adamspofford-dfinity: I'm trying to break absolutely nothing (see PossiblyStr) - we can evolve the config to make more sense in a future PR

ericswanson-dfinity (approved):

PR #2290, opened 2022-06-24
chore: Support local CMC in e2e tests

adamspofford-dfinity:
This patches the cycles minting canister before installing it as part of the ledger package, so the XDR:ICP rate can be set. The method used is copied from nns-dapp. Unfortunately *creating* a canister requires a subnet, which the local replica doesn't have, so for now we can only test cycle *top-up*.

Resolves SDK-514.


sesi200:
LGTM, but do we have a strategy to keep this at least somewhat up-to-date?

adamspofford-dfinity:
We would simply update the IC commit here when we do it elsewhere as well.

scripts/update-replica.sh
```diff
@@ -26,4 +26,9 @@ niv update sandbox-launcher-x86_64-linux -a rev=$SHA
 echo "Writing asset sources"
 ./scripts/write-dfx-asset-sources.sh
 
+read -r -p "Update the ledger for e2e tests? (y/n)" yn
+if [[ ! $yn =~ [nN] ]]; then
+    ./scripts/update-nns.bash "$SHA"
```
sesi200, line 31: Does this also update nns-canister.patch? I don't see how the patch file is staying updated.

sesi200: I think this is it: https://github.com/dfinity/nns-dapp/blob/main/e2e-tests/scripts/nns-canister.patch

sesi200 (approved):

ericswanson-dfinity (approved):

PR #2297, opened 2022-06-27
feat: Enable threshold ecdsa signature

ninegua:
Turn on ECDSA signature signing by default with a test key named "Secp256k1:dfx_test_key".

# Description

This enables developers to use threshold ECDSA system APIs when deploying code in a local dev environment.

# How Has This Been Tested?

It has been tested manually to be working.

# Checklist:

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -5,6 +5,11 @@
 
 == DFX
 
+=== feat: Enable threshold ecdsa signature
+
+ECDSA signature signing is now enabled by default in new projects, or by running `dfx start --clean`.
+A test key id "Secp256k1:dfx_test_key" is ready to be used by locally created canisters.
+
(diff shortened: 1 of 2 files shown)


ninegua:
> with a caveat that we must wait for a full DKG interfaceinterval (500 blocks in the default setting) before starting to use the key.

This will be fully addressed in https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/6069. So the situation now is:

1. With this PR, users will be able to start using ECDSA key with id `Secp256k1:dfx_test_key` in dfx, but need to wait until a fresh `dfx start` has gone through the first 1000 rounds (no good indicator to find out, but roughly 1000 seconds).
2. Once MR 6069 is merged and dfx gets the the replica from a release afterwards, users will be able to use the key immediately after `dfx start`.

ericswanson-dfinity (approved):

PR #2305, opened 2022-06-29
chore: Update deny.toml

adamspofford-dfinity:
Our denylist includes GPL, but not AGPL or LGPL. We would be better off getting rid of it and sticking with cargo-deny's built-in copyleft list. 

`abnf` needs an explicit exception because it was changed later and is a build dependency.


raymondk:
Closed this PR because it's old.

PR #2310, opened 2022-07-04
feat: Allow users to configure logging level of bitcoin adapter

ielashi:
# Description

The bitcoin adapter's logging can be very verbose if `debug` logging is enabled, making it difficult to make sense of what's going on. On the other hand, these logs are useful for triaging problems.

To get the best of both worlds, this commit adds support for an additional configuration option in `dfx.json`:

```
  "defaults": {
    "bitcoin": {
      "enabled": true,
      "nodes": ["127.0.0.1:18444"],
      "log_level": "info" <------- users can now configure the log level
    },
```

By default, a log level of "info" is used, which is relatively quiet. Users can change it to "debug" for more verbose logging.

# How Has This Been Tested?

I ran `dfx start` with various configurations of `dfx.json` and verified from the logs that the bitcoin adapter gets initialized with the expected log level.

# Checklist:

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -5,6 +5,20 @@
 
 == DFX
 
+=== feat: Allow users to configure logging level of bitcoin adapter
+
+The bitcoin adapter's logging can be very verbose if debug logging is enabled, making it difficult to make sense of what's going on. On the other hand, these logs are useful for triaging problems.
+
+To get the best of both worlds, this commit adds support for an additional configuration option in dfx.json:
+
+    "bitcoin": {
+      "enabled": true,
+      "nodes": ["127.0.0.1:18444"],
+      "log_level": "info" <------- users can now configure the log level
+    }
+
+By default, a log level of "info" is used, which is relatively quiet. Users can change it to "debug" for more verbose logging.
+
 === chore: update Candid UI canister with commit bffa0ae3c416e8aa3c92c33722a6b1cb31d0f1c3
 
 This includes the following changes:
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -9,7 +9,7 @@
 
 The bitcoin adapter's logging can be very verbose if debug logging is enabled, making it difficult to make sense of what's going on. On the other hand, these logs are useful for triaging problems.
 
-To get the best of both worlds, this commit adds support for an additional configuration option in dfx.json:
+To get the best of both worlds, this release adds support for an additional configuration option in dfx.json:
 
     "bitcoin": {
       "enabled": true,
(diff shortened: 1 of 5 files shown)


CHANGELOG.adoc
```diff
@@ -5,6 +5,20 @@
 
 == DFX
 
+=== feat: Allow users to configure logging level of bitcoin adapter
+
+The bitcoin adapter's logging can be very verbose if debug logging is enabled, making it difficult to make sense of what's going on. On the other hand, these logs are useful for triaging problems.
+
+To get the best of both worlds, this commit adds support for an additional configuration option in dfx.json:
```
sesi200, line 12:
```suggestion
To get the best of both worlds, this release adds support for an additional configuration option in dfx.json:
```

src/dfx/src/lib/bitcoin/adapter/config.rs
```diff
@@ -35,6 +36,45 @@ impl Default for IncomingSource {
     }
 }
 
+/// Represents the log level.
+#[derive(Clone, Debug, Serialize, Deserialize, Copy, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum Level {
```
sesi200, line 42:
```suggestion
/// Represents the log level of the bitcoin adapter.
#[derive(Clone, Debug, Serialize, Deserialize, Copy, PartialEq, Eq)]
#[serde(rename_all = "snake_case")]
pub enum BtcLogLevel {
```
This is only the BTC adapter log level, right? And not a general replica level.

src/dfx/src/config/dfinity.rs
```diff
@@ -26,6 +27,7 @@ const EMPTY_CONFIG_DEFAULTS: ConfigDefaults = ConfigDefaults {
 const EMPTY_CONFIG_DEFAULTS_BITCOIN: ConfigDefaultsBitcoin = ConfigDefaultsBitcoin {
     enabled: false,
     nodes: None,
+    log_level: Level::Info,
```
sesi200, line 30:
```suggestion
    log_level: Level::default(),
```

sesi200 (changes requested):

src/dfx/src/lib/bitcoin/adapter/config.rs
ielashi: It is, but it is implied that this is the case because it's defined in the `bitcoin::adapter::config` module. Otherwise, we'd need to prefix `BtcAdapter` to every struct in this file. In any case, I updated the struct name to what you suggested.

src/dfx/src/config/dfinity.rs
ielashi: I don't think this is possible. We're declaring a constant, and we can only use a constant function here. The default method is not a constant function.

sesi200: ah, makes sense

sesi200 (approved):
LGTM

ericswanson-dfinity (approved):

PR #2311, opened 2022-07-04
feat: rename canisters in new projects to <project>_frontend and <project>_backend

dprats:
This PR creates the following structure for new projects created via `dfx new...`

Suppose, user ran `dfx new hello`...User will get two smart contract canisters:

* `hello_backend`
* `hello_frontend`


ericswanson-dfinity (commented):
I wonder if the rust project backend .did file, which used to be `<project>/<project>.did` should be `<project>_backend/<project>_backend.did`

Screen Shot 2022-07-05 at 4 59 39 PM

ericswanson-dfinity (approved):

PR #2335, opened 2022-07-11
feat: canister metadata command

ericswanson-dfinity:
# Description

Adds `dfx canister metadata <canister> <metadata name>` command, for example `dfx canister metadata something_backend candid:service` to display the candid service definition for the backend canister.

Open questions:
1. In addition to general utility, this is intended to aid deprecation of the `/_/candid` endpoint.  That endpoint, however, has options for formatting the candid definition in either `.js` or `.ts` format.  Should this command have some option to perform that transformation, or is there some other tool that is better suited to the purpose?  What is an example of a consumer of that functionality?

Fixes https://dfinity.atlassian.net/browse/SDK-346

# How Has This Been Tested?

Added e2e tests

# Checklist:


chenyan-dfinity:
> Should this command have some option to perform that transformation, or is there some other tool that is better suited to the purpose? What is an example of a consumer of that functionality?

Either Candid UI or `didc` can handle the transformation. The `metadata` command should not handle the transformation as it's too specific to the Candid metadata. Also the `_/candid` transformation is only used by Candid UI, which is deprecated now already.

src/dfx/src/commands/canister/metadata.rs
```diff
@@ -0,0 +1,41 @@
...
+    let canister_id = Principal::from_text(callee_canister)
+        .or_else(|_| canister_id_store.get(callee_canister))?;
+
+    fetch_root_key_if_needed(env).await?;
+    let metadata = agent
+        .read_state_canister_metadata(canister_id, &opts.metadata_name, false)
+        .await
+        .with_context(|| format!("Failed to read controllers of canister {}.", canister_id))?;
+
+    stdout().write_all(&metadata)?;
```
chenyan-dfinity, line 38: Not sure the best way to display this. Some of the metadata may not be printable ascii. Maybe we can check if `metadata` is printable, if not, we display the hex instead.

sesi200: Would it make sense to have an `--outfile` flag instead/in addition?

ericswanson-dfinity: @chenyan-dfinity @sesi200 I wondered about these too.  Not sure of the best way either.

The thinking was that by writing the raw data bytes to stdout, they can be written to a file or piped to another program.

While the metadata may not be printable ascii, what if the developer wants to run the actual contents through `hexdump` or some other tool?  If `dfx canister metadata ... | hexdump -C`, and the metadata is not entirely printable ascii, what would the expected behavior be?

What would the proposed rules be?  Would they more clear than always writing as-is to stdout?  The below seem overly complicated to me.
- if `--outfile` specified, write data to specified file as-is.  `--outfile -` means write data as-is to stdout.
- If no `--outfile` specified, and metadata is printable ascii, write that to stdout.
- If no `--outfile` specified, and metadata is not all printable ascii, then write an explanation to stderr and display as hex.

sesi200: I have no strong opinion either way and have no problem merging this as-is. I was mostly thinking about invoking dfx from somewhere that doesn't have such trivial redirect mechanics, but I might just be overthinking it.

My approach would have been to print to `--outfile` if specified, otherwise print to stdout. Reason being that I'd want to give a way to not see garbage printed if it is expected.

src/dfx/src/commands/canister/metadata.rs
```diff
@@ -0,0 +1,41 @@
...
+use crate::lib::models::canister_id_store::CanisterIdStore;
+use crate::lib::root_key::fetch_root_key_if_needed;
+use crate::Environment;
+
+use anyhow::{anyhow, Context};
+use clap::Parser;
+use ic_types::Principal;
+use std::io::{stdout, Write};
+
+/// Sign a canister call and generate message file.
```
sesi200, line 11:
```suggestion
/// Displays metadata in a canister.
```

sesi200 (approved):

src/dfx/src/commands/canister/metadata.rs
sesi200: I went ahead and merged this already

ericswanson-dfinity: Thanks for noticing this!

PR #2338, opened 2022-07-12
chore: Remove `dfx config` from scripts and documentation

adamspofford-dfinity:
This removes the documentation references to `dfx config`, and replaces every reference to it in the CI scripts with an equivalent `jq`. 

Given the number of such commands we may want to replace the `cat <<<` version with moreutils' `sponge` in a future PR.

Not to be merged until 0.11 is finalized.


adamspofford-dfinity:
I was thinking it made sense for it to be deprecated for one release, undocumented for a second, and then removed in the third.

PR #2345, opened 2022-07-14
fix: improve responsiveness of `greet` method in Motoko project template

matthewhammer:
# Description

`greet` is marked as an update call, but it performs no state updates.  It can be faster as a query call.


ggreif (approved):

ericswanson-dfinity (dismissed):
Would you please add an entry to the changelog?

ericswanson-dfinity (approved):

PR #2351, opened 2022-07-15
fix: Make build field optional

adamspofford-dfinity:
Prior to 494c566, someone could leave off the `build` field of the config for a custom canister and never receive an error about it if they never ran `dfx build`. This restores that behavior by making a missing `build` field equivalent to `build: []`.


ericswanson-dfinity (commented):
Please add a note to the changelog

ericswanson-dfinity (approved):

PR #2355, opened 2022-07-15
fix: Allow dependencies to not take arguments if not being installed

adamspofford-dfinity:
Fixes SDK-524, SDK-526. 

Currently, if `dfx deploy` deploys multiple canisters due to dependency declarations, it will require arguments for all of them, even if the install step ends up getting skipped for one reason or another, such as being already installed. This fixes this bug by lazily producing the arguments only upon actual installation.


ericswanson-dfinity (changes requested):
Please add a note in the changelog

ericswanson-dfinity (approved):

PR #2361, opened 2022-07-20
chore: add a comprehensive dfx json reference

sesi200:
# Description

Adds `dfx schema`, a command that prints the current schema for `dfx.json`. Also added a workflow to keep a copy of the output up to date at `docs/dfx-json-schema.json` so that the docs can reference the current schema.

First part of [SDK-287](https://dfinity.atlassian.net/browse/SDK-287)

# How Has This Been Tested?

- manually
- workflow checks that `docs/dfx-json-schema.json` is up to date
- e2e checks that `dfx schema` produces valid json

# Checklist:


smallstepman:
I think building a separate workflow that updates the JSON schema file will suffice. I'm not sure how having e2e test would then be useful. The workflow could run on each push for each PR, and simply do 
```
if ! diff -q old.json new.json &>/dev/null; then
  git add new.json
  git commit -m "" 
  git push
fi
```
The workflow could be made `:required` so it is impossible to merge without completing it first.

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,45 @@
+name: Update Docs
+on:
+  pull_request:
```
smallstepman, line 3:
```suggestion
  pull_request
```
the `:` is unnecessary here

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  update_dfx_json_schema:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        rust: ["1.60.0"]
+        os: [ ubuntu-20.04 ]
+    steps:
+      - uses: actions/checkout@v2
```
smallstepman, line 14:
```suggestion
      - uses: actions/checkout@v3.0.2
```
what do you think about using the latest?

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.ref }}
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
smallstepman, line 24:
```suggestion
```
did you manage to get the cache working? otherwise, maybe we should drop it?

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Install Rust
```
smallstepman, line 25: more of a fyi, cause I'm not 100% certain here (help @adamspofford-dfinity), but I think this step is unnecessary since we have `rust-toolchain.toml` file - cargo should be able to use the correct version here

to verify, you might want to try `rustc --version` instead and see if its `1.60.0`

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Check cargo build
+        run: cargo build
+      - name: Show download worked
+        run: cargo run -- --version
+      - name: Update docs/dfx-json-schema.json
+        run: |
+          cargo run -- schema --outfile docs/dfx-json-schema.json
+          cat docs/dfx-json-schema.json
+          if ! git diff-index --quiet HEAD; then
```
smallstepman, line 39:
```suggestion
          if ! git diff-index --quiet HEAD -- docs/dfx-json-schema.json; then
```
have you tried this one?

.github/workflows/update-docs.yml
smallstepman, line 3:
```suggestion
  pull_request
```
the `:` is unnecessary

smallstepman:
ooof chicken and egg problem https://github.com/dfinity/sdk/runs/7451711903?check_suite_focus=true
you probably want to `touch docs/dfx-json-schema.json` and git add it.

smallstepman (approved):

PR #2369, opened 2022-07-22
feat: move all the args to the end

### LINKED ISSUE #835: Fix the awkward syntax for dfx commands
The syntax for some of the dfx canister commands is rather awkward and inconsistent. I can think a few different ways to make the commands lines more naturally readable. For example, one simple and straightforward approach  might be to remove "canister" and use command lines like this:
```
dfx register [--network network-name] --all|--canister-name (replacing canister create)
dfx build [--netwokl network-name] --all|--canister-name (adding back all or name)
dfx deploy [--network network-name] --all|--canister-name (replacing canister install)
dfx call [--network network-name] --canister-name --method-name
dfx ping [--network network-name] --all|--canister-name (it would be cool in the future to see 
if all my canisters are running or a particular one is stopped)
dfx request-status
dfx start | dfx stop | dfx config
``` 
etc could all rename as is as local commands ...

I'm not a fan of inserting the network option in the middle of the noun (canister) and verb/action (create) and it makes the help awkward and build doesn't fallow the same pattern.
Even something like
`dfx connect --network name register|build|deploy --all `or `dfx canister register|build|deploy --network name --all might` be better than what we have at the moment.

adamspofford-dfinity:
The standard format of commands is `cmd subcmd1 subcmd2 --namedflag1 --namedflag2`. In `dfx`, it is instead `cmd --namedflag1 subcmd1 --namedflag2 subcmd2`. This is a source of confusion for new users and experienced users alike. This PR moves all the flags to the end of the command; the new valid syntax is `dfx canister create --all --identity alice --network ic --wallet "$WALLET"`.

Implements SDK-604.

Fixes #835.


ericswanson-dfinity (approved):

ericswanson-dfinity:
Please wait to merge this until after dfx 0.11.1 is released.  We'll include this breaking change in dfx 0.12.0.

CHANGELOG.adoc
```diff
@@ -23,6 +23,19 @@ always writes them under the .dfx/ directory.
...
+[source, bash]
+----
+dfx --identity alice canister --network ic --wallet "$WALLET" create --all
+----
+This command should now read:
+[source, bash]
+----
+dfx canister create --all --network ic --wallet "$WALLET" --identity alice
+----
+
```
ericswanson-dfinity, line 38: This should go in the `UNRELEASED` section above

adamspofford-dfinity: Good catch!

PR #2375, opened 2022-07-25
fix: "canister alias not defined" in the Motoko language server

rvanasa:
# Description

This PR is a step towards fixing dfinity/vscode-motoko#25.

- Calling `dfx _language-service` now passes `--actor-idl` and `--actor-alias` flags to `mo-ide` [as was previously intended and documented](https://github.com/dfinity/motoko/blob/master/design/DFX-Interface.md#invoking-the-ide). 
- `dfx build` copies Candid files to `.dfx/(network)/lsp` for use by the language server protocol (LSP). In the future, this directory could be dynamically updated by `dfx` to inform `mo-ide` of interface changes (as outlined in #472). 

# How Has This Been Tested?

Using the executable created by `cargo build` and after updating the version in `dfx.json`:

- `.dfx/(network)/lsp/` directory generates as expected when running `dfx build`
- Fixes the VSCode error reported in dfinity/vscode-motoko#25

# Checklist:

- [x] I have made corresponding changes to the documentation ([already documented](https://github.com/dfinity/motoko/blob/master/design/DFX-Interface.md#invoking-the-ide)).


src/dfx/src/commands/language_service.rs
```diff
@@ -40,7 +44,35 @@ pub fn exec(env: &dyn Environment, opts: LanguageServiceOpts) -> DfxResult {
...
-        let package_arguments = package_arguments::load(env.get_cache().as_ref(), packtool)?;
+
+        let mut package_arguments = package_arguments::load(env.get_cache().as_ref(), packtool)?;
+
+        // Include actor alias flags
+        let canister_names = config
+            .get_config()
+            .get_canister_names_with_dependencies(None)?;
+        let agent_env = create_agent_environment(env, None /* opts.network */)?;
+        let canister_id_store = CanisterIdStore::for_env(&agent_env)?;
```
ericswanson-dfinity, line 55: We can use the Environment here, already in scope.
```suggestion
        let canister_id_store = CanisterIdStore::for_env(env)?;
```

src/dfx/src/commands/language_service.rs
```diff
@@ -40,7 +44,35 @@ pub fn exec(env: &dyn Environment, opts: LanguageServiceOpts) -> DfxResult {
...
+                .or_else(|_| canister_id_store.get(&canister_name[..]))
+            {
+                Ok(canister_id) => package_arguments.append(&mut vec![
+                    "--actor-alias".to_owned(),
+                    canister_name,
+                    Principal::to_text(&canister_id),
+                ]),
+                Err(_) => (),
+            };
+        }
```
ericswanson-dfinity, line 67: - no need for `.into_iter()` in the for loop
- `get_canister_names_with_dependencies` returns canister names, not principal IDs, so we don't need to try `Principal::from_text`
- `&canister` covers it, no need for `[..]`, unless I am missing something
- addresses lint warning `error: you seem to be trying to use 'match' for destructuring a single pattern. Consider using 'if let'`

```suggestion
        for canister_name in canister_names {
            if let Ok(canister_id) = canister_id_store.get(&canister_name)
            {
                package_arguments.append(&mut vec![
                    "--actor-alias".to_owned(),
                    canister_name,
                    Principal::to_text(&canister_id),
                ])
            };
        }
```

ericswanson-dfinity (changes requested):

src/dfx/src/commands/language_service.rs
rvanasa: Using `env` instead of `agent_env` causes the following error at runtime:
`'internal error: entered unreachable code: NetworkDescriptor only available from an AgentEnvironment'`
Is there a way to avoid this using the provided (non-Agent)Environment? 

Thank you for taking a look at this PR.

src/dfx/src/commands/language_service.rs
rvanasa: Thanks, looks good. It's been a few years since I've used Rust (and it's my first time in the DFX repo), so I appreciate the suggestions.

src/dfx/src/commands/language_service.rs
ericswanson-dfinity: Oh yes, you are right.  I've submitted https://github.com/dfinity/sdk/pull/2376, which will provide `CanisterIdStore::new` for this purpose.

src/dfx/src/commands/language_service.rs
```diff
@@ -40,7 +44,38 @@ pub fn exec(env: &dyn Environment, opts: LanguageServiceOpts) -> DfxResult {
...
+        let canister_names = config
+            .get_config()
+            .get_canister_names_with_dependencies(None)?;
+        let network_descriptor = get_network_descriptor(
+            env.get_config(),
+            None, /* opts.network */
+            LocalBindDetermination::ApplyRunningWebserverPort,
+        )?;
+        let canister_id_store =
+            CanisterIdStore::new(&network_descriptor, env.get_config(), env.get_temp_dir())?;
```
rvanasa, line 60: Is this what you were imagining for `CanisterIdStore::new()`? Besides still having to call `get_network_descriptor()` because of the non-AgentEnvironment, everything else seems cleaner. 

Also, the language server is no longer resolving canister ids as of merging #2376 (with or without the AgentEnvironment). Everything still works as expected on the master branch. Do you have a guess to what could be causing the empty CanisterIdStore? (@ericswanson-dfinity)

ericswanson-dfinity: Yes, this is what I expected it would look like.  At first glance, I don't see why canister id lookup would be failing.

ericswanson-dfinity (dismissed):

ericswanson-dfinity (approved):

PR #2393, opened 2022-08-01
fix: do not add candid:service metadata to custom canisters

ericswanson-dfinity:
# Description

It turns out it's surprising that dfx adds custom metadata to custom canisters.  For the sake of build reproducibility, let's have dfx leave custom canisters alone.  The build step can use ic-wasm if appropriate, instead.

Fixes https://dfinity.atlassian.net/browse/SDK-644

# How Has This Been Tested?

Updated existing e2e tests.

# Checklist:

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -5,6 +5,10 @@
 
 == DFX
 
+=== fix: dfx no longer adds candid:service metadata to custom canisters
+
+Instead, you can use https://github.com/dfinity/ic-wasm in your custom canister's build step if necessary.
+
 === feat!: removed the internal webserver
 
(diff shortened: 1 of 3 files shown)


krpeacock:
I'm not certain we should act on this immediately - it may have been unexpected to some internal folks, but I'd like input from the community on what their expected behavior is. Particularly @lastmjs

lastmjs:
We haven't played with this yet, but off the top of my head I was thinking it would be a good idea for the metadata to be generated automatically for Azle canisters, just like Rust and Motoko canisters. I would prefer to not have to add extra plumbing to the commands we ask our developers to use, we're trying to allow them to use plain `dfx` commands like `dfx deploy` as much as possible. FYI all Azle canisters are custom canisters.

ericswanson-dfinity:
We can add an option, but if we do, and then later implement wasm metadata configuration in dfx.json, then we'll have to remove it.  This is why the last attempt made it so the candid:service metadata is only added if there is a build step.

PR #2407, opened 2022-08-03
fix: Wait for install to be reported in state tree

adamspofford-dfinity:
Fixes SDK-544.

No way of e2e testing this for now unfortunately. It successfully installed a canister on mainnet without a spurious error, but to see if it actually fixed the problem will require waiting for it to flake and seeing that it doesn't, without replica throttling functionality.


src/dfx/src/lib/operations/canister/install_canister.rs
```diff
@@ -136,7 +140,47 @@ pub async fn install_canister(
...
+                            env.get_logger(),
+                            "Waiting for module change to be reflected in system state tree..."
+                        )
+                    }
+                    sleep(Duration::from_millis(500)).await;
+                } else {
+                    bail!("The reported module hash is neither the existing module or the new one. \
+                        It has likely been modified while this command is running. \
+                        The state of the canister is unknown. \
+                        For this reason, no post-installation tasks have been run, including asset uploads.")
```
ericswanson-dfinity, line 168: Can we display the three module hashed involved here?  That will make it easier for the developer to follow up with `dfx canister info` and troubleshoot.

src/dfx/src/lib/operations/canister/install_canister.rs
```diff
@@ -136,7 +140,47 @@ pub async fn install_canister(
...
+            }
+            Err(AgentError::LookupPathAbsent(_) | AgentError::LookupPathUnknown(_)) => {
+                times += 1;
+                if times % 4 == 0 {
+                    info!(
+                        env.get_logger(),
+                        "Waiting for module change to be reflected in system state tree..."
+                    )
+                }
+                sleep(Duration::from_millis(500)).await;
```
ericswanson-dfinity, line 179: What would you think about using a Waiter here, with some finite timeout?

ericswanson-dfinity (changes requested):

src/dfx/src/lib/operations/canister/install_canister.rs
```diff
@@ -113,14 +114,17 @@ pub async fn install_canister(
...
-        && wasm_module_already_installed(&wasm_module, installed_module_hash.as_deref())
+        && installed_module_hash
+            .as_deref()
+            .map_or(true, |old_hash| old_hash == new_hash)
         && !upgrade_unchanged
     {
         println!(
             "Module hash {} is already installed.",
-            hex::encode(installed_module_hash.unwrap())
+            hex::encode(installed_module_hash.as_ref().unwrap())
```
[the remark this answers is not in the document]
adamspofford-dfinity: You're right, I got the condition mixed up.

src/dfx/src/lib/operations/canister/install_canister.rs
adamspofford-dfinity: I'm not really sure how Waiter works, but it looks blocking, which doesn't play well with async.

adamspofford-dfinity: Wait, is that how we managed to starve the single-threaded web server however long ago?

ericswanson-dfinity: That's why we use Garcon, which provides async_await().  Example:

https://github.com/dfinity/sdk/blob/master/src/dfx/src/commands/ledger/mod.rs#L165

adamspofford-dfinity: Hmm. Some of those just block the thread anyway and then return a ready future, and others spin a new thread and sleep it. Both of these things defeat all benefits of asynchrony.

ericswanson-dfinity (changes requested):
Please add an entry to the changelog

ericswanson-dfinity (approved):

PR #2409, opened 2022-08-04
chore: ledger account-id --of-canister also accepts principal

sesi200:
# Description

It is now possible to do e.g. `dfx ledger account-id --of-canister fg7gi-vyaaa-aaaal-qadca-cai` as well as `dfx ledger account-id --of-canister my_canister_name` when checking the ledger account id of a canister.
Previously, dfx only accepted canister aliases and produced an error message that was hard to understand.

# How Has This Been Tested?

Added e2e for account-id

# Checklist:


docs/cli-reference/dfx-ledger.md
```diff
@@ -45,7 +45,7 @@ You can use the following optional flags with the `dfx ledger account-id` comman
 |------------------------------|--------------------------------------------------------|
 | `-h`, `--help`               | Displays usage information.                            |
 | `-V`, `--version`            | Displays version information.                          |
-| `--of-canister <ALIAS>`      | Alias of the canister controlling the account          |
+| `--of-canister <ALIAS>`      | Alias or name of the canister controlling the account  |
```
ericswanson-dfinity, line 48: Maybe I'm missing something, but doesn't `--of-principal` already provide this functionality? 

And if it didn't, wouldn't we be making this accept either the name(alias) or the principal?

```suggestion
| `--of-canister <ALIAS>`      | Alias or principal of the canister controlling the account  |
```

src/dfx/src/commands/ledger/account_id.rs
```diff
@@ -15,7 +15,7 @@ pub struct AccountIdOpts {
     pub of_principal: Option<Principal>,
 
     #[clap(long, value_name = "ALIAS")]
-    /// Alias of the canister controlling the account.
+    /// Alias or name of the canister controlling the account.
```
ericswanson-dfinity, line 18:
```suggestion
    /// Alias or principal of the canister controlling the account.
```

ericswanson-dfinity (changes requested):

docs/cli-reference/dfx-ledger.md
sesi200: You're completely right with the wording. No idea what I was thinking there...

Yes, `--of-principal` has this functionality already, but as a user when I want the account id of a _canister_ principal, I first think of using the `--of-canister` since I'm thinking first and foremost of the string as representing a _canister_.

ericswanson-dfinity (approved):

PR #2413, opened 2022-08-05
feat(frontend-canister): impl `get` and `set` methods for asset properties

smallstepman:
# Description

Implements [SDK-817](https://dfinity.atlassian.net/browse/SDK-817) 

# How Has This Been Tested?
e2e
# Checklist:


src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -555,6 +555,33 @@ impl State {
...
+            .get(&key)
+            .ok_or_else(|| "asset not found".to_string())?;
+
+        Ok(AssetProperties {
+            max_age: asset.max_age,
+            headers: asset.headers.clone(),
+        })
+    }
+
+    pub fn set_asset_properties(&mut self, arg: SetAssetPropertiesArguments) -> Result<(), String> {
```
ericswanson-dfinity, line 571: Please add state machine tests for these.  The state machine tests should go into as much detail as possible.  For example:
- what happens if you pass Some for headers and None for max_age
- have a test that demonstrates that after you've set either to Some(), you cannot change either back to None (and: is this desired?)

ericswanson-dfinity (changes requested):

src/canisters/frontend/ic-certified-assets/assets.did
```diff
@@ -77,6 +77,12 @@ type StreamingStrategy = variant {
   };
 };
 
+type SetAssetPropertiesArguments = record {
+  key: Key;
+  max_age: opt nat;
+  headers: opt vec HeaderField;
+};
```
ericswanson-dfinity, line 84: Let's discuss what the semantics of this call should be, but afterwards, we should document those semantics here.

smallstepman: do you think something like this:
```
type SetAssetPropertiesArguments = record {
  key: Key;
  max_age: opt opt nat;
  headers: opt opt vec HeaderField;
};
```
would work better? `opt opt` would allow to:
- `null` - indicates the value should not change
- `opt null` - set the value to `None`
- `opt opt 5` - set the value to `Some(5)`

src/canisters/frontend/ic-certified-assets/src/types.rs
```diff
@@ -152,6 +152,6 @@ pub struct AssetProperties {
 #[derive(Clone, Debug, CandidType, Deserialize)]
 pub struct SetAssetPropertiesArguments {
     pub key: Key,
-    pub max_age: Option<u64>,
-    pub headers: Option<HashMap<String, String>>,
+    pub max_age: Option<Option<u64>>,
```
ericswanson-dfinity, line 155: Other "Option":
```
pub enum AssertProperty {
    MaxAge(Option<u64>),
    Headers(Option<HashMap<String, String>>),
}
pub struct SetAssetPropertiesArguments {
    pub key: Key,
    pub properties: Vec<AssetProperty>,
}
```

src/canisters/frontend/ic-certified-assets/assets.did
ericswanson-dfinity: This is what I would suggest.  What do you think?
```
pub enum AssetProperty {
    MaxAge(Option<u64>),
    Headers(Option<HashMap<String, String>>),
}
pub struct SetAssetPropertiesArguments {
    pub key: Key,
    pub properties: Vec<AssetProperty>,
}
```

smallstepman: I'm not sure -- the thing I don't like about this solution is it allows the user to send
```
SetAssetPropertiesArguments {
  key: "/index.html",
  properties: [MaxAge(Some(1)), MaxAge(Some(2)), MaxAge(Some(3))]
}
```
which is resolvable, but doesn't feel smooth to me

ericswanson-dfinity: Sure.  In this case, the answer to that is "don't do that."

smallstepman: what is in your view the advantage of using `Vec<Enum>` for this?

ericswanson-dfinity: For example, the compiler can ensure that the code handles all of the cases.  Not so with n Options.
But I think it's mostly aesthetics.  Imagine if in the future if there were 8 properties.
I think overall I don't like any of the solutions yet.  What would you think about `SetAssetMaxAge` and `SetAssetHeaders` ?

smallstepman: > For example, the compiler can ensure that the code handles all of the cases. Not so with n Options.
> ...
> Imagine if in the future if there were 8 properties.

Personally, I'd much rather have code poorly optimized by the compiler, than ask the user to spend x8 more cycles to update the properties of an asset. Of course, we can try guessing how often the user will try to update ALL properties (vs just one), but we won't know for sure until we have telemetry.

To be honest, I'm also not too concerned with aesthetics here - this is a utility method and not the core part of the frontend canister public interface, and notably, in 99% of the cases it will be utilized by `ic_asset::sync via dfx deploy`, and not via manual `dfx canister call` or similar

I'm also not in love with any of the solutions, but I like `opt opt` better than 
- `Vec<Enum>`, because vec enum has ambiguity (allows for dups, when dups are present order matters)
- having one `set` method for each asset property, because that burns more cycles 

Sadly I don't have better ideas, but there are other tasks requiring my focus, so unless you have some other ideas, I'd like to move forward with currently proposed implementation

ericswanson-dfinity: > code poorly optimized by the compiler

Nothing above has anything to do with compiler optimization.  Rather, that `match` can ensure that all possible cases are covered.

That said, just go ahead with the `opt opt` approach.

smallstepman: > Rather, that match can ensure that all possible cases are covered.

sorry, I don't think I follow, can you explain? I don't get how would that be different from:

https://github.com/dfinity/sdk/blob/e2e186dfb4c8da7b33443c0703da863fbfda9c69/src/canisters/frontend/ic-certified-assets/src/state_machine.rs#L577-L582

ericswanson-dfinity: Happy to.  Referring to the fact that the compiler will report an error if one skips an enum type in a `match`, whereas it won't notice a missing `if let Some(...)` statement for one of many `opt opt` fields.  Of course tests should catch this also, but that was what I was referring to by `the compiler can ensure that the code handles all of the cases.`

```
match (prop) {
    MaxAge(max_age) => asset.max_age = max_age,
    Headers(headers) => asset.headers = headers,
    Redirect(url) => asset.redirect = redirect, // compiler error to leave this line out
}
```
Anyway, there aren't any right answers here, and `opt opt` may even be better in terms of backwards compatibility in the interface.

e2e/tests-dfx/assetscanister.bash
```diff
@@ -726,3 +726,52 @@ CHERRIES" "$stdout"
...
+        "cache": { "max_age": 2000 },
+        "headers": { "x-key": "x-value" }
+      }
+    ]' > src/e2e_project_frontend/assets/.ic-assets.json5
+
+    dfx deploy
+    ID=$(dfx canister id e2e_project_frontend)
+    PORT=$(get_webserver_port)
+
+    assert_command dfx canister --network "http://localhost:$PORT" call "$ID" get_asset_properties '("/somedir/upload-me.txt")'
```
ericswanson-dfinity, line 749: What is the rationale for passing the network by port here?  If not passing it, or even passing it by name, and possibly passing the canister by name, will the output `record` contain the field names rather than numbers? (`headers` rather than `1_661_489_734`)
```suggestion
    assert_command dfx canister call e2e_project_frontend get_asset_properties '("/somedir/upload-me.txt")'
```

ericswanson-dfinity (changes requested):

e2e/tests-dfx/assetscanister.bash
smallstepman: the rationale: 🤦 

so, I believe I added `--network ....` because it didn't work otherwise -- not sure why it works now, perhaps someone fixed the default port in `agent-rs`?

after removing the --network flag I can see field names listed correctly in the response (as opposed to numbers "`1_661_489_734`") - why? what's the reason it makes such a difference?

also gonna change the call by canister name to call by `ID`

ericswanson-dfinity: The reason it makes a difference is dfx isn't finding a candid file when the network is passed by URL.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ## DFX
 
+### feat(frontend-canister): add `get_asset_properties` and `set_asset_properties` to frontend canister
+
```
ericswanson-dfinity, line 8: Please update the `Frontend canister` section in `Dependencies`

CHANGELOG.md
```diff
@@ -6,18 +6,29 @@
...
 
-### feat: write canister metadata sections for dfx pull
+It's now possible to get and set AssetProperties for assets in frontend canister, like so:
+```
+dfx canister call e2e_project_frontend set_asset_properties '( record { key="asset.txt"; max_age=opt(opt(5:nat64)); headers=opt(opt(vec{record {"new-key"; "new-value"}})) })'
+dfx canister call e2e_project_frontend get_asset_properties '("/asset.txt")'
+```
+If you wish to set either of properties to `null`, do the following:
+```
+dfx canister call e2e_project_frontend set_asset_properties '( record { key="asset.txt"; max_age=opt(null); headers=opt(null) })'
```
ericswanson-dfinity, line 16: Is this how we want to describe the usage in the changelog?  Wouldn't developers instead configure this through ic-asset.json?  Describing `canister call` as the method for setting this data will be confusing, in opinion.   Given that the sync process doesn't yet handle updates of this type of data, I would suggest to mention that these fields can be set in ic-asset.json but will only affect newly created files, until we implement proper synchronization handling.

In other words, at this point this is more like "support for future work."

smallstepman: right, I like this better

I guess I wanted to empower our users, but you may be right that it could cause some confusion

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 ## DFX
 
+### feat(frontend-canister): add `get_asset_properties` and `set_asset_properties` to frontend canister
+
+As part of creating the support for future work, it's now possible to get and set AssetProperties for assets in frontend canister. 
+This change will trigger the update process for frontend canister (new module hash: `0a064423e6d7b262f9c39a782135db4e6207829910e6f033077e72a2aaaee064`).
```
ericswanson-dfinity, line 10: Please update the relevant section in 'dependencies', rather than here.
```suggestion
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (changes requested):
If you didn't update the asset canister with the new `./scripts/update-frontend-canister.sh`, you'll need to do so now (and update the module hash again).

PR #2414, opened 2022-08-05
feat(frontend-canister): configurable (per-asset) traffic redirection

smallstepman:
# Description

- Adds a way for the end developer to configure per-asset HTTP redirects via `.ic-assets.json` file.
- `RedirectUrl` is only valid if at least one of `host` or `path` is defined (checked in `RedirectUrl::is_valid()`)
- It's possible to redirect the traffic to any other valid web address (while this will work on mainnet without issues, there are scenarios which on `localhost` are challenging to do, like for example, redirect-fetch picture coming from outside of `localhost` https://stackoverflow.com/questions/25277457/google-chrome-redirecting-localhost-to-https) 

# How Has This Been Tested?
- [x] unit tests
- [x] e2e tests

# Checklist:


src/canisters/frontend/ic-asset/src/plumbing.rs
```diff
@@ -277,12 +277,13 @@ async fn upload_content_chunks(
             create_chunk(canister_call_params, batch_id, data_chunk, semaphores).map_ok(
                 move |chunk_id| {
                     println!(
-                        "  {}{} {}/{} ({} bytes)",
+                        "  {}{} {}/{} ({} bytes){}",
```
smallstepman, line 280: added additional logging, e.g.
```
Uploading assets to asset canister...
Starting batch.
Staging contents of new and changed assets:
  /favicon.ico 1/1 (15406 bytes)
  /sample-asset.txt 1/1 (24 bytes)
  /index.js.map 1/1 (673524 bytes), with config:
    - header: Strict-Transport-Security: max-age=31536000; includeSubDomains
    - header: Referrer-Policy: same-origin
    - header: X-Content-Type-Options: nosniff
    - header: X-Frame-Options: DENY
    - header: X-XSS-Protection: 1; mode=block
    - cache: max_age = 3999
    - redirect: RedirectConfig { from: None, to: RedirectUrl { host: Some("hwvjt-wqaaa-aaaam-qadra-cai.ic0.app"), path: Some("/img/IC_logo_horizontal.svg") }, user_agent: None, response_code: 301 }

```

src/canisters/frontend/ic-certified-assets/assets.did
```diff
@@ -77,6 +87,12 @@ type StreamingStrategy = variant {
   };
 };
 
+type AssetProperties = record {
+  max_age: opt nat64;
+  headers: opt vec HeaderField;
+  redirect: opt AssetRedirect
```
smallstepman, line 93:
```suggestion
  redirect: opt vec AssetRedirect
```
What do you think about this? I think it would make sense to be able to define multiple redirect rules, tho that may affect performance when someone adds too many rules. But then, the only case were I think it would be useful is testing in local env, e.g.
```json
[
  {
    "match": "**/*", 
    "redirects": [
       {"from": {"host": "localhost:3333"}, "to":{"host": "somewhereelse.com"}},
       {"from": {"host": "raw.ic0.app"}, "to":{"host": "ic0.app"}},
    ]
  }
]
```

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -36,12 +44,137 @@ pub struct AssetEncoding {
...
+    pub(crate) path: Option<String>,
+}
+
+#[derive(Default, Clone, Debug, CandidType, Deserialize)]
+pub struct AssetRedirect {
+    pub from: Option<RedirectUrl>,
+    pub to: RedirectUrl,
+    pub user_agent: Option<Vec<String>>,
+    pub response_code: u16,
+}
```
smallstepman, line 59: I have had a hard time coming up with the right place to put these types (the choice between `state_machine.rs` and `types.rs`

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -36,12 +44,137 @@ pub struct AssetEncoding {
...
+                return false;
+            }
+            // TODO: performance hit for long lists. Consider using:
+            // - https://docs.rs/regex/latest/regex/struct.RegexSet.html
+            //   > The key advantage of using a regex set is that it will
+            //   > report the matching regexes using a single pass through
+            //   > the text. If one has hundreds or thousands of regexes
+            //   > to match repeatedly (like a URL router for a complex
+            //   > web application or a user agent matcher), then a regex set
+            //   > can realize huge performance gains.
```
smallstepman, line 90: The intended use case for this is to specify a list of web crawlers, to be able to redirect them efficiently. The list of all known crawlerss isn't exactly short https://www.useragentstring.com/pages/useragentstring.php?typ=Crawler , so I think it may have sense to implement this using `RegexSet`. I haven't checked the performance or anything, but I suspect there is some overhead involved in spawning such object as `RegexSet` if `user_agent` would be a short list. 

I did check if it's possible to somehow initialize such object client side, and (again)somehow transfer raw bytes which would serve as a blueprint for regex-automata, but couldn't find anything like that. It's possible my understanding about how regex engine works is off (so e.g. there is no step that transforms list of regexes into - as I called it - raw bytes blueprint)

I'd love to read your opinion

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -678,59 +823,113 @@ fn create_token(
...
+        self.headers
+            .iter()
+            .find_map(|(k, v)| k.eq_ignore_ascii_case(header_key).then(|| v))
     }
-    if let Some(arg_headers) = asset.headers.as_ref() {
-        for (k, v) in arg_headers {
-            headers.push((k.to_owned(), v.to_owned()));
+}
+
+impl HttpResponse {
```
smallstepman, line 854: I did a small refactor, and have put all the different HTTP responses under one umbrella

src/canisters/frontend/ic-certified-assets/src/tests.rs
```diff
@@ -476,3 +485,516 @@ fn supports_custom_http_headers() {
         response,
     );
 }
+
+#[cfg(test)]
+mod test_http_redirects {
```
smallstepman, line 490: I'd appreciate it if you could eyeball the test cases

smallstepman:
I think I need to add more details in changlog about how to use `redirect` field in `.ic-assets.json`

src/canisters/frontend/ic-certified-assets/assets.did
smallstepman: or the following use case:
- redirect all user-generated traffic from raw to non-raw
- redirect all bot-generated traffic from non-raw to raw

i think this would require two separate redirect rules

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -36,12 +44,192 @@ pub struct AssetEncoding {
...
+            }
+        }
+
+        Some(location)
+    }
+
+    fn redirect(&self, req: &HttpRequest) -> Option<HttpResponse> {
+        if let Err(e) = self.is_valid() {
+            return Some(HttpResponse::build_400(&e));
+        }
```
roman-kashitsyn, line 173: I think it's too late to check validity at this point. Ideally, we should validate that the configuration is consistent on batch commit.

smallstepman: validity is also being checked during batch commit
https://github.com/dfinity/sdk/pull/2414/files#diff-b24a9df9fbc0c1766d0396a9748c7b50b8dc4cebacb9f7230d42d096ed16754fR31-R65

Somewhere around this or next sprint, we will also add support for CRUD methods for properties for each asset in the asset canister. The line you've reviewed is a safety measure if someone were to attempt to change the properties of the asset and set an invalid redirect configuration; however, since there is no official support for mentioned CRUD methods, maybe it's a good idea to move this into "CRUD" PR - is the idea sound to you?

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -36,12 +44,192 @@ pub struct AssetEncoding {
...
+                if !self.matches_user_agent(&req_user_agent) {
+                    return None;
+                }
+            }
+        }
+
+        if let Some(location) = self.get_location_url(req) {
+            let req_host = req.get_header_value("Host");
+            if req_host.is_none() {
+                Some(HttpResponse::build_400("Host header is missing"))
```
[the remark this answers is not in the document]
smallstepman: Good question, they do, but I don't know how strong are the guarantees. I hope @Daniel-Bloom-dfinity will know for sure - are there cases where HTTP `host: <addr>` header is not being preserved in transport between boundary node and canister?

frederikrothenberger:
> My biggest concern is that there is no way to certify a redirect. There is no proof anywhere that redirects were indeed configured by the canister owner. A malicious replica could redirect all requests to an arbitrary malicious canister, defeating the purpose of certified assets.
> 
> I'm not sure we can ship such a feature before we have a V2 certification designed by @frederikrothenberger. And even that won't support regular expressions in paths.

Thanks @roman-kashitsyn for pinging me on this!

@smallstepman: Cool feature, unfortunately this won't work on mainnet. The reason is, that the service worker will block all redirects received by `http_request` candid call for security reasons.

However, `http_request_update` can be used until the [new asset certification](https://docs.google.com/document/d/1eS9iQE33ehCgSEsB1_42rfjGZnSF6PkBQfm9_EkutQs) is implemented. Please adapt your implementation accordingly. Also, this will incur a latency hit of about 2 seconds, because the update call will have to go through consensus (which makes it safe in the first place). Maybe include a bit of documentation to explain this.

Feel free to ping me if you have any questions about it. 🙂 

Before releasing this feature with `dfx`, please do some testing on mainnet to make sure that it works with the service worker.

CC: @nathanosdev, @raymondk

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -39,13 +40,205 @@ pub struct AssetEncoding {
     pub sha256: [u8; 32],
 }
 
+#[derive(Deserialize, CandidType, Serialize, Debug, Default, Clone, PartialEq, Eq)]
+pub struct RedirectUrl {
+    pub(crate) host: Option<String>,
+    pub(crate) path: Option<String>,
```
ericswanson-dfinity, line 46: Regex objects are generally not cheap to construct, but once built, they can be used:
```suggestion
    pub(crate) host: Option<Regex>,
    pub(crate) path: Option<Regex>,
```

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -153,6 +240,213 @@ impl HttpResponse {
...
+}
+
+impl HttpRedirect {
+    pub fn redirect(
+        redirect_config: &Option<Self>,
+        allow_raw_access: Option<bool>,
+        req: &HttpRequest,
+        upgraded: bool,
+    ) -> Option<HttpResponse> {
+        let redirect_from_raw = allow_raw_access.map_or(false, |v| v)
```
ericswanson-dfinity, line 276:
```suggestion
        let redirect_from_raw = allow_raw_access.unwrap_or(false)
```

PR #2417, opened 2022-08-05
docs: Revert "docs: fixing links"

aisconnolly:
Reverts dfinity/sdk#2410

The 'fix' here breaks the build in the portal repo. Will revisit how to deal with this.


aisconnolly:
False alarm, apologies for the inconvenience!

PR #2434, opened 2022-08-10
feat: gzip uploaded canister files

adamspofford-dfinity:
Implements SDK-627.


src/dfx/src/commands/canister/install.rs
```diff
@@ -52,6 +52,10 @@ pub struct CanisterInstallOpts {
     /// Specifies a particular WASM file to install, bypassing the dfx.json project settings.
     #[clap(long, conflicts_with("all"))]
     wasm: Option<PathBuf>,
+
+    /// Skips GZIP-compressing the module.
+    #[clap(long)]
+    uncompressed: bool,
```
sesi200, line 58: How realistic is it to auto-detect this from the file name?

adamspofford-dfinity: The code already does detect it from something much more reliable, which is that a WASM module always starts with the bytes `\0asm`. This is in case, for whatever reason, you *specifically* don't want your uncompressed WASM file to be compressed.

adamspofford-dfinity:
Permanently blocked on replica decisionmaking.

PR #2435, opened 2022-08-11
fix: can install wasm to canister id

sesi200:
# Description

`dfx canister install --wasm my.wasm <canister id>` fails because the canister is not mentioned in `dfx.json`. This should not be the case.

Fixes error reported [here](https://dfinity.slack.com/archives/CGA566TPV/p1660215644715569)

# How Has This Been Tested?

Added e2e

# Checklist:


Dfinity-skaestle:
Thanks Severin, this is awesome!

PR #2440, opened 2022-08-12
feat: Print the dashboard URL on startup

adamspofford-dfinity:
Implements SDK-620. 

I am not sure if this is the best place to put it.


sesi200:
Also needs a changelog entry

PR #2445, opened 2022-08-15
chore: Change jq commands to use sponge

adamspofford-dfinity:
Currently, we have a large number of commands that go `cat <<<"$(jq ... foo.json)" >foo.json`. This is hard to read, hard to write, and breaks vscode's syntax highlighter depending on the quoting. This PR changes it to `jq ... foo.json | sponge foo.json`, utilizing moreutils' sponge, which also lets us get rid of every single SC2094 suppression.


ericswanson-dfinity (approved):
Nice!

PR #2464, opened 2022-08-19
feat: simple cycles faucet code redemption

sesi200:
# Description

Using `dfx wallet --network ic redeem-faucet-coupon <my coupon>` faucet users have a much easier time to redeem their codes.
If the active identity has no wallet configured, the faucet supplies a wallet to the user that this command will automatically configure.
If the active identity has a wallet configured already, the faucet will top up the existing wallet.

Moved the functions `format_as_trillions` and `pretty_thousand_separators` to shared utils from `command/balance.rs`

Fixes [SDK-549](https://dfinity.atlassian.net/browse/SDK-549)

# How Has This Been Tested?

Added an e2e test that demonstrates the expected behaviour.

# Checklist:


e2e/tests-dfx/wallet.bash
```diff
@@ -176,3 +176,45 @@ teardown() {
     assert_command_fail dfx wallet upgrade
     assert_match "There is no wallet defined for identity 'default' on network 'local'.  Nothing to do."
 }
+
+@test "redeem-faucet-coupon can set a new wallet and top up an existing one" {
+    dfx_new hello
+    dfx_start
+    setup_actuallylocal_shared_network
```
ericswanson-dfinity, line 183: The "actuallylocal" network is used for tests that are testing some functionality that uses a network that is configured as a "network provider" (see ConfigNetworkProvider, meaning it only defines a url), but that "actually" points to the local network.  So if it isn't required (and my guess is that it isn't here), it's better to use the normal local network definition.
```suggestion
```

src/dfx/src/lib/identity/mod.rs
```diff
@@ -558,7 +559,7 @@ impl Identity {
         match Identity::wallet_canister_id(network, name)? {
             None => {
                 // If the network is not the IC, we ignore the error and create a new wallet for the identity.
-                if !network.is_ic || create {
+                if (!network.is_ic && std::env::var("DISABLE_AUTO_WALLET").is_err()) || create {
```
ericswanson-dfinity, line 562: I'm having a little trouble following this logic (to be fair, it was hard to follow before, too).

If this environment variable is controlling something only on mainnet, should we reflect that in the name?

I've submitted https://github.com/dfinity/sdk/pull/2466 , which removes the `create` parameter altogether, and may help here.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/identity/mod.rs
sesi200: > I've submitted https://github.com/dfinity/sdk/pull/2466 , which removes the create parameter altogether, and may help here.
Good change, but does not affect this.

This does only change the behaviour on non-mainnet (the `!is_ic` flag). The 'problem' is that on local (and test) networks dfx automatically creates a wallet whenever `get_or_create_wallet` is called. This is very convenient for `dfx start --background && dfx deploy` since the wallet is automatically created, but it makes it impossible to test the case where no wallet exists.

The introduced flag means that no wallet gets automatically created on non-mainnets. Otherwise it would not be possible to test the case where the user has no wallet configured yet.

ericswanson-dfinity, line 562: Let's prefix this with `DFX_`:, `DFX_DISABLE_AUTO_WALLET`
```suggestion
                if (!network.is_ic && std::env::var("DFX_DISABLE_AUTO_WALLET").is_err()) || create {
```

src/dfx/src/lib/identity/mod.rs
```diff
@@ -548,6 +548,7 @@ impl Identity {
     /// Gets the currently configured wallet canister. If none exists yet and `create` is true, then this creates a new wallet. WARNING: Creating a new wallet costs ICP!
     ///
     /// While developing locally, this always creates a new wallet, even if `create` is false.
+    /// This can be prohibited by setting the DISABLE_AUTO_WALLET env var.
```
ericswanson-dfinity, line 551:
```suggestion
    /// This can be inhibited by setting the DISABLE_AUTO_WALLET env var.
```

src/dfx/src/lib/identity/mod.rs
```diff
@@ -612,6 +613,7 @@ impl Identity {
...
```
ericswanson-dfinity, line 616:
```suggestion
    /// This can be inhibited by setting the DFX_DISABLE_AUTO_WALLET env var.
```

ericswanson-dfinity (approved):

PR #2473, opened 2022-08-23
feat: Support installing canisters not in dfx.json

bitdivine:
# Description
Installing NNS canisters involves operating on canisters not listed in dfx.json.

In two places, the existing code checks for canister data in dfx.json.  It doesn't matter if the data is missing, but if the canister is not listed at all, the cli bails out.  E.g. it checks whether the canister has a remote canister ID; it doesn't need to have one, but if the canister is not listed, we crash.  Instead, we would like to treat this in the same way as the data being absent.

We wish to support the case where one dfx server is shared across multiple projects and explicitly do not want to add entries to the dfx.json of a particular project unless the user chooses to do so.

# How Has This Been Tested?
In the development branch `dfx nns install` this change allows NNS canisters to be installed without listing them in dfx.json.

# Checklist:

### DIFF
--- a/src/dfx/src/lib/operations/canister/create_canister.rs
+++ b/src/dfx/src/lib/operations/canister/create_canister.rs
@@ -42,7 +42,8 @@ pub async fn create_canister(
 
     if let Some(remote_canister_id) = config
         .get_config()
-        .get_remote_canister_id(canister_name, &network_name)?
+        .get_remote_canister_id(canister_name, &network_name)
+        .unwrap_or_default()
     {
         bail!(
             "{} canister is remote on network {} and has canister id: {}",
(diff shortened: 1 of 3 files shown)


src/dfx/src/lib/operations/canister/create_canister.rs
```diff
@@ -42,7 +42,8 @@ pub async fn create_canister(
 
     if let Some(remote_canister_id) = config
         .get_config()
-        .get_remote_canister_id(canister_name, &network_name)?
+        .get_remote_canister_id(canister_name, &network_name)
+        .unwrap_or_default()
```
bitdivine, line 46: If creating a canister not in dfx.json, this would bail.  Now, when checking whether there is already a defined remote canister ID, we don't care if there is no canister entry whatsoever in dfx.json.

sesi200 (approved):

Issue #2357, opened 2022-07-18
dfx deploy with gzip


lastmjs:
Version `0.10.x` of `dfx` shipped with support for gzip Wasm binaries locally, but it doesn't work with `dfx deploy`. I would like to be able to simply change the path to the Wasm binary in my `dfx.json` and use the same process as without a gzipped binary

I would like to the following `dfx.json` file to be valid:

```json
{
    "canisters": {
        "bytes": {
            "type": "custom",
            "build": "npx azle bytes",
            "root": "src",
            "ts": "src/bytes.ts",
            "candid": "src/bytes.did",
            "wasm": "target/wasm32-unknown-unknown/release/bytes.wasm.gz",
            "declarations": {
                "output": "test/dfx_generated/bytes"
            }
        }
    }
}
```

sesi200:
Agreed, there should be an easy way to do this.

smallstepman:
fixed in https://github.com/dfinity/sdk/pull/2482
to be released in dfx 0.12.0

gabrielnic:
I posted this question on the forum but maybe I can get an answer here as well https://forum.dfinity.org/t/dfx-deploy-with-gzip/16671 thanks

PR #2482, opened 2022-08-23
fix: Only add custom section for wasm-format files

### LINKED ISSUE #2459: `dfx` attempts to add candid service metadata to Gzip compressed Wasm, and fails
I compressed my canister Wasm using `gzip --to-stdout --best myproject.wasm > myproject.wasm.gz` and encountered this error:

`Failed to add candid service metadata from .dfx/local/canisters/myproject/myproject.did to .dfx/local/canisters/myproject/myproject.wasm.: Failed to parse file .dfx/local/canisters/myproject/myproject.wasm: Bad magic number (at offset 0)`

The `"wasm"` field of my canister in `dfx.json` points to `myproject.wasm.gz` but if I change it back to `myproject.wasm` I don't get the above error.

I'm building my Rust canister using a custom `build` step. Maybe that has something to do with it: https://github.com/dfinity/sdk/blob/97b3de9285c1dcf0e565453b80379c957fe403e4/CHANGELOG.md#fix-dfx-now-only-adds-candidservice-metadata-to-custom-canisters-that-have-at-least-one-build-step

I'm using dfx 0.11.1

adamspofford-dfinity:
Previously we attempted to add a section to all custom canisters with a build task, via walrus. This modifies the check to also make sure the file is actually a wasm module, i.e. starts with `b"\0asm"`. That way, build tasks can instead generate a GZIP archive, or whatever else the replica might support in the future. Fixes #2459 and #2357, fixes SDK-696, implements SDK-627.


e2e/assets/gzip/dfx.json
```diff
@@ -0,0 +1,16 @@
...
+            "candid": "main.did",
+            "wasm": "main.wasm.gz",
+            "build": "sh -c 'gzip -c main.wasm >main.wasm.gz'"
+        }
+    },
+    "networks": {
+        "local": {
+            "bind": "127.0.0.1:8000"
+        }
+    }
```
ericswanson-dfinity, line 15: Tests should use the shared local network, unless they are specifically testing project-specific local networks
```suggestion
    }
```

ericswanson-dfinity (changes requested):

e2e/assets/gzip/dfx.json
adamspofford-dfinity: Oh, I wasn't sure what that meant.

ericswanson-dfinity (approved):

PR #2489, opened 2022-08-24
feat(frontend-canister): use JSON5 file format for `.ic-assets.json5` config files 

smallstepman:
# Description

Allow to use [JSON5](https://json5.org/) file format for `.ic-assets.json5` (renamed from `.ic-assets.json`) files.
### Example usage
`.ic-assets.json5`
```json5
// comment
[
// comment
  {
    "match": "*",
    /*
    look at these beatiful keys below not wrapped in quotes
*/  cache: { max_age: 999 } 
  }
]
```

Closes [SDK-680](https://dfinity.atlassian.net/browse/SDK-680)

# How Has This Been Tested?

Unit tested

# Checklist:


sesi200 (approved):

smallstepman:
@sesi200 @ericswanson-dfinity is it reasonable to straight out abandon `.ic-assets.json` and replace it with `.ic-assets.json5`, or instead we should support both?

sesi200:
I think we should support both. 2 months ago I wasn't even aware json5 was a thing, so I would have assumed that something called `.json5` is a typo and would have 'corrected' it without a second thought.

smallstepman:
Nice, I haven't thought about it from that perspective and I agree.

PR #2493, opened 2022-08-24
test: replace cargo build test with required unit (cargo test) workflow

ericswanson-dfinity:
We already have several other workflows that check `cargo build`

Aggregated matrix results for a required status check


sesi200 (approved):

smallstepman:
what do you think about `cargo test --workspace --all-features --no-fail-fast`?

PR #2498, opened 2022-08-25
chore: release 0.11.2

smallstepman:


smallstepman:
manifest.json has incorrect value for `latest`, but I don't think it matters anyway, since another PR (promote) will follow

ericswanson-dfinity (approved):

ericswanson-dfinity:
> manifest.json has incorrect value for `latest`, but I don't think it matters anyway, since another PR (promote) will follow

That's good actually, though it doesn't matter as long as it's on a branch.  We update the `latest` value only when we want that to be the default version to install, that is, when we promote.

PR #2499, opened 2022-08-26
feat: secp256k1 keys

adamspofford-dfinity:
Implements SDK-595. In line with quill's preferred format, this changes the default identity creation command to use secp256k1 keypairs instead of ed25519 keypairs. Other than general bitcoin friendliness, this enables using bip39 seed phrases which allows for paper backups of the identity pem.


src/dfx/src/lib/identity/pem_encryption.rs
```diff
@@ -21,7 +20,6 @@ pub fn load_pem_file(path: &Path, config: Option<&IdentityConfiguration>) -> Dfx
     let content = std::fs::read(path)
         .with_context(|| format!("Failed to read {}.", path.to_string_lossy()))?;
     let content = maybe_decrypt_pem(content.as_slice(), config)?;
-    identity_utils::validate_pem_file(&content)?;
```
ericswanson-dfinity, line 24: What's the rationale for validating the pem file at only one call point, rather than here?

ericswanson-dfinity (changes requested):

src/dfx/src/lib/identity/pem_encryption.rs
adamspofford-dfinity: Because in two out of the three code paths that call this, right after loading it, it is parsed into an identity, which is also the validation step. So the same work is being performed twice.

ericswanson-dfinity: > in two out of the three code paths
Okay.  Then the third case is export.  Is it intentional then that in that case, we no longer validate the key before exporting it?

adamspofford-dfinity: Ah, I seem to have misremembered, I was thinking of *import*.

ericswanson-dfinity (approved):

e2e/tests-dfx/identity_command.bash
```diff
@@ -452,3 +452,14 @@ XXX
...
+
+@test "identity: can import a seed phrase" {
+    reg="seed phrase for identity 'alice': ([a-z ]+)"
+    assert_command dfx identity new --disable-encryption alice
+    [[ $stderr =~ $reg ]]
+    echo "${BASH_REMATCH[1]}" >seed.txt
+    principal=$(dfx identity get-principal --identity alice)
+    assert_command dfx identity import alice2 --seed-file seed.txt --disable-encryption
+    assert_command dfx identity get-principal --identity alice2
+    assert_eq "$principal"
```
ericswanson-dfinity, line 464: Why not make this test use a concrete seed phrase, and then verify against the known fixed principal?

adamspofford-dfinity: Making sure random inputs consistently work. But I also see the use of testing known input, so let's do both.

PR #2500, opened 2022-08-26
chore: remove unused dependency pathdiff

adamspofford-dfinity:
The pathdiff dependency from ic-asset is unused. This removes it.


smallstepman (approved):
thanks!

PR #2359, opened 2022-07-19
feat(frontend-canister)!: default secure configuration for assets in frontend project template

smallstepman:
# Description

blocked by:
- [x] https://github.com/dfinity/sdk/pull/2489
- [x] #2824 
  
Closes https://dfinity.atlassian.net/browse/SDK-473
Closes https://dfinity.atlassian.net/browse/PSEC-710

# How Has This Been Tested?

e2e

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [x] I have edited the CHANGELOG accordingly.
- [ ] I have made corresponding changes to the documentation.

PR #2509, opened 2022-08-29
fix(frontend-canister): copy `.ic-assets.json` file from `src` to `dist`

smallstepman:
# Description

for default node starter template: adds webpack copy plugin to copy `.ic-assets.json*` file from `src` to `dist` 

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 ## DFX
 
+### fix: For default node starter template: copy `ic-assets.json5` file from `src` to `dist`
+
 ### refactor: Move replica URL functions into a module for reuse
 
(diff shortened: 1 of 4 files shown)


ericswanson-dfinity (changes requested):
Can we have an e2e test that demonstrates that this works?

The test should use `dfx_new_frontend`, then add a `.ic-assets.json` alongside index.html or index.js, then deploy, then use curl -v or something to make sure that headers get through.

smallstepman:
what do you think about instead writing e2e here: https://github.com/dfinity/sdk/pull/2359 ?

there, we will already have a default `.ic-assets.json` file to test with

ericswanson-dfinity:
Since this PR changes the behavior, I think it'll be better to include a test in this PR with it.  The test can be much simpler than the test in that other PR.

ericswanson-dfinity (approved):

PR #2510, opened 2022-08-29
feat: Backwards compatibility for flags

bitdivine:
# Description
The location of some flags has moved.  For example:
```
Old: dfx canister --network local id mydapp
New: dfx canister id --network local mydapp
```
This is quite disruptive, as it breaks almost every script that uses dfx.

This PR implements backwards-compatibility by accepting the old format in a few common cases.  It does not attempt to cover all possible situations where we know how to rewrite from the old syntax to the new.  It removes 90% of the pain with 10% of the code.

## Alternatives considered
Update all the code.  This would however break any case where say `sns` was used with an older version of dfx.   Also, updating to this latest dfx would break our CI in many places and I doubt we could fix it all at once.

# How Has This Been Tested?
I have used this for an SNS deployment, as that uses quite a few scripts in our inventory, as well as the sns binary which has calls to dfx baked into the code.

If this general approach seems OK I am happy to write bash tests as well.

# Checklist:


ericswanson-dfinity:
Closing because https://github.com/dfinity/sdk/pull/2521 has made the changes in a backwards-compatible way.

PR #2513, opened 2022-08-29
test: run cargo clean before cargo build in e2e workflow

ericswanson-dfinity:
In the e2e workflow, call `cargo clean` before any `cargo build` in order to avoid any problems with stale inputs.

In particular scripts/prepare-dfx-assets.sh used to call `cargo` to build icx-proxy, so it was advantageous to not have to do that every time.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -30,9 +30,9 @@ jobs:
           rustup default ${{ matrix.rust }}
           rustup component add rustfmt
       - name: Build
-        # note: since there is no 'cargo clean' here, the results of scripts/prepare-dfx-assets.sh
-        # are often cached.
-        run: cargo build --locked --release
+        run: |
+          cargo clean
+          cargo build --locked --release
         env:
           OPENSSL_STATIC: yes
       - name: Strip binaries
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,9 +29,7 @@ jobs:
           rustup default ${{ matrix.rust }}
           rustup component add clippy
       - name: Run Lint
-        run: |
-          cargo clean
-          cargo clippy --verbose --tests --benches -- -D clippy::all
+        run: cargo clippy --verbose --tests --benches -- -D clippy::all
         env:
           RUST_BACKTRACE: 1
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,7 +29,9 @@ jobs:
           rustup default ${{ matrix.rust }}
           rustup component add clippy
       - name: Run Lint
-        run: cargo clippy --verbose --tests --benches -- -D clippy::all
+        run: |
+          cargo clean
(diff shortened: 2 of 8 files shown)


.github/workflows/e2e.yml
```diff
@@ -30,9 +30,9 @@ jobs:
           rustup default ${{ matrix.rust }}
           rustup component add rustfmt
       - name: Build
-        # note: since there is no 'cargo clean' here, the results of scripts/prepare-dfx-assets.sh
-        # are often cached.
-        run: cargo build --locked --release
+        run: |
+          cargo clean
```
smallstepman, line 34: since we're running `cargo clean`, I think we can skip caching `target` directory
https://github.com/dfinity/sdk/pull/2513/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR25

.github/workflows/unit.yml
```diff
@@ -29,7 +29,9 @@ jobs:
           rustup default ${{ matrix.rust }}
 
       - name: Check cargo test
-        run: cargo test --workspace --all-features --no-fail-fast
+        run: |
+          cargo clean
```
smallstepman, line 33: same here https://github.com/dfinity/sdk/pull/2513/files#diff-e17ac7ab478a18e9dac875e191cc3e0d754fab62eddddfb05d4619495be927d9L23

.github/workflows/lint.yml
```diff
@@ -29,7 +29,9 @@ jobs:
           rustup default ${{ matrix.rust }}
           rustup component add clippy
       - name: Run Lint
-        run: cargo clippy --verbose --tests --benches -- -D clippy::all
+        run: |
+          cargo clean
```
smallstepman, line 33: same here https://github.com/dfinity/sdk/pull/2513/files#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2R23

.github/workflows/unit.yml
ericswanson-dfinity: What I wonder about this is: does that mean that `cargo clean` here is a mistake?

For the e2e case, the prepare-dfx-assets.sh step is complex enough that it would be pretty easy to get a stale result.  but here?

smallstepman: I don't know if it's a mistake, but I'm unsure about what's the benefit of running `cargo clean` here. Would love to hear @adamspofford-dfinity thoughts

ericswanson-dfinity: I reverted this change

.github/workflows/lint.yml
ericswanson-dfinity: Reverted

smallstepman (approved):

ericswanson-dfinity:
Closing because this adds 20-25 minutes to the build time, and also we have not identified any problems due to not running cargo clean here.

PR #2521, opened 2022-08-30
feat: Accept flags in more positions

adamspofford-dfinity:
#2369 moved flags to a more traditional location, but in doing so broke scripts, as well as making the code worse. This enables both behaviors with code closer to the original.


ericswanson-dfinity (approved):
🔥

ericswanson-dfinity (changes requested):
Please update the changelog section added previously

PR #2556, opened 2022-09-09
fix: use cargo cache for e2e tests

ericswanson-dfinity:
# Description

The e2e test setup scripts change the HOME directory to point to a temporary directory, so that tests don't alter any user files, and vice-versa.
    
This has the effect of making cargo re-download everything into the cache for every test.  Our tests don't need this level of isolation from the cargo cache, and neither do our tests need to incur this cost.
    
Therefore, this change sets `CARGO_HOME=$HOME/.cargo` (the usual default), to take advantage of the cargo cache.

# How Has This Been Tested?

Manually, though existing e2e tests will cover it.

Before (local):
```
bats -T e2e/tests-dfx/rust.bash
rust.bash
 ✓ rust starter project can build and call in 281000ms [281000]
 ✓ rust canister can resolve dependencies in 609000ms [609000]
 ✓ rust canister can have nonstandard target dir location in 295000ms [295000]
 ✓ rust canister fails to build with missing lockfile in 276000ms [276000]

4 tests, 0 failures in 1461 seconds
```

After (local):
```
bats -T e2e/tests-dfx/rust.bash
rust.bash
 ✓ rust starter project can build and call in 83000ms [83000]
 ✓ rust canister can resolve dependencies in 89000ms [89000]
 ✓ rust canister can have nonstandard target dir location in 88000ms [88000]
 ✓ rust canister fails to build with missing lockfile in 87000ms [87000]

4 tests, 0 failures in 347 seconds
```

CI (note that times are not always lower)
| Test | [Before](https://github.com/dfinity/sdk/runs/8277317623?check_suite_focus=true) | [After](https://github.com/dfinity/sdk/actions/runs/3026267345) |
| --- | --- | --- |
| rust / ic-ref / ubuntu | 11:25 | 10:13 |
| rust / replica / ubuntu | 13:14 | 15:23 |
| rust / ic-ref / macos | 18:03 | 16:08 |
| rust / replica / macos | 23:11 | 20:39 |

# Checklist:

- [ ] (not needed) I have edited the CHANGELOG accordingly.
- [ ] (not needed) I have made corresponding changes to the documentation.


e2e/utils/_.bash
```diff
@@ -34,6 +34,7 @@ standard_setup() {
 
     cd "$x/working-dir" || exit
 
+    export CARGO_HOME="$HOME/.cargo"
```
bitdivine, line 37: Two environment variables that might be interesting in this context:
* `CARGO_NET_OFFLINE=true` will try to work offline, using cached data, and only access the network if it really needs to.
* `CARGO_NET_RETRY=10` will try harder if the network is inaccessible.  Sometimes CI fails for me due to a "spurious network error" and hitting retry passes, but this wastes a lot of time compared with just trying a bit harder to download rust packages.  Of course, if the local cache is used more, CI will download less so this will be an issue less frequently.

See: https://doc.rust-lang.org/cargo/reference/config.html#netretry

PR #2557, opened 2022-09-09
fix: make frontend canister build reproducible and checked by github actions

mraszyk:


.github/workflows/build-frontend-canister.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  frontend:
+    runs-on: ubuntu-latest
```
ericswanson-dfinity, line 11: Suggestion: specify a name here, that is similar to the other required status names, to make it easy to specify
```suggestion
    runs-on: ubuntu-latest
    name: frontend-canister-up-to-date:required
```

.github/workflows/build-frontend-canister.yml
```diff
@@ -0,0 +1,32 @@
...
+          rustup target add wasm32-unknown-unknown
+          cargo install --version 0.1.5 ic-wasm
+      - name: Build frontend canister
+        run: |
+          ./scripts/update-frontend-canister.sh
+      - name: Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: assetstorage
+          path: ${{ github.workspace }}/src/distributed/assetstorage.wasm.gz
```
ericswanson-dfinity, line 26: What's the purpose of uploading this an an artifact here?  It's not used in a later step

scripts/update-frontend-canister.sh
```diff
@@ -8,7 +8,20 @@ die() {
...
 
-export RUSTFLAGS="--remap-path-prefix=\"${PWD}\"=./ --remap-path-prefix=\"${HOME}\"=_/"
+SCRIPT=$(readlink -f "${0}")
+SCRIPT_DIR=$(dirname "${SCRIPT}")
+cd ${SCRIPT_DIR}/..
+
+if [ -z "${CARGO_HOME}" ]
+then
+  export CARGO_HOME="${HOME}/.cargo"
+fi
```
ericswanson-dfinity, line 18: This is already the default.  Am I missing something, or does this have no effect?

ericswanson-dfinity (changes requested):

scripts/update-frontend-canister.sh
mraszyk: It sets the environment variable to the default value if the environment variable is undefined (`-z "${CARGO_HOME}"`).

.github/workflows/build-frontend-canister.yml
mraszyk: The purpose is to allow the developer to check in the artifact in case the test fails (alternatively, the developer would need to run `./scripts/update-frontend-canister.sh`).

scripts/update-frontend-canister.sh
ericswanson-dfinity: Right, but according to https://doc.rust-lang.org/cargo/reference/environment-variables.html, that's the default.  Setting it here makes the reader wonder why it's set.  I'd recommend not to set it, and let cargo set its own defaults.
```suggestion
```

mraszyk: It's not set for the sake of cargo, but for the sake of remapping path prefixes in `RUSTFLAGS`.

ericswanson-dfinity (approved):

PR #2559, opened 2022-09-10
fix: Use NNS did files that matches the wasms

bitdivine:
# Description
At present NNS did files are downloaded from the head of the ic repo main branch.  However the canister wasms are from a fixed commit, which may not be compatible with head of master.

So now the did files are pulled from the same fixed commit as the canister wasms.

# How Has This Been Tested?
I had a very confusing error which I eventually tracked down to these mismatched did files.  With this change, the error is gone.

# Checklist:


chenyan-dfinity:
How do we build NNS canister? If it's build by dfx, the did file is available as a custom section in the wasm module, and can be fetched from the state tree on-chain.

If we use a custom build script, we probably need to add that custom section via `ic-wasm` to be consistent with the rest of the ecosystem.

ericswanson-dfinity (approved):

ericswanson-dfinity (approved):

bitdivine:
@chenyan-dfinity 
> If it's build by dfx, the did file is available as a custom section in the wasm module, and can be fetched from the state tree on-chain.

That is very interesting.  I didn't know that.  Do you know the command for getting the did file for an arbitrary canister?   This could be very useful.

Regarding this PR; in the interest of speed I'd like to merge it as is but it would be interesting to see what could be done with this idea.  I suppose that there is no guarantee that the local dfx server is running when `dfx nns import` is run.  We could make that a requirement.  We could also pull this from prod, but that conflicts with the idea of dfx being locked to a specific version of the IC.  I'd like to explore the options.  Can the wasm also be downloaded?  I thought the answer to that was no but I could be wrong.

chenyan-dfinity:
> Do you know the command for getting the did file for an arbitrary canister?

`dfx canister metadata <canister_name/id> candid:service`.

To patch the did file into Wasm module, see https://github.com/dfinity/ic-wasm#metadata.

Right, you cannot download wasm module, but you can download any metadata, which is essentially a special named custom section.

PR #2573, opened 2022-09-15
chore: If pushing twice to the same branch, later jobs kill earlier ones

bitdivine:
# Description
CI sometimes, perhaps even often, has to wait minutes to tens of minutes for runners to be available.  Reducing the total number of CPU minutes we spend is a way of freeing up compute capacity.

If a developer (e.g. myself) pushes a branch, then notices a change that needs to be made and pushes again, the earlier job still runs to completion even though it is no longer of interest.  It would be nice to kill the earlier job when it is superseded.

Github's concurrency provides setting to do just this:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

I have opted for the configuration where each workflow has its own groups, so that a kill command from one workflow doesn't kill others.  This lets us choose for which workflows this policy is applied.  I have applied the concurrency limit to every workflow but we may decide that it is only appropriate for some.

# How Has This Been Tested?
* Feel free to push empty commits to this PR and watch CI: `git commit -m bump --allow-empty`
  * I personally have observed this working.  I pushed, waited until there were about 10 entries in https://github.com/dfinity/sdk/actions , then pushed again.  After about 20 seconds the earlier jobs that had not completed were greyed out.

# Checklist:

   - I think there is no corresponding documentation so this is a no-op.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## DFX
 
+### chore: Kill unwanted CI jobs
+
+If pushing twice to the same branch in quick succession, the earlier CI jobs are cancelled in favour of the later ones.
+
 ### feat: add --mode=auto
 
 When using `dfx canister install`, you can now pass `auto` for the `--mode` flag, which will auto-select `install` or `upgrade` depending on need, the same way `dfx deploy` does. The default remains `install` to prevent mistakes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,8 @@
 
 ## DFX
 
+
 ### fix: `cargo run -p dfx -- --version` prints correct version
-
-### chore: Kill unwanted CI jobs
-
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -4,9 +4,12 @@
 
 ## DFX
 
-
 ### fix: `cargo run -p dfx -- --version` prints correct version
 
+### chore: Kill unwanted CI jobs
+
+If pushing twice to the same branch in quick succession, the earlier CI jobs are cancelled in favour of the later ones.
+
```
ericswanson-dfinity, line 12: We track changes to dfx in the changelog, but not things like changes to our CI process.
```suggestion
```

ericswanson-dfinity (changes requested):

CHANGELOG.md
bitdivine: Ack.  Makes sense.  Undone.

ericswanson-dfinity (approved):

PR #2574, opened 2022-09-16
fix: Valid canister-based env vars

bitdivine:
# Description
Hyphens are not valid in shell environment variables, but do occur in canister names such as nns-dapp.  This poses a problem for vars with names such as `CANISTER_ID_${CANISTER_NAME}`.

Changes:
- In env vars set by dfx, replace hyphens by underscores.

# How Has This Been Tested?

# Checklist:


sesi200 (approved):

lwshang:
These env vars are conceptually a contract between dfx and underlying canister building processes. Rust CDK is relying on them. So this PR introduces a breaking change that dfx won't work with Rust CDK. If we do want to merge this PR. We need make corresponding change on the CDK side.

bitdivine:
> These env vars are conceptually a contract between dfx and underlying canister building processes. Rust CDK is relying on them. So this PR introduces a breaking change that dfx won't work with Rust CDK. If we do want to merge this PR. We need make corresponding change on the CDK side.

Good to know.  thank you @lwshang   Who works on the CDK and could help with this?  Note that, as "-" is invalid in an env var, this PR doesn't change any env vars but it does create new ones.

ericswanson-dfinity (approved):
I don't know why we didn't also make the environment variable names all uppercase, but they certainly shouldn't include dashes.

PR #2578, opened 2022-09-16
feat: ic-wasm shrink by default

lwshang:
# Description

Default to run ic-wasm shrink when build canisters.
This behavior applies to Motoko, Rust and Custom canisters.
`dfx build --no-shrink` to opt out.

# How Has This Been Tested?

CI

# Checklist:


src/dfx/src/commands/build.rs
```diff
@@ -24,6 +24,10 @@ pub struct CanisterBuildOpts {
     #[clap(long)]
     check: bool,
 
+    /// Disable default WASM shrink after build.
+    #[clap(long)]
+    no_shrink: bool,
+
```
ericswanson-dfinity, line 30: If we make this a per-canister option in dfx.json option instead of a `dfx build` option, then the developer won't have to remember to specify it every time, and it will work with `dfx deploy`.

src/dfx/src/lib/builders/custom.rs
```diff
@@ -150,6 +150,14 @@ impl CanisterBuilder for CustomBuilder {
             add_candid_service_metadata = false;
         }
 
+        // Custom canister may have WASM gzipped
+        if config.shrink_after_build
+            && wasm.extension().and_then(std::ffi::OsStr::to_str) == Some("wasm")
+        {
+            info!(self.logger, "Shrink WASM module size.");
+            super::shrink_wasm(&wasm)?;
+        }
```
ericswanson-dfinity, line 159: Instead of looking at the file extension, we can inspect the file header.  This information is already available in this function, immediately above, lines 145-150.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/builders/mod.rs
```diff
@@ -376,6 +376,23 @@ impl BuildConfig {
     }
 }
 
+fn shrink_wasm(wasm_path: impl AsRef<Path>) -> DfxResult {
+    let wasm_path = wasm_path.as_ref();
+    let wasm = std::fs::read(wasm_path).expect("Could not read the WASM module.");
```
ericswanson-dfinity, line 381: Is there a reason not to use context here?
```suggestion
    let wasm = std::fs::read(wasm_path).context("Could not read the WASM module.")?;
```

src/dfx/src/lib/builders/mod.rs
```diff
@@ -376,6 +376,23 @@ impl BuildConfig {
...
+    let mut module_config = walrus::ModuleConfig::new();
+    module_config.generate_name_section(true);
+    module_config.generate_producers_section(false);
+    let mut module = module_config
+        .parse(&wasm)
+        .expect("Could not parse the WASM module.");
+    ic_wasm::shrink::shrink(&mut module);
+    module
+        .emit_wasm_file(wasm_path)
+        .unwrap_or_else(|_| panic!("Could not write shrinked WASM to {:?}", wasm_path));
```
ericswanson-dfinity, line 391: Same here, unless there's something I'm missing, we should propagate the error rather than panic
```suggestion
        .with_context(|| format!("Could not write shrinked WASM to {:?}", wasm_path))?;
```

src/dfx/src/lib/builders/mod.rs
```diff
@@ -376,6 +376,23 @@ impl BuildConfig {
...
 
+fn shrink_wasm(wasm_path: impl AsRef<Path>) -> DfxResult {
+    let wasm_path = wasm_path.as_ref();
+    let wasm = std::fs::read(wasm_path).expect("Could not read the WASM module.");
+    let mut module_config = walrus::ModuleConfig::new();
+    module_config.generate_name_section(true);
+    module_config.generate_producers_section(false);
+    let mut module = module_config
+        .parse(&wasm)
+        .expect("Could not parse the WASM module.");
```
ericswanson-dfinity, line 387:
```suggestion
        .context("Could not parse the WASM module.")?;
```

src/dfx/src/lib/builders/mod.rs
```diff
@@ -376,6 +376,23 @@ impl BuildConfig {
     }
 }
 
+fn shrink_wasm(wasm_path: impl AsRef<Path>) -> DfxResult {
```
ericswanson-dfinity, line 379: Something like
```suggestion
#[context("Failed to shrink wasm at {}.", &wasm_path.display())]
fn shrink_wasm(wasm_path: impl AsRef<Path>) -> DfxResult {
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #2584, opened 2022-09-19
feat: Add subnet type argument when creating canisters

dsarlis:
# Description

This PR implements a new optional argument when creating canisters through dfx that allows users to choose a specific "type" of subnet that their canister can be created on. These user-facing types are different than the existing types of subnets in the registry (system,verified, application) and are mostly useful for allowing users to choose a subnet with some certain characteristics.

Closes SDK-700

# How Has This Been Tested?

This has been tested by deploying a version of the CMC canister on a testnet supporting the new concept of "types" of subnets, giving my dfx principal some ICP, create some types of subnets and assigning them and then creating canisters on a specific type and observing that the canister gets created on a subnet of that type.

There are also a couple e2e tests that were added that cover the new code paths.

# Checklist:


dsarlis:
An open question: Should we also expose a sub-command that allows listing of the available types? We can retrieve them from the cycles minting canister. In fact, I have added a check that the desired type is in the list of available types whenever a request to create a canister is submitted but perhaps it's better to have a separate command to list the types for better UX. Wdyt?

src/dfx/src/commands/ledger/mod.rs
```diff
@@ -238,6 +242,34 @@ pub async fn notify_top_up(
     Ok(result)
 }
 
+pub async fn validate_subnet_type(agent: &Agent, subnet_type: &Option<String>) -> DfxResult<()> {
+    let result = agent
+        .update(
+            &MAINNET_CYCLE_MINTER_CANISTER_ID,
+            GET_SUBNET_TYPES_TO_SUBNETS_METHOD,
+        )
```
sesi200, line 250: This looks like it will blow up in some local development mode (unavoidable call to hard-coded ID that is only valid on mainnet, CMC is not necessarily even installed in local mode even if ledger is present)

sesi200: also, is it even necessary to validate? Doesn't the canister creation call fail if an invalid subnet type is supplied?

dsarlis: That's a good point. I thought it might be nicer to do this validation on the dfx side as well. You're right though that canister creation will fail if an unknown subnet type is supplied (with a relevant error message). In local development mode, I don't think you ever create canisters with `dfx ledger` (you need to setup a ledger and CMC first correctly so I doubt users do that normally). That said, we can skip this validation, if you think it's not the best idea.

In this case, having a way to view all available types is even more useful I think (I was hoping that if you use the wrong one and then we show you the full list is helpful in order to figure out which subnet types you can use).

sesi200: I agree having a way to display available types is a good idea. And I think we can do all of what you suggest when handling errors, which will save us a an update call round-trip in most cases. (If the long-approved error codes were finally implemented this would be an even easier task. I have to start bothering people about that.)

dsarlis: Ok, I can then remove the additional check through the update call and add an extra command to display types. Any recommendations on the level this command should be added at? E.g. `dfx canister ...` or just `dfx ...`?

sesi200: eeugh... hardest question so far 😄 I feel like this should be part of `canister create`, but maybe you could make it just a flag: `dfx canister create --available-subnet-types`. But I'm really not sure what the cleanest solution is. @ericswanson-dfinity do you have any good ideas?

ericswanson-dfinity: Since this is used with `dfx ledger create-canister`, what would you think about `dfx ledger subnet-types` ?  Technically it's call to the cycles minter, but the other `dfx ledger` operations are calls to the ledger and the cycles minter in combination.

sesi200: `subnet-types` or `show-subnet-types`? I really want to add autocomplete at some point, so the length wouldn't be _that_ big of an issue

ericswanson-dfinity: Either way, but a separate command for sure

dsarlis: I have added a sub-command to list the subnet types and removed the automatic check when creating the canister as discussed. Let me know how it looks.

src/dfx/src/commands/ledger/show_subnet_types.rs
```diff
@@ -0,0 +1,38 @@
...
+pub async fn exec(env: &dyn Environment) -> DfxResult {
+    let agent = env
+        .get_agent()
+        .ok_or_else(|| anyhow!("Cannot get HTTP client from environment."))?;
+
+    fetch_root_key_if_needed(env).await?;
+
+    let result = agent
+        .update(
+            &MAINNET_CYCLE_MINTER_CANISTER_ID,
```
sesi200, line 23: Can you add a `--cmc-canister-id` flag in case someone has the CMC installed in a different place? You can copy/paste from `dfx ledger transfer --ledger-canister-id`

dsarlis: Done.

sesi200 (changes requested):
Looking good! Few more minor things:

- `/CHANGELOG.md` needs to be updated
- `/docs/cli-reference/dfx-ledger.md` needs to be updated
- Any chance this can be tested locally? If you need help with writing tests I'm happy to help.

dsarlis:
> Looking good! Few more minor things:
> 
> * `/CHANGELOG.md` needs to be updated
> * `/docs/cli-reference/dfx-ledger.md` needs to be updated
> * Any chance this can be tested locally? If you need help with writing tests I'm happy to help.

Cool, I was expecting someone to say we're at this point. Since I'm not usually contributing to the SDK I will likely need some help here.

* Any guidelines for changelog entries? Or example maybe that I could follow?
* Re local testing: I'm not sure how feasible it is given you need a CMC running. Do you guys have something along those lines that I could look at?

sesi200:
> Any guidelines for changelog entries? Or example maybe that I could follow?
Just follow the examples in the existing changelog. The only real guideline is that you should put your change in the `unreleased` section and use the `chore`/`feat`/... in the heading of your entry

> I'm not sure how feasible it is given you need a CMC running.
Strictly speaking you only need a canister that follows the interface you're testing 😄 `e2e/tests-dfx/ledger.bash` shows one way to do it, helper functions are defined in `e2e/utils/_.bash` and `e2e/utils/assertions.bash`.

Testing `show-subnet-types`:
- Option 1: make a mini mock canister that follows the tested interface
- Option 2: test against mainnet
- Option 3: `dfx nns install`, but I'm not sure how stable that is right now

Testing `--subnet-type`: I have no idea how this works on a local replica. I suppose even an error message can demonstrate that the option was passed correctly...

Also, I'm much more responsive via Slack, so feel free to ask more questions there

sesi200 (approved):

PR #2586, opened 2022-09-20
refactor: Factor out common code for sns cli

bitdivine:
# Description
There is a function that calls the sns cli and that deals with corner cases.  This is used in e.g. `dfx sns config validate`.  In this PR the common code is also used in `dfx sns config create`.

# How Has This Been Tested?
There is no change in functionality.  Existing tests should suffice.

# Checklist:

  This is a noop
  This is a noop


src/dfx/src/lib/sns/create_config.rs
```diff
@@ -1,40 +1,21 @@
...
-    command
-        .stdin(process::Stdio::null())
-        .output()
-        .map_err(anyhow::Error::from)
-        .and_then(|output| {
-            if output.status.success() {
-                Ok(())
-            } else {
-                Err(anyhow!(
-                    "Failed to create an SNS configuration.\n{:?} {:?}\nStdout:\n{:?}\n\nStderr:\n{:?}",
```
bitdivine, line 31: Note: The context message on the function provides the important contextual information, and the library function provides stdout and stderr if it fails.

PR #2587, opened 2022-09-20
fix: Lock SNS version

bitdivine:
# Description
`dfx sns import` was pulling did files from master, so the did files did not match the installed wasms, causing some commands that used to work to fail.

Now, it was actually useful to be able to pull recent did files, so while setting the default reference to download from to the replica commit, I maintained the ability to pull from master by introducing an environment variable `DFX_IC_REF` that can be used to override the commit.  I have added the same env var to `dfx nns install` and `dfx sns import` so that all the wasms are in sync, with the same commit. 

# How Has This Been Tested?
Existing tests for the SNS should suffice.

# Checklist:

  - This is a noop, I think.  It is a fix to an as yet unreleased functionality.
  - This is a noop.


src/dfx/src/commands/nns/import.rs
```diff
@@ -24,7 +24,7 @@ pub async fn exec(env: &dyn Environment, opts: ImportOpts) -> DfxResult {
     let mut config = config.as_ref().clone();
 
     let network_mappings = get_network_mappings(&opts.network_mapping)?;
-    let ic_commit = replica_rev();
+    let ic_commit = std::env::var("DFX_IC_REF").unwrap_or_else(|_| replica_rev().to_string());
```
ericswanson-dfinity, line 27: "ic ref" already has a meaning in dfx, which is the IC reference implementation.   Could we call this `DFX_IC_COMMIT` instead?

bitdivine: Sure thing.  Done.

ericswanson-dfinity (approved):

PR #2588, opened 2022-09-20
chore: update JSON schema docs

ericswanson-dfinity:
# Description

I thought we had some github workflows triggering to add these changes to PRs in flight, but a few recent changes are not reflected in the JSON schema docs.

# How Has This Been Tested?

n/a

# Checklist:

### DIFF
--- a/docs/dfx-json-schema.json
+++ b/docs/dfx-json-schema.json
@@ -192,7 +192,7 @@
             },
             "candid": {
               "title": "Candid File",
-              "description": "Path to this canister's candid interface declaration.",
+              "description": "Path to this canister's candid interface declaration.  A URL to a candid file is also acceptable.",
               "type": "string"
             },
             "type": {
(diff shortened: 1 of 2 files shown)


lwshang:
It seems that the workflow for updating schemas is not correct: https://github.com/dfinity/sdk/blob/4fa74b9f038ab4a5b5a7d45a0cc132b0ffff7006/.github/workflows/update-docs.yml#L41

`git status` should be changed to `git diff --name-only`

ericswanson-dfinity:
> It seems that the workflow for updating schemas is not correct:
> 
> https://github.com/dfinity/sdk/blob/4fa74b9f038ab4a5b5a7d45a0cc132b0ffff7006/.github/workflows/update-docs.yml#L41
> 
> `git status` should be changed to `git diff --name-only`

I would expect something like `"$(git status --porcelain)" != ""`, but this PR is only concerned with updating the out-of-date docs.

lwshang (approved):

PR #2599, opened 2022-09-23
feat: Lazy download

bitdivine:
# Description
Some but not all nns-wasms are downloaded only if they do not already exist.  With this change all NNS wasm downloads occur only if the wasm is not already present.  This should help testing when one has a slow internet connection.

# How Has This Been Tested?
dfx nns install now works offline, if all assets are cached:

* `dfx new toy`
* `cd toy`
* `dfx start --clean`
* `dfx nns install`
* `dfx stop`
* DISCONNECT NETWORK CONNECTION
* `dfx start --clean`
* `dfx nns install`

# Checklist:

  - This is a noop at the moment - there is no nns documentation.  I have started NNS documentation in a separate PR.


src/dfx/src/lib/nns/install_nns.rs
```diff
@@ -399,19 +417,10 @@ pub async fn download_ic_repo_wasm(
     fs::create_dir_all(wasm_dir)
```
bitdivine, line 417: Note: download() and download_gz() are the only methods for downloading used in this file.  Making those two methods "lazy" covers all uses.

sesi200 (approved):

PR #2606, opened 2022-09-26
test: Retry in the case of network errors

bitdivine:
# Description
This is an attempt to reduce test flakiness.  In particular, many tests die due to external network errors.  These failures can probably be addressed by retries.

Note:  This does not attempt to address internal network errors.  E.g. port conflicts (I have seen this only ~once~ twice on a random port) or `dfx start` failing to respond to http requests.

Is it reasonable to use retries?  If a network request fails 0.1% of the time but we make 1000 requests in a test suite we would expect most test runs to fail.  403 failures are relatively common, presumably because we run a lot of tests in parallel, batter some servers and get rate limited.  An alternative would be to focus on 403s by making fewer requests.  But we also get other errors so looking at the 403s is probably still worthwhile but won't cover all our problems.

Errors observed:
- Cargo network failure.
  - This can presumably be addressed with CARGO_NET_RETRIES
- Assets failed to download
  - This is curl.  There are a lot of curl commands in the repo.  Curl  has a retries parameter that can go in .curlrc
- npm failed to download webpack, so later webpack was not found.
  - Curious that the test did not fail immediately.
  - There is an nmp retry parameter that could have been bumped.
  - The request failed due to a 409, so presumably webpack was being updated or changed in some way at the moment of the request.  retries normally apply to 5xx errors so the 409 would probably still have killed the test.
  - Still, there is probably no harm in bumping the npm retry count.
  - TODO: bump fetch-retries to 10.

# How Has This Been Tested?
It is actually hard to verify that e.g. the CURL retries parameter is being used.  Maybe best to rerun CI on this PR several times and see how often it fails.

- First run (no changes active) about 5 tests failed.
- Second run (cargo and curl parameters active) but 2 tests failed, one from npm install and one derived test.
- Third run (just hit rerun) succeeded.
- 4th run, with an .npmrc in the root of the project.  No errors.
  - Question: Is HOME set to the project root in github actions?  Can we simply put .rc files in the project root?
  - A: No, at least there are different variables for home and the working directory: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#file-systems
  - So, actually the .npmrc file has done nothing.  TODO: Wait for the next npm install error to figure out where these errors actually occur, then in that workflow put the .npmrc in HOME.

# Checklist:

  - This is a CI change, it doesn't affect the end user, so no CHANGELOG entry is provided.
  - This is similarly a noop.


bitdivine:
This is passing consistently so I am marking it as ready.  Unfortunately there have been no npm network errors so the npm settings aren't invoked anywhere yet.

sesi200 (approved):

PR #2616, opened 2022-10-03
fix(generate): add missing typescript types and fix issues with bindings array in dfx.json

### LINKED ISSUE #2615: Breaking changes to language binding assets associated with dfx generate in 0.12.0-beta.3 result in idlFactory type error.
I currently have some software tooling that generates canister declaration files and then copies them over to another directory.

One of the developers using this tooling on the `0.12.0-beta.3` version just reported the following error when trying to start up their frontend after copying over the declarations from the backend canister directory. (I have a script that runs dfx generate before copying over the declarations from the backend to the frontend directory).

It results in the following error.

```
TS2305: Module '"../declarations/index/index"' has no exported member 'idlFactory'.
```

For comparison, no other developers (running <= 0.11.x) have this same issue. 

I dug a bit deeper into this and found that it's because in `0.12.x` the declarations generate an additional index.d.ts file that is not present in dfx versions `<= 0.11.x does not`. (See https://github.com/dfinity/sdk/blob/master/src/dfx/assets/language_bindings/index.d.ts.hbs)

Because of this addition, now TypeScript expects for the idlFactory type to be defined in this `index.d.ts` file and will complain if it is not present there (with the error shown above), instead of just pulling it from `./<canister_name>.did.js.

nathanosdev:
fixes #2615

# Description

This fixes two issues, the first was reported in https://github.com/dfinity/sdk/issues/2615.
The 2nd was an issue where the `canister.declarations.bindings` array in `dfx.json` was not always working as expected. 

Fixes #2615 

# How Has This Been Tested?

e2e tests have been provided and it has been manually tested in a new `dfx` project.

# Checklist:

  - No documentation changes necessary

### DIFF
--- a/src/dfx/src/lib/builders/mod.rs
+++ b/src/dfx/src/lib/builders/mod.rs
@@ -118,16 +118,6 @@ pub trait CanisterBuilder {
                 )
             })?;
         }
-        std::fs::create_dir_all(&generate_output_dir).with_context(|| {
-            format!(
-                "Failed to create dir: {}",
-                generate_output_dir.to_string_lossy()
-            )
-        })?;
-
-        let generated_idl_path = self.generate_idl(pool, info, config)?;
-
-        let (env, ty) = check_candid_file(generated_idl_path.as_path())?;
 
         let bindings = info
             .get_declarations_config()
@@ -145,6 +135,17 @@ pub trait CanisterBuilder {
             );
         }
 
+        std::fs::create_dir_all(&generate_output_dir).with_context(|| {
+            format!(
+                "Failed to create dir: {}",
+                generate_output_dir.to_string_lossy()
+            )
+        })?;
+
+        let generated_idl_path = self.generate_idl(pool, info, config)?;
+
+        let (env, ty) = check_candid_file(generated_idl_path.as_path())?;
+
         // Typescript
         if bindings.contains(&"ts".to_string()) {
             let output_did_ts_path = generate_output_dir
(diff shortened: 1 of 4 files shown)


src/dfx/src/lib/builders/mod.rs
```diff
@@ -118,16 +118,6 @@ pub trait CanisterBuilder {
                 )
             })?;
         }
-        std::fs::create_dir_all(&generate_output_dir).with_context(|| {
```
nathanosdev, line 121: This code was generating the declarations directory and the DID file before checking if the bindings array is empty.
There's some code lower down in this file that deletes the DID file, but we never get that far because of the early return for an empty bindings array.

So the old flow was like this:
- create directory
- create did file
- if bindings array is empty, return early
- create JS bindings if applicable
- create TS bindings if applicable
- delete DID bindings if not applicable

Now it's like this:
- if bindings array is empty, return early
- create directory
- create did file
- create JS bindings if applicable
- create TS bindings if applicable
- delete DID bindings if not applicable

src/dfx/src/lib/builders/mod.rs
```diff
@@ -174,78 +177,8 @@ pub trait CanisterBuilder {
                 )
             })?;
             eprintln!("  {}", &output_did_js_path.display());
-            // index.js
-            let mut language_bindings = crate::util::assets::language_bindings()
```
nathanosdev, line 178: Moved this to a separate function that takes a param "ts" or "js" and will only generate the files for the provided extension.

PR #2663, opened 2022-10-06
feat: Make platform-specific code Windows-compatible

adamspofford-dfinity:
The road to Windows support starts with cleaning up the Unix-env assumptions. Thankfully there are not many.

- `$HOME` and `/tmp` usage is replaced with dirs_next functions on Windows, including for dotfolder assumptions like `.cache`. These will wind up in `~\AppData` locations.

- File modes are conditionally compiled in; they are only used to set the execute permission, which is on-by-default in Windows.

- Where we require an explicit file path to the pkcs11 library, the correct default path for the platform is displayed in help strings.

- Where directory symlinks are created on Linux, junctions are created on Windows, as symlinks require Windows to be in developer mode.

- The color support check has been replaced from an OS check to an actual color support check (we had previously been giving the wrong answer for alternate macOS terminals like iTerm)

This PR goes out of its way not to change any existing behavior. dirs-next is only used on Windows; dfx still does the wrong thing on macOS. Perhaps whether to migrate to Apple's official application directories should be revisited in a future PR. 

Also, no CI checks are added yet. 'Builds on Windows' is rather a low bar; things like $HOME can't easily be caught in CI until the code is actually ready to run the full battery of e2e tests.

Implements SDK-751.


src/dfx/src/config/cache.rs
```diff
@@ -181,22 +191,25 @@ pub fn install_version(v: &str, force: bool) -> DfxResult<PathBuf> {
             file.unpack_in(temp_p.as_path())
                 .context("Failed to unpack archive asset.")?;
 
-            let full_path = temp_p.join(file.path().context("Failed to get file path.")?);
-            let mut perms = std::fs::metadata(full_path.as_path())
-                .with_context(|| {
+            #[cfg(unix)]
```
sesi200, line 194:
```suggestion
            // Not necessary on Windows as these permissions are set by default.
            #[cfg(unix)]
```

src/dfx/src/config/cache.rs
```diff
@@ -211,19 +224,22 @@ pub fn install_version(v: &str, force: bool) -> DfxResult<PathBuf> {
...
-                    "Failed to read file metadata for {}.",
-                    dfx.to_string_lossy()
-                )
-            })?
-            .permissions();
-        perms.set_mode(EXEC_READ_USER_ONLY_PERMISSION);
-        std::fs::set_permissions(&dfx, perms).with_context(|| {
-            format!("Failed to set file metadata for {}.", dfx.to_string_lossy())
-        })?;
+        #[cfg(unix)]
```
sesi200, line 227:
```suggestion
        // Not necessary on Windows as these permissions are set by default.
        #[cfg(unix)]
```

src/dfx/src/lib/dist.rs
```diff
@@ -9,12 +9,11 @@ use indicatif::{ProgressBar, ProgressDrawTarget};
 use semver::Version;
 use std::fs;
 use std::io::Write;
-use std::path::Path;
 use tar::Archive;
 
 pub static DEFAULT_RELEASE_ROOT: &str = "https://sdk.dfinity.org";
-pub static CACHE_ROOT: &str = ".cache/dfinity/versions/";
-pub static DOWNLOADS_DIR: &str = ".cache/dfinity/downloads/";
+pub static CACHE_ROOT: &str = "dfinity/versions/";
```
sesi200, line 15: `CACHE_ROOT` feels like it's becoming a misleading name, but I don't really have a better idea

src/dfx/src/lib/identity/identity_manager.rs
```diff
@@ -499,16 +509,22 @@ To create a more secure identity, create and use an identity that is protected b
...
-        .map_err(|_| DfxError::new(IdentityError::CannotFindHomeDirectory()))?;
-    let root = config_root.unwrap_or(home);
-
-    Ok(PathBuf::from(root)
-        .join(".dfinity")
-        .join("identity")
-        .join("creds.pem"))
+fn get_legacy_creds_pem_path() -> DfxResult<Option<PathBuf>> {
+    if cfg!(windows) {
+        Ok(None)
```
sesi200, line 514:
```suggestion
        // There was no Windows support back then.
        Ok(None)
```

src/dfx/src/lib/manifest.rs
```diff
@@ -149,22 +148,26 @@ pub fn get_latest_release(release_root: &str, version: &Version, arch: &str) ->
     archive
         .unpack(&current_exe_dir)
         .with_context(|| format!("Failed to unpack to {}.", current_exe_dir.to_string_lossy()))?;
-    b.set_message("Setting permissions");
-    let mut permissions = fs::metadata(&current_exe_path)
-        .with_context(|| {
+    #[cfg(unix)]
```
sesi200, line 151:
```suggestion
    // Not necessary on Windows as these permissions are set by default.
    #[cfg(unix)]
```

src/dfx/src/config/cache.rs
adamspofford-dfinity: I think these go in the wrong direction re: which is base knowledge and which requires clarifying, but comments here can't hurt.

sesi200 (approved):

src/dfx/Cargo.toml
```diff
@@ -88,6 +89,11 @@ url = "2.1.0"
 walkdir = "2.2.9"
 wasmparser = "0.87.0"
 which = "4.2.5"
+dirs-next = "2.0.0"
```
ericswanson-dfinity, line 92: We already have `directories-next` as a dependency.  Can we get what we need from there rather than also adding `dirs-next` ?

ericswanson-dfinity (changes requested):

src/dfx/src/lib/config.rs
```diff
@@ -6,10 +6,19 @@ use std::path::PathBuf;
...
+    let p = {
+        let home = std::env::var_os("HOME").context("Failed to resolve 'HOME' env var.")?;
+        let root = config_root.unwrap_or(home);
+        PathBuf::from(root).join(".config").join("dfx")
+    };
+    #[cfg(windows)]
+    let p = config_root.map_or_else(
+        || dirs_next::config_dir().unwrap().join("dfx"),
+        PathBuf::from,
+    );
```
ericswanson-dfinity, line 21: Suggest using `ProjectDirs::from("org", "dfinity", "dfx")` from directories_next here, so that on Windows these files will sit in the same directory as networks.json

ericswanson-dfinity:
I recommend to add windows to the build matrix in at least one of these two places, preferably both.  While "Builds on Windows" may be a low bar, our codebase prior to this PR would have failed even that.
- https://github.com/dfinity/sdk/blob/master/.github/workflows/e2e.yml#L27
- https://github.com/dfinity/sdk/blob/master/.github/workflows/publish.yml#L33
-

src/dfx/Cargo.toml
ericswanson-dfinity: The rest (or most of them) of the dependencies are in alphabetical order.  Adding this in alphabetical order might have helped.

adamspofford-dfinity:
An additional trap there is that it doesn't actually build *on* Windows, only *for* Windows, due to our use of Bash scripts in build.rs. So this would go as part of a separate task using `cross` instead of as part of an existing matrix, *or* the build script would be ported to purely Rust, which I kind of want to do anyway, but which should probably be a separate PR.

src/dfx/src/util/mod.rs
```diff
@@ -343,6 +349,14 @@ pub fn pretty_thousand_separators(num: String) -> String {
         .collect::<_>()
 }
 
+#[cfg_attr(not(windows), allow(unused))]
```
ericswanson-dfinity, line 352: Could we remove this line by referencing `project_dirs` from [here](https://github.com/dfinity/sdk/blob/master/src/dfx/src/config/dfinity.rs#L886)?
```suggestion
```

ericswanson-dfinity (approved):

PR #2666, opened 2022-10-07
fix: dfx generate now works without a canister id

### LINKED ISSUE #2665: dfx generate is not compiling the candid if the canister is not deployed
Currently, if you have a child canister that is in the project, but does not get deployed directly by dfx and rather a parent canister, you cannot generate the bindings. The generation only after creating the canister id. A usual workaround is specifying the management or anonymous principal id in canister_ids.json, but as generate only reads the local ids and they are always ephemeral, it requires manual intervention any time dfx is reset

ozwaldorf:
# Description

Dfx previously would refuse to generate bindings unless a canister id was specified. We can easily do it without, and have the ability to generate without deploying everything locally

Fixes #2665 

# How Has This Been Tested?

Tested against a local project. You can see the cargo run generate with the changes, and the current version of dfx generate

image

# Checklist:


sesi200 (changes requested):
Thank you very much for starting this! Two more things:
- To show that it works without canister id, please adapt the tests in `e2e/tests-dfx/generate.bash` to not do `dfx canister create --all` and `dfx canister install --all`.
- This panics with Motoko projects. To reproduce, run `dfx new hello`, `cd hello`, and then `dfx generate`

If this is too much to fix for you, I can also take over from here. Just let me know

ozwaldorf:
> Thank you very much for starting this! Two more things:
> 
> * To show that it works without canister id, please adapt the tests in `e2e/tests-dfx/generate.bash` to not do `dfx canister create --all` and `dfx canister install --all`.
> * This panics with Motoko projects. To reproduce, run `dfx new hello`, `cd hello`, and then `dfx generate`
> 
> If this is too much to fix for you, I can also take over from here. Just let me know

Hmm, so did some digging and it seems like the dfx build process is dependent on the canister id to be initialized for grabbing information with it for a build to succeed. I was able to switch the entire build modal and dependency graph to use the config canister name instead of the id to pull in information, but still it seems the motoko compiler requires the id present as well as dfx. This might be still a good refactor though, as would allow running `dfx build` for non motoko canisters without having the canister id initialized like generate can now.

With this information there's some options (feel free to suggest any others)
1) we keep the requirement for motoko to be generated, and just allow non motoko to generate without canister ids (short term - easy)
2) we remove the prebuild for motoko and decouple generating candid from generating bindings (also short term - easy)
3) we remove the canister id requirement from motoko's compiler and allow building without deploying (mid term - need more research)

I personally align with 2, as other languages already have the concept of generating the did and generating the bindings separately. For example, with rust a common practice is having a single rust test that writes out the did file, completely separate from building or deploying 

#3 is probably the cleanest solution though, where motoko could compile for a canister that is not initialized yet

sesi200:
Hmm, it looks like option 3 could already be possible (`moc` is the Motoko compiler):
``` bash
❯ $(dfx cache show)/moc --idl src/hello_backend/main.mo
❯ cat main.did 
service : {
  greet: (text) -> (text) query;
}
```
This minimal option doesn't quite work with imports (e.g. by adding `import Trie "mo:base/Trie";` at the top of the .mo file), but I think we can get there just by comparing with the normal Motoko build command. Also, if we know we'll only use the .did output and nothing else we can inject dummy canister IDs for dependency canister IDs, so `moc` should also be satisfied in the end.

If you're willing to do the work for option 3 I'd be very grateful, but I can also live with the other options and do the rest when I get around to it.

ozwaldorf:
replaced by #2988

PR #2667, opened 2022-10-08
feat: Add a -y flag to deploy/install

adamspofford-dfinity:
Implements SDK-701. Automatically confirms y/n checks.


sesi200 (approved):

dfx-json:
once a release is cut that includes this change, can we please let the folks at Aikin know?

PR #2670, opened 2022-10-10
fix: dfx schema should not break on invalid dfx.json

### LINKED ISSUE #2437: `dfx schema` requires a valid `dfx.json`
I was setting up a new remote canister and was trying to see if I could omit some fields.

I wanted to view the schema to get a better sense of what I was doing but I couldn’t because I had omitted some required fields. The file was still valid JSON.

It seems like an unnecessary coupling to not be able to view the schema because of the state of the file.

sesi200:
# Description

`dfx schema` should not refuse to work when a broken dfx.json is present.

Fixes #2437, [SDK-667](https://dfinity.atlassian.net/browse/SDK-667)

# How Has This Been Tested?

Added e2e.

# Checklist:


lwshang (approved):

src/dfx/src/main.rs
```diff
@@ -201,7 +201,10 @@ fn main() {
                 Err(e) => Err(e),
             }
         }
-        Err(e) => Err(e),
+        Err(e) => match command {
+            commands::Command::Schema(_) => commands::exec_without_env(command),
+            _ => Err(e),
+        },
```
ericswanson-dfinity, line 207: If the failure of the `schema` command can't depend on the failure to construct an Environment, what's the purpose of displaying the earlier error?
```suggestion
        Err(_) => match command {
            commands::Command::Schema(_) => commands::exec_without_env(command),
            Err(e) => Err(e),
        },
```

sesi200: I think you're misreading the code (meaning I have to clarify it somehow). The match doesn't match on the result of the schema command -  it figures out if the command being run can be executed without having an Environment available.

A different way of writing a similar thing would be: 
```suggestion
        Err(original_error) => commands::exec_without_env(command).or_else( || Err(original_error)),
        },
```

PR #2678, opened 2022-10-12
feat: Warn if an unencrypted identity is used

adamspofford-dfinity:
This logs a warning if the user performs a mainnet action while using an unencrypted identity. Implements SDK-641.


e2e/tests-dfx/identity.bash
```diff
@@ -160,3 +160,13 @@ teardown() {
     assert_command dfx canister call --output idl e2e_project_frontend retrieve '("B")'
     assert_eq '(blob "hello")'
 }
+
+@test "using an unencrypted identity on mainnet provokes a warning" {
+    assert_command dfx ledger balance --network ic
+    assert_match WARN "$stderr"
```
ericswanson-dfinity, line 166: Please include some part of the expected warning message.  This would pass if there were any warning, not just the one added in this PR.

This suggestion specifies the whole warning message, but feel to replace parts of it with `.*` as long as assertion identifies the expected warning message.
```suggestion
    assert_match "WARN: The default identity is not stored securely. Do not use it to control a lot of cycles/ICP. Create a new identity with \`dfx identity create\` and use it in mainnet-facing commands with the \`--identity\` flag" "$stderr"
```

e2e/tests-dfx/identity.bash
```diff
@@ -160,3 +160,13 @@ teardown() {
...
+    assert_command "${BATS_TEST_DIRNAME}/../assets/expect_scripts/init_alice_with_pw.exp"
+    assert_command "${BATS_TEST_DIRNAME}/../assets/expect_scripts/get_ledger_balance.exp"
+    dfx identity new bob --disable-encryption
+    assert_command dfx ledger balance --network ic --identity bob
+    assert_match WARN "$stderr"
```
ericswanson-dfinity, line 171:
```suggestion
    assert_match "WARN: The bob identity is not stored securely. Do not use it to control a lot of cycles/ICP. Create a new identity with \`dfx identity create\` and use it in mainnet-facing commands with the \`--identity\` flag" "$stderr"
```

script.exp
```diff
@@ -0,0 +1,150 @@
+#!/usr/local/bin/expect -f
```
ericswanson-dfinity, line 1: It looks like this file was not intended to be added so source control.

ericswanson-dfinity (changes requested):

script.exp
adamspofford-dfinity: Whoops.

ericswanson-dfinity (approved):

ericswanson-dfinity (approved):

PR #2686, opened 2022-10-14
chore: remove unused variable

nmattia:
This removes a variable that wasn't being used (warnings were silenced because name was `_all`).


sesi200 (approved):

nmattia:
> thread 'main' panicked at 'Cannot parse version.: Error("unexpected character '+' after major version number")', src/dfx/src/config/mod.rs:10:51

Don't think that's me

sesi200:
> unexpected character '+' after major version number

AAARRGH not that again 🤦 that's definitively not you. let me see what I can do

PR #2687, opened 2022-10-14
fix: do not build remote canisters

sesi200:
# Description

If a canister config specifies a “remote” canister ID, like the NNS subnet canisters do, or if the canister is deployed from another directory locally, it should skip the build step.

Fixes [SDK-766](https://dfinity.atlassian.net/browse/SDK-766)

# How Has This Been Tested?

Added checks to CI that remote canisters are not processed in:
- `dfx deploy`
- `dfx build`
- `dfx generate`

# Checklist:

### DIFF
--- a/src/dfx/src/commands/generate.rs
+++ b/src/dfx/src/commands/generate.rs
@@ -23,6 +23,7 @@ pub struct GenerateOpts {
 
 pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
     let env = create_agent_environment(env, opts.network.network)?;
+    let log = env.get_logger();
 
     // Read the config.
     let config = env.get_config_or_anyhow()?;
@@ -48,11 +49,11 @@ pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
         })
         .collect();
 
-    // Get pool of canisters to build
     let canister_pool = CanisterPool::load(&env, false, &canisters_to_load)?;
 
-    // This is just to display an error if trying to generate before creating the canister.
+    // This is just to display an error if trying to generate before creating the canister(s).
     let store = CanisterIdStore::for_env(&env)?;
+
     // If generate for motoko canister, build first
     let mut build_before_generate = Vec::new();
     for canister in canister_pool.get_canister_list() {
@@ -62,7 +63,7 @@ pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
             if info.is_motoko() {
                 build_before_generate.push(canister_name.to_string());
                 trace!(
-                    env.get_logger(),
+                    log,
                     "Found Motoko canister '{}' - will have to build before generating IDL.",
                     canister_name
                 );
@@ -81,15 +82,12 @@ pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
         .map(|v| v.is_empty())
         .unwrap_or(true)
     {
-        slog::info!(
-            env.get_logger(),
-            "Building canisters before generate for Motoko"
-        );
+        slog::info!(log, "Building canisters before generate for Motoko");
         let runtime = Runtime::new().expect("Unable to create a runtime");
-        runtime.block_on(canister_pool.build_or_fail(env.get_logger(), &build_config))?;
+        runtime.block_on(canister_pool.build_or_fail(log, &build_config))?;
     }
 
-    for canister in canister_pool.with_canisters_to_build(&generate_config) {
+    for canister in canister_pool.canisters_to_build(&generate_config) {
         canister.generate(&canister_pool, &generate_config)?;
     }
 
--- a/src/dfx/src/commands/generate.rs
+++ b/src/dfx/src/commands/generate.rs
@@ -76,11 +76,11 @@ pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
     let generate_config =
         BuildConfig::from_config(&config)?.with_canisters_to_build(canisters_to_generate);
 
-    if !build_config
+    if build_config
         .canisters_to_build
         .as_ref()
-        .map(|v| v.is_empty())
-        .unwrap_or(true)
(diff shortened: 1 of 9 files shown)


src/dfx/src/commands/generate.rs
```diff
@@ -31,40 +33,62 @@ pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
...
-    if build_before_generate {
-        slog::info!(
-            env.get_logger(),
-            "Building canisters before generate for Motoko"
-        );
+    if !build_config
+        .canisters_to_build
+        .as_ref()
+        .map(|v| v.is_empty())
+        .unwrap_or(true)
```
ericswanson-dfinity, line 83: I had some trouble understanding which direction this was going, in particular the logic around 'what if it's empty / what if it's None)', especially with respect to `.unwrap_or(true)`.  What would you think about this instead?
```suggestion
    if build_config
        .canisters_to_build
        .as_ref()
        .map(|v| !v.is_empty())
        .unwrap_or(false)
```

src/dfx/src/lib/builders/mod.rs
```diff
@@ -365,6 +365,8 @@ pub struct BuildConfig {
     pub lsp_root: PathBuf,
     /// The root for all build files.
     pub build_root: PathBuf,
+    /// If only a subset of canisters should be built, then canisters_to_build contains these canisters' names.
```
ericswanson-dfinity, line 368: So `None` means to build all the canisters?

src/dfx/src/lib/models/canister.rs
```diff
@@ -238,15 +239,48 @@ impl CanisterPool {
...
+    fn step_prebuild_all(&self, log: &Logger, build_config: &BuildConfig) -> DfxResult<()> {
+        // moc expects all .did files of dependencies to be in <output_idl_path> with name <canister id>.did.
+        // Because remote canisters don't get built (and the did file not output in the right place) the .did files have to be copied over manually.
+        for canister in self.canisters.iter().filter(|c| c.info.is_remote()) {
+            let maybe_from = if let Some(remote_candid) = canister.info.get_remote_candid() {
+                Some(remote_candid)
+            } else {
+                canister.info.get_output_idl_path()
+            };
+            if maybe_from.is_some() && maybe_from.as_ref().unwrap().exists() {
```
ericswanson-dfinity, line 251: I think it would be more clear to have different warnings for these two cases:
```suggestion
            if let Some(from) = maybe_from.as_ref() {
                if from.exists() {
                } else {
                    warn!( something about idl not found at path )
                }
            } else {
                warn!( something about idl path not specified )
            }
```

src/dfx/src/lib/models/canister.rs
```diff
@@ -238,15 +239,48 @@ impl CanisterPool {
...
+                let from = maybe_from.unwrap();
+                let to = build_config.idl_root.join(format!(
+                    "{}.did",
+                    canister.info.get_canister_id()?.to_text()
+                ));
+                if std::fs::copy(&from, &to).is_err() {
+                    warn!(
+                                log,
+                                "Failed to copy remote canister candid from {} to {}. This may produce errors during the build.",
```
ericswanson-dfinity, line 260: is this always the remote canister idl?

src/dfx/src/lib/operations/canister/deploy_canisters.rs
```diff
@@ -193,16 +194,18 @@ async fn register_canisters(
 #[context("Failed to build call canisters.")]
 async fn build_canisters(
     env: &dyn Environment,
-    canister_names: &[String],
+    referenced_canisters: &[String],
```
ericswanson-dfinity, line 197: It looks like you're taking care of this here, but I wanted to mention: one of the things to make sure of here is that referenced canister ids are still passed to moc, and in environment variables, for remote canisters.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/builders/mod.rs
sesi200: Correct. Expanding the comment to explain that as well.

src/dfx/src/commands/generate.rs
sesi200: much better!

src/dfx/src/lib/models/canister.rs
sesi200: added separate warnings

src/dfx/src/lib/models/canister.rs
sesi200: No, you're right, we can't be sure. Updating the message

src/dfx/src/lib/operations/canister/deploy_canisters.rs
sesi200: Yes, this is what I do to still supply the ids. This is tested for moc in `remote.bash` in test `build + install + call -- remote`. I also added a test that makes sure the environment variables are set correctly, too.

src/dfx/src/commands/generate.rs
```diff
@@ -31,40 +33,62 @@ pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
...
+    let canisters_to_generate = canisters_to_load
+        .clone()
+        .into_iter()
+        .filter(|canister_name| {
+            !matches!(
+                config
+                    .get_config()
+                    .get_remote_canister_id(canister_name, &env.get_network_descriptor().name),
+                Ok(Some(_))
+            )
```
ericswanson-dfinity, line 48: There's a `is_remote_canister` method for convenience:
```suggestion
            config
                .get_config()
                .is_remote_canister(canister_name, &env.get_network_descriptor().name)
                .unwrap_or(false)
```

ericswanson-dfinity (approved):
There are two places where I'd suggest filtering with `is_remote_canister` rather than matching `get_remote_canister_id` to `Ok(Some(_))`.  Up to you.

PR #2688, opened 2022-10-14
feat: run build command in /bin/sh

nmattia:
This stops running commands directly and instead runs commands in `/bin/sh`.

Before this, build commands were parsed using the third-party crate `shell_words` (which tries to replicate shell argument parsing) and the command executable would be called directly.

With these changes, commands are run via `sh`, which is more consistent with other tools like `npm` and allows shell constructs (`foo && bar | baz`). This also simplifies the code a bit.

This won't work on pure Windows but will work on WSL since `sh` is POSIX. If we do want to run anything on pure Windows then we'll probably need some local config like npm: https://docs.npmjs.com/cli/v8/using-npm/config#script-shell

**Note**: I haven't updated the documentation since I assume most people will expect commands to be shell commands by default; happy to update it though.

# How Has This Been Tested?

I've added an e2e test that uses shell constructs.

# Checklist:

### DIFF
--- a/src/dfx/src/lib/builders/custom.rs
+++ b/src/dfx/src/lib/builders/custom.rs
@@ -123,6 +123,12 @@ impl CanisterBuilder for CustomBuilder {
         let vars = super::environment_variables(info, &config.network_name, pool, &dependencies);
 
         let mut add_candid_service_metadata = false;
+
+        // If there's any build command, add metadata
+        if !build.is_empty() {
+            add_candid_service_metadata = true;
+        }
+
         for command in build {
             info!(
                 self.logger,
@@ -131,15 +137,8 @@ impl CanisterBuilder for CustomBuilder {
                 command
             );
 
-            // First separate everything as if it was read from a shell.
-            let args = shell_words::split(&command)
-                .with_context(|| format!("Cannot parse command '{}'.", command))?;
-            // No commands, noop.
-            if !args.is_empty() {
-                add_candid_service_metadata = true;
-                run_command(args, &vars, info.get_workspace_root())
-                    .with_context(|| format!("Failed to run {}.", command))?;
-            }
+            run_shell_command(&command, &vars, info.get_workspace_root())
+                .with_context(|| format!("Failed to run {}.", command))?;
         }
 
         let mut file =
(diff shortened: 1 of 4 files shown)


src/dfx/src/lib/builders/custom.rs
```diff
@@ -202,16 +201,10 @@ impl CanisterBuilder for CustomBuilder {
...
-        .canonicalize()
-        .or_else(|_| which::which(command_name))
-        .map_err(|_| anyhow!("Cannot find command or file {command_name}"))?;
-    let mut cmd = Command::new(&canonicalized);
-
-    cmd.args(arguments)
+fn run_shell_command(cmd_str: &str, vars: &[super::Env<'_>], cwd: &Path) -> DfxResult<()> {
+    let mut cmd = Command::new("sh");
+    cmd.arg("-c")
+        .arg(cmd_str)
```
ericswanson-dfinity, line 207: Some of the test failures are due to the fact that the previous code goes to some trouble to make sure that if you put `<name of executable in your project directory>` in your build script, dfx calls it, rather then reporting an error because you didn't put `./<name of executable in your project directory>`.

That's what `cwd.join(command_name).canonicalize()` is about.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/builders/custom.rs
nmattia: I see! that makes sense. I'll re-add something like that.

nmattia:
> Please mark Unix-specific code cfg(unix) and provide a cfg(windows) version.

Sure! Cmd.exe or powershell?

> Though the inherent platform-dependence of a resulting canister project makes me wonder if this is a good change in the first place.

For canister developers that stick to simple commands (i.e. what's currently working) this won't change a thing. For canister developers that need more, it'll be a good option. So I think it makes sense, but @adamspofford-dfinity @ericswanson-dfinity and @sesi200 please confirm before I go ahead with the changes.

adamspofford-dfinity:
Definitely PowerShell; cmd is deprecated, and a Unix user can use PowerShell Core the same as a Windows user can use Git Bash given a platform-native script. 

Additionally, make sure build and post-install tasks are using the same system, and then if the shell_words crate is no longer being used, remove it as a dependency.

nmattia:
hi there! didn't drop the ball but was a bit snowed under. Will pick this up this week.

nmattia:
Quick update:

@adamspofford-dfinity and I clarified what needed to be done esp. on Windows, here's what we came up with:

* `dfx` should use `pwsh` when available, `powershell` otherwise (idea is to have sensible default)
* if the e.g. `post_build` is a file that exists, then execute it; otherwise don't try to parse it but feed it to `$SHELL -c <here>` (which works with powershell as well). This means that the behavior changes in case of e.g. `scripts/foo 2` (right now it'd call `scripts/foo` with argument `2`, whereas changes discussed here would feed it as in to the shell which will fail)

I tried to get my hands on a windows machine for a few days but unfortunately don't have any more time to dedicate to this right now, so I'll close this for now.

PR #2699, opened 2022-10-17
chore: Change update-frontend-canister.sh to use Docker

adamspofford-dfinity:
Rust builds are not reproducible across platforms, even for neutral targets like wasm. #2557 thus unfortunately cannot be run from Macs without using a docker image, and may even have problems across Linux distros. This PR moves the bulk of the build logic into a Dockerfile and has the script use it as a buildkit. 

It also restricts the check to only run when frontend canister code is actually changed, to reduce CI failure spam.

The only change made to the remap logic is that github.com-1ecc6299db9ec823 is no longer shortened to github - if it were not github.com-1ecc6299db9ec823, that'd be extremely important information, because an alternative registry would have been in use. I assume this hash was interpreted as a Git commit hash which could change, but it is a stable string since 2018 edition.


scripts/update-frontend-canister.Dockerfile
```diff
@@ -0,0 +1,16 @@
+ARG RUST_VERSION
+FROM rust:${RUST_VERSION} AS builder
+RUN cargo install ic-wasm --version 0.2.0
```
ericswanson-dfinity, line 3: Please remove the corresponding `cargo install ic-wasm` from the github workflow

ericswanson-dfinity (changes requested):

scripts/update-frontend-canister.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY . /build
+WORKDIR /build
+RUN export RUSTFLAGS="--remap-path-prefix $CARGO_HOME=/cargo" && \
+    cargo build -p ic-frontend-canister --release --target wasm32-unknown-unknown
```
ericswanson-dfinity, line 7: should this have `--locked` ?

ericswanson-dfinity:
Is it expected that running `./scripts/update-frontend-canister.sh` from this branch will result in no change to `src/distributed/assetstorage.wasm.gz` ?   I am still seeing differences when I run `./scripts/update-frontend-canister.sh ` locally (though they seem smaller, as measured by file size)
```
$ uname -a
Darwin Eric-Swansons-MacBook-Pro.local 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64

$ ls -l *.wasm
-rwxr-xr-x  1 ericswanson  staff  1122801 Oct 17 17:31 from-running-script.wasm
-rwxr-xr-x  1 ericswanson  staff  1123053 Oct 17 17:29 from-branch.wasm

```

adamspofford-dfinity:
Didn't realize it was important. The only change I wanted to make was not remapping any alternate registries; I don't see a reason the default registry can't be remapped the existing way.

ericswanson-dfinity (changes requested):
Would you please add a note to .github/CONTRIBUTING about what to do when updating the frontend canister?  It should include something about updating the changelog with the module hash and a link to the PR (since the commit won't be available).  It should also mention how to verify that a given module hash corresponds to the code at a particular commit.

When I run scripts/update-frontend-canister.sh, I get a change to assetstorage.wasm.gz .  So this PR should probably update the canister and document the new module hash (in turn, showing an example of the upgrade process)

ericswanson-dfinity (approved):

PR #2706, opened 2022-10-21
chore: Rewrite build script in Rust

adamspofford-dfinity:
The build script currently requires a shell script to curl all the bundled assets and pack them into tarballs. This PR rewrites the script in Rust, which not only does all the fetching in parallel but supports Windows. 

The format of the sources manifest has been changed from a shell script to a .toml file, as the latter is far more easily consumed from Rust and will help eliminate some edge cases like `motoko-base`.

Given Windows assets, DFX now builds on Windows. Linux assets are substituted as a stopgap since technically any external program can be used through WSL, but we probably want to have a proper Windows asset set eventually.


ericswanson-dfinity (changes requested):
Is the change to assetstorage.wasm.gz intentional?

adamspofford-dfinity:
Because we have stuck them in a workspace, the frontend canister crate shares any updates in Cargo.lock from other crates.

adamspofford-dfinity:
This one I can revert - itertools only got updated because a previous version of this PR depended on it. But prepare for more such hash-changes.

ericswanson-dfinity:
Okay.  If we're updating the asset canister, please note this in the changelog, including the new module hash.

ericswanson-dfinity (approved):

PR #2719, opened 2022-10-25
feat: sns deploy network flag

bitdivine:
# Description
Currently `dfx sns deploy` works only on localhost.  This PR enables the standard `--network` flag and passes it through.

# How Has This Been Tested?
- The `dfx sns deploy` test is currently disabled in CI to to version alignment issues between the IC and SDk repos, however I have added a comment to exercise the --network flag when that test is re-enabled.
- I have tested `dfx sns deploy --network local` using this code in the snsdemo5.  At least 3 other people have gone through snsdemo5 successfully.
- I could not test the same against testnets at the time due to boundary node issues.  Once all the relevant parts are working and merged, probably in one week, the snsdemo can be exercised against testnets.

# Checklist:


src/dfx/src/lib/sns/deploy.rs
```diff
@@ -1,30 +1,54 @@
...
 #[context("Failed to deploy SNS with config: {}", path.display())]
-pub fn deploy_sns(env: &dyn Environment, path: &Path) -> DfxResult<String> {
+pub fn deploy_sns(
+    env: &dyn Environment,
+    path: &Path,
+    network: Option<String>,
+) -> DfxResult<String> {
+    // For networks other than ic and local we need a provider URL:
+    let agent_environment = create_agent_environment(env, network.clone())?;
+    let network_descriptor = agent_environment.get_network_descriptor();
```
ericswanson-dfinity, line 22: Since we're only using the agent environment to get the network descriptor, we can instead create the network descriptor directly.

```
        let network_descriptor = create_network_descriptor(
            env.get_config(),
            env.get_networks_config(),
            network,
            None,
            LocalBindDetermination::ApplyRunningWebserverPort,
        )?;
```

ericswanson-dfinity (changes requested):

src/dfx/src/lib/models/canister_id_store.rs
```diff
@@ -177,6 +177,10 @@ impl CanisterIdStore {
         }
         self.save_ids()
     }
+
+    pub fn get_path(&self) -> Option<&PathBuf> {
+        self.path.as_ref()
+    }
```
ericswanson-dfinity, line 183: Someone told me a long time ago that best practice is to pass around / return `&Path` rather than `&PathBuf`
```suggestion
    pub fn get_path(&self) -> Option<&Path> {
        self.path.as_deref()
    }
```

src/dfx/src/lib/sns/deploy.rs
```diff
@@ -1,30 +1,54 @@
...
+    // The underlying sns binary recognises "local", "ic" and provider URLs, so we need one of those:
+    let sns_network_param = match network.as_ref().map(|network| network.as_ref()) {
+        None => "local",
+        Some("ic") => "ic",
+        Some("local") => "local",
+        Some(url) if url::Url::parse(url).is_ok() => url,
+        _ => network_descriptor.first_provider().with_context(|| {
```
ericswanson-dfinity, line 29: What's an example value that would meet the conditions asked for in the above code?
 1) doesn't parse as a valid url
 2) is then intuitively ignored in order to instead use the network descriptor's first provider
I guess I would expect the following, and no `_ =>` case
```
Some(url) -> url::Url::parse(url)?,
```

bitdivine:
Closing as new work on this is planned to be in another repository.

PR #2741, opened 2022-11-04
fix: Add ECParameters to generated secp256k1 keys

adamspofford-dfinity:
Not required for DFX to function, but required to be able to run `openssl ec`.


dfx-json (approved):
Nice work with the quick turnaround!

PR #2752, opened 2022-11-08
fix: default not shrink for custom canisters

lwshang:
# Description

For custom canisters, `dfx` should better not do shrink by default. Developer still can enable it in `dfx.json`.

# How Has This Been Tested?

An e2e test was added.

# Checklist:


src/dfx/src/lib/canister_info.rs
```diff
@@ -202,7 +202,7 @@ impl CanisterInfo {
         &self.args
     }
 
-    pub fn get_shrink(&self) -> bool {
+    pub fn get_shrink(&self) -> Option<bool> {
         self.shrink
     }
```
ericswanson-dfinity, line 207: I'm having a hard time following `!matches!(canister_info.get_shrink(), Some(false))`.  How about this?
```suggestion
    pub fn get_shrink(&self) -> bool {
        match self.type_specific {
            CanisterTypeProperties::Rust { .. } | CanisterTypeProperties::Motoko => {
                self.shrink.unwrap_or(true)
            }
            CanisterTypeProperties::Custom { .. } | CanisterTypeProperties::Assets { .. } => {
                self.shrink.unwrap_or(false)
            }
        }
    }
```

adamspofford-dfinity: For future reference, `== Some(true)` works fine too.

lwshang: That `matches!` actually came from `cargo clippy`.

While moving such canister type specific logic into the general `CanisterInfo` is not good for encapsulation.
I would still keep the logic in each canister type builder and replace `matches` with more readable syntax.

src/dfx/src/lib/builders/motoko.rs
```diff
@@ -166,7 +166,9 @@ impl CanisterBuilder for MotokoBuilder {
         };
         motoko_compile(&self.logger, cache.as_ref(), &params)?;
 
-        if canister_info.get_shrink() {
+        // Default to shrink Motoko canister unless explicitly disable
+        let shrink = canister_info.get_shrink() != Some(false);
```
ericswanson-dfinity, line 170: This is better, but still hard to read.  Using `unwrap_or` removes the need for the comment, even.
```suggestion
        let shrink = canister_info.get_shrink().unwrap_or(true);
```

src/dfx/src/lib/builders/custom.rs
```diff
@@ -140,8 +140,10 @@ impl CanisterBuilder for CustomBuilder {
             }
         }
 
+        // Default **NOT** to shrink custom canister unless explicitly enable
+        let shrink = info.get_shrink() == Some(true);
```
ericswanson-dfinity, line 144:
```suggestion
        let shrink = info.get_shrink().unwrap_or(false);
```

src/dfx/src/lib/builders/rust.rs
```diff
@@ -92,7 +92,9 @@ impl CanisterBuilder for RustBuilder {
         );
         let output = cargo.output().context("Failed to run 'cargo build'. You might need to run `cargo update` (or a similar command like `cargo vendor`) if you have updated `Cargo.toml`, because `dfx build` uses the --locked flag with Cargo.")?;
 
-        if canister_info.get_shrink() {
+        // Default to shrink Rust canister unless explicitly disable
+        let shrink = canister_info.get_shrink() != Some(false);
+        if shrink {
```
ericswanson-dfinity, line 97:
```suggestion
        let shrink = canister_info.get_shrink().unwrap_or(true);
        if shrink {
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #2775, opened 2022-11-18
feat: use keyring for pem file storage

sesi200:
# Description

By default, PEM files are stored in the OS-provided keyring. If that is not available, they are encrypted with a password when writing them to disk.

Plaintext PEM files are still available (e.g. for use in non-interactive situations like CI), but not recommended for use since they put the keys at risk.

To force the use of one specific storage mode, use the `--storage-mode` flag with either `--storage-mode password-protected` or `--storage-mode plaintext`.

The flag `--disable-encryption` is deprecated in favour of `--storage-mode plaintext`.

Fixes [SDK-596](https://dfinity.atlassian.net/browse/SDK-596)

# How Has This Been Tested?

tested in e2e in various places

# Checklist:


src/dfx/src/lib/identity/keyring_proxy.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+struct KeyringProxy {
+    pub kv_store: HashMap<String, String>,
+}
+
+impl KeyringProxy {
+    fn get_location() -> DfxResult<PathBuf> {
+        Ok(PathBuf::from(std::env::var("HOME")?).join("mock_keyring.json"))
```
ericswanson-dfinity, line 47: A couple suggestions:
- Call all of this `KeyringMock`, `DFX_CI_USE_MOCK_KEYRING`, and so forth, since unless I'm misunderstanding the structure this piece is providing a mock for the keyring, not proxying to anything.
- Make the 'DFX_CI_USE_MOCK_KEYRING' environment variable be the path to this json file.  An empty string or even a value of `/dev/null` could be taken to mean `unavailable`.  This makes it so there are no magic filenames (`$HOME/mock_keyring.json`).

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 # UNRELEASED
 
+## DFX
+
+### feat: use OS-native keyring for pem file storage
+
+If keyring integration is available, PEM files are now by default stored in the OS-provided keyring.
+If it is not available, it will fall back on the already existing password-encrypted PEM files.
```
ericswanson-dfinity, line 10: It looks like the `default` identity is still stored unencrypted at `~/.config/dfx/identity/default/identity.pem`.  Is that intentional, and if so, should we mention that here?  It doesn't match the behavior described in this changelog entry.

src/dfx/src/lib/identity/keyring_proxy.rs
sesi200: Right. In my head, it proxied the call to the OS keyring, but calling it a mock makes more sense.

`DFX_CI_MOCK_KEYRING_LOCATION` now points to the file, and if it's empty string the keyring is unavailable.

CHANGELOG.md
sesi200: Mentioning it explicitly now. Yes, this is intentional because we don't want to touch keyring in any default flows. Otherwise this will hang 99.9% of all CI uses of dfx because it will be stuck on waiting for approval.

ericswanson-dfinity (approved):

PR #2796, opened 2022-12-02
feat: dfx pull can resolve dependencies from on-chain metadata

lwshang:
# Description

On the consumer side, `dfx pull` can fetch `dfx:deps` metadata and resolve dependencies recursively.

SDK-792

# How Has This Been Tested?

Add e2e tests cover happy/sad paths.

# Checklist:


e2e/tests-dfx/pull.bash
```diff
@@ -34,4 +34,71 @@ teardown() {
...
+    cd src/onchain_b
+    ic-wasm empty.wasm -o main.wasm metadata "dfx:deps" -d "rrkah-fqaaa-aaaaa-aaaaq-cai;onchain_a" -v public
+    cd ../../ # go back to root of "onchain" project
+    dfx deploy
+
+    cd ../app
+    assert_command_fail dfx pull
+    assert_contains "Failed while fetch and parse \`dfx:deps\` metadata from canister ryjl3-tyaaa-aaaaa-aaaba-cai."
+    assert_contains "Failed to parse \`dfx:deps\` entry: rrkah-fqaaa-aaaaa-aaaaq-cai. Expected \`name:Principal\`."
+}
```
ericswanson-dfinity, line 104: It looks like the code handles this case, but the tests don't cover it:
- dfx project: depends on onchain_a, onchain_b
- onchain_a and onchain_b both depend on onchain_c
- dfx pull fetches dependencies for onchain_c exactly once

lwshang: It's not trivial to test it in current e2e test framework. Any idea?

ericswanson-dfinity: What about the e2e framework makes it difficult?  It looked like you already have in this test setup that sets metadata to indicate dependencies between canisters.

It is because log output goes to stderr?  After `assert_command_fail` "$stderr" should give you that.

lwshang: How can I assert if a line only shows once in the output?

ericswanson-dfinity: sort | uniq -c | grep ...

PR #2805, opened 2022-12-06
feat(frontend-canister): better control over authorized users of asset canister

sesi200:
# Description

The asset canister now has two new functions:
- Query function `list_authorized` displays a list of all principals that are currently authorized to change assets and the list of authorized principals.
- Update function `deauthorize` that removes a principal from the list of authorized principals.

In addition, the update function `authorize` has new behavior:
Now, controllers of the asset canister are always allowed to authorize new principals (including themselves).

Fixes [SDK-883](https://dfinity.atlassian.net/browse/SDK-883)

# How Has This Been Tested?

added e2e

# Checklist:


e2e/tests-dfx/assetscanister.bash
```diff
@@ -14,6 +14,34 @@ teardown() {
...
+  install_asset assetscanister
+  dfx_start
+  assert_command dfx deploy
+  [ "$USE_IC_REF" ] && dfx canister update-settings --add-controller "$(dfx canister id e2e_project_frontend)" --all # Dec. 2022: ic-ref has not received the change yet that a canister is always allowed to check it's status
+
+  # deployer is automatically authorized
+  assert_command dfx canister call e2e_project_frontend list_authorized '()'
+  assert_contains "$PRINCIPAL"
+
+  assert_command dfx canister call e2e_project_frontend deauthorize "(principal \"$PRINCIPAL\")"
```
ericswanson-dfinity, line 28: Let's have this test demonstrate that `deauthorize` can only be called by an authorized caller

src/canisters/frontend/ic-certified-assets/src/lib.rs
```diff
@@ -24,15 +29,55 @@ thread_local! {
 
 #[update]
 #[candid_method(update)]
-fn authorize(other: Principal) {
+async fn authorize(other: Principal) {
+    let caller = caller();
+    let is_authorized = STATE.with(|s| s.borrow().is_authorized(&caller));
+    if is_authorized {
+        STATE.with(|s| {
+            if let Err(msg) = s.borrow_mut().authorize(&caller, other) {
```
ericswanson-dfinity, line 37: Since we've already checked `is_authorized` by this point, let's call `authorize_unconditionally` here and delete the `State::authorize` method.
```suggestion
            if let Err(msg) = s.borrow_mut().authorize_unconditionally(&caller, other) {
```

src/canisters/frontend/ic-certified-assets/src/lib.rs
```diff
@@ -24,15 +29,55 @@ thread_local! {
...
+                    trap(
+                        "The caller is not authorized and not part of the canister's controllers.",
+                    );
+                }
+            }
+        }
+    }
+}
+
+#[update(guard = "is_authorized")]
```
ericswanson-dfinity, line 64: I think if `authorize` allows controllers that are not on the authorized list, then so should `deauthorize`.  Thinking about this sequence of events from the perspective of such a controller:
1. Controller calls 'authorize' for a given principal
2. Controller for whatever reason wants to deauthorize that same principal (this could be immediately, or theoretically after some time, perhaps due to a proposal passing)
3. Controller calls `deauthorize` for that same principal
4. Access denied. Controller would have to add themselves to the authorized list first.

e2e/tests-dfx/assetscanister.bash
```diff
@@ -14,6 +14,34 @@ teardown() {
...
+  assert_not_contains "$PRINCIPAL"
+
+  # while not authorized, dfx deploy fails
+  echo "new file content" > 'src/e2e_project_frontend/assets/new_file.txt'
+  assert_command_fail dfx deploy
+
+  # canister controllers may always authorize principals, even if they're not authorized themselves
+  assert_command dfx canister call e2e_project_frontend authorize "(principal \"$PRINCIPAL\")"
+  assert_command dfx canister call e2e_project_frontend list_authorized '()'
+  assert_contains "$PRINCIPAL"
```
ericswanson-dfinity, line 39: Suggestion: authorize a different identity here, then deploy using it.  The existing tests in this file don't adequately cover authorized/non-authorized functionality.

ericswanson-dfinity (changes requested):

sesi200:
@ericswanson-dfinity all great suggestions. Included all of it

ericswanson-dfinity (approved):

PR #2806, opened 2022-12-06
fix(ci): docker build frontend canister on linux/amd64 platform

smallstepman:
# Description

`docker build` on macOS M1 without `--platform linux/amd64` flag frontend canister build artifact will differ from the made by CI workflow (`name: Check frontend canister build`). 

This change won't have any effect on the current CI workflow, however, it will make building the frontend canister locally a bit longer on macOS M1.


smallstepman:
could you try again? I've swapped the order of stuff in Dockerfile

If that fails, can you try one more time, but please issue `cargo install ic-wasm --version 0.2.0 --force` before running `./scripts/upd....` and see if that helps

PR #2824, opened 2022-12-15
feat(frontend-canister): add `allow_raw_access` config option

smallstepman:
# Description

`allow_raw_access` 
  - when unset or set to `false` - will redirect the traffic from `.raw.ic0.app` to `.ic0.app`,
  - when set to `true`, will allow to access the asset in frontend canister on the mainnet via `.raw.ic0.app` domain

Below `.ic-assets.json` example shows how to restrict access for `raw` domain for all assets
  ```json
  [
    {
      "match": "*/**",
      "allow_raw_access": false
    }
  ]
  ```

Fixes https://dfinity.atlassian.net/browse/SDK-474

# How Has This Been Tested?

unit + manual tests

# Checklist:


src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -155,4 +188,20 @@ impl HttpResponse {
             streaming_strategy: None,
         }
     }
+
+    pub fn build_redirect(status_code: u16, location: String) -> HttpResponse {
+        if ![300, 301, 302, 303, 304, 307, 308].contains(&status_code) {
+            return HttpResponse::build_400(&format!(
+                "incorrect asset redirect configuration: response_code \"{}\" is not valid HTTP respone code",
```
sesi200, line 195:
```suggestion
                "incorrect asset redirect configuration: response_code \"{}\" is not valid HTTP response code",
```

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -80,6 +80,39 @@ impl HttpRequest {
...
+                if let Some(t) = self.url.split("canisterId=").nth(1) {
+                    let x = t.split_once('&');
+                    if let Some(c) = x {
+                        return c.0;
+                    }
+                }
+            }
+        }
+        "canister-id-is-unreachable"
+    }
```
sesi200, line 107: I think this can be replaced with https://docs.rs/ic-cdk/latest/ic_cdk/api/fn.id.html

smallstepman: good idea, however as discussed I don't know how to make it work. 
the following changes 
image
make it not possible to run `cargo test` due to compile linking error 
```
cargo test --benches --tests --all-features 
   Compiling ic-certified-assets v0.2.5 (/src/canisters/frontend/ic-certified-assets)
error: linking with `cc` failed: exit status: 1
  |
  = note: "cc" "-arch" "arm64" "/var/folders/qr/2sz_4fyx75z7smnl9k554p000000gn/T/rustcSNu69F/symbols.o" 
...  
= note: Undefined symbols for architecture arm64:
            "_canister_self_copy", referenced from:
                ic_cdk::api::id::h072c680cecf25c11 in libic_cdk-f3b53b5d1a916317.rlib(ic_cdk-f3b53b5d1a916317.ic_cdk.8608ab03-cgu.9.rcgu.o)
            "_canister_self_size", referenced from:
                ic_cdk::api::id::h072c680cecf25c11 in libic_cdk-f3b53b5d1a916317.rlib(ic_cdk-f3b53b5d1a916317.ic_cdk.8608ab03-cgu.9.rcgu.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
          
error: could not compile `ic-certified-assets` due to previous error
```

maybe @ericswanson-dfinity will have an idea?

smallstepman:
an outstanding question is whether we should do something special (besides changelog entry) to inform our users that updating the frontend canister wasm will have an effect where already uploaded assets will become unreachable from the `.raw` domain

this is because the body of this if statement
https://github.com/dfinity/sdk/blob/5cf73266d9e761b029095f6e69ea806598d6c53b/src/canisters/frontend/ic-certified-assets/src/state_machine.rs#L465 
will get executed if the `allow_raw_access` is set to `None`, which I think is what will happen during a stable state upgrade from the old to the new version.

this is also a bit inconvenient for the user because current tooling does not support changing asset properties for already deployed assets (at least not via `.ic-assets.json` file - its only possible to do it via manual candid call to `set_asset_properties` method)

another option would be to change the logic in if statement to not capture the `None` value as its currently doing

ericswanson-dfinity:
this example in the description is the equivalent of default behavior
```
[
  {
    "match": "*/**",
    "allow_raw_access": false
  }
]
```

I'd suggest showing an example configuration that does something like configure robots.txt to allow raw access, or some other configuration that in some way matches how a developer might likely actually set up a configuration.

ericswanson-dfinity (changes requested):
Let's start with a CHANGELOG entry

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -155,4 +188,20 @@ impl HttpResponse {
...
         }
     }
+
+    pub fn build_redirect(status_code: u16, location: String) -> HttpResponse {
+        if ![300, 301, 302, 303, 304, 307, 308].contains(&status_code) {
+            return HttpResponse::build_400(&format!(
+                "incorrect asset redirect configuration: response_code \"{}\" is not valid HTTP response code",
+                status_code
+            ));
+        }
```
ericswanson-dfinity, line 198: This check seems unnecessary, since the only value passed is 308.  For this PR, I suggest removing the parameter altogether and renaming this function `build_permanent_redirect`.

If there is a later PR that allows for configurable status codes, I would suggest to validate the values when reading the configuration, and _maybe_ in the canister when setting the configuration, rather than when processing requests.

src/canisters/frontend/ic-certified-assets/src/http.rs
ericswanson-dfinity: No, but @adamspofford-dfinity did.

> It was just outdated ic0. `cargo update -p ic0`
> 0.18.6 added back the shim for compiling on desktop

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -459,6 +462,9 @@ impl State {
 
         if let Some(certificate_header) = index_redirect_certificate {
             if let Some(asset) = self.assets.get(INDEX_FILE) {
+                if asset.allow_raw_access.map_or(true, |v| !v) && req.is_raw_domain() {
```
ericswanson-dfinity, line 465: For clarity.  With `map_or`, default of true (the inverse of the default of `allow_raw_access`), and negating the value in the closure, there's a lot going on here.
```suggestion
                if !asset.allow_raw_access() && req.is_raw_domain() {
```

along with
```
impl Asset {
    pub fn allow_raw_access(&self) -> bool {
        self.allow_raw_access.unwrap_or(false)
    }
}

```

CHANGELOG.md
```diff
@@ -4,6 +4,18 @@
 
 ## DFX
 
+### feat(frontend-canister)!: add `allow_raw_access` config option
+
+By default, the traffic to the frontend canister will now restrict the access from `canister-id.raw.ic0.app` domain, and will automatically redirect all requests to certified domain (`canister-id.ic0.app`), unless configured explicitly. Below is the example configuration to allow access to `robots.txt` file from "raw" domain:
```
ericswanson-dfinity, line 9:
```suggestion
By default, the frontend canister will now restrict the access of traffic to the `<canister-id>.raw.ic0.app` domain, and will automatically redirect all requests to the certified domain (`<canister-id>.ic0.app`), unless configured explicitly. Below is an example configuration to allow access to the `robots.txt` file from the "raw" domain:
```

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -80,6 +80,47 @@ impl HttpRequest {
...
+            "https://{canister_id}.ic0.app{path}",
+            canister_id = ic_cdk::api::id().to_text(),
+            path = self.url
+        );
+        #[cfg(test)]
+        let location = format!(
+            "https://{canister_id}.ic0.app{path}",
+            canister_id = self.get_canister_id(),
+            path = self.url
+        );
```
ericswanson-dfinity, line 96: Is this necessary after updating the ic0 crate?

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -80,6 +80,47 @@ impl HttpRequest {
...
+        HttpResponse::build_redirect(308, location)
```
ericswanson-dfinity, line 97: Can `308` be a const somewhere?

src/canisters/frontend/ic-certified-assets/src/http.rs
smallstepman: looks like its still necessary. attempting to run `cargo test` while relying only on `ic_cdk::api::id()` results in the following error
```
thread 'tests::allow_raw_access::redirects_from_raw_to_certified' panicked at 'canister_self_size should only be called inside canisters.', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/ic0-0.18.7/src/ic0.rs:118:9
```

ericswanson-dfinity: Okay, how about some DRY then

```suggestion
        #[cfg(not(test))]
        let canister_id = ic_cdk::api::id().to_text();
        #[cfg(test)]
        let canister_id = self.get_canister_id();

        let location = format!(
            "https://{canister_id}.ic0.app{path}",
            canister_id,
            path = self.url
        );
```

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -80,6 +82,45 @@ impl HttpRequest {
...
+            if host_header.contains(".localhost") || host_header.contains(".app") {
+                return host_header.split('.').next().unwrap();
+            } else if let Some(t) = self.url.split("canisterId=").nth(1) {
+                let x = t.split_once('&');
+                if let Some(c) = x {
+                    return c.0;
+                }
+            }
+        }
+        "canister-id-is-unreachable"
```
ericswanson-dfinity, line 114: Since this is just for tests and not expected to be reached, maybe this should panic

ericswanson-dfinity (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,32 @@
...
+At the present time, there are two ways to reconfigure an existing asset:
+1. re-create the asset
+  1. delete the asset in your project's directory 
+  1. execute `dfx deploy`
+  1. re-create the asset in your project's directory
+  1. modify `.ic-assets.json` acordingly 
+  1. execute `dfx deploy`
+2. via manual candid call 
+  ```
+  dfx canister call PROJECTE_NAME_frontend set_asset_properties '( record { key="/robots.txt"; allow_raw_access=opt(opt(true)) })'
```
ericswanson-dfinity, line 29:
```suggestion
  dfx canister call PROJECT_NAME_frontend set_asset_properties '( record { key="/robots.txt"; allow_raw_access=opt(opt(true)) })'
```

smallstepman:
ready for re-review @ericswanson-dfinity

ericswanson-dfinity (approved):

PR #2832, opened 2022-12-20
docs: detail for canister update-settings --set-controller

ericswanson-dfinity:
# Description

Add some detail related to `dfx canister update-settings --set-controller` and multiple controllers.

Fixes https://dfinity.atlassian.net/browse/SDK-823


dfx-json (approved):
👍

PR #2838, opened 2022-12-22
feat: add telemetry canister

ericswanson-dfinity:
# Description

Adds the telemetry canister (backend) with methods for storing and reporting command usage.

See [SDK-841](https://dfinity.atlassian.net/browse/SDK-841) and [design doc](https://docs.google.com/document/d/1i__s7RHSgZY5Ix2Hl-GL8kpau9trhu6vC6dudP4c67A/edit?usp=sharing).

Note for reviewers: Some aspects touched on in this PR remain as future work:
- [SDK-864 access control](https://dfinity.atlassian.net/browse/SDK-864)
- [SDK-865 refine hash methods](https://dfinity.atlassian.net/browse/SDK-865)

# How Has This Been Tested?

Added e2e tests

[SDK-841]: https://dfinity.atlassian.net/browse/SDK-841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/canisters/telemetry/src/telemetry_backend/main.mo
```diff
@@ -0,0 +1,208 @@
...
+
+  let nsPerDay = 86_400 * 1000_000_000;
+  let nsPer30Days = 30 * nsPerDay;
+
+  stable var v0 : ?V0 = ?Data.v0();
+  let data : V0 = switch (v0) {
+    case null Prelude.unreachable();
+    case (?v0) v0;
+  };
+
```
ericswanson-dfinity, line 39: Reviewers: See https://github.com/dfinity/sdk/commit/eba34882a276dc48225b3156be20eb3c3aeeef30#diff-c95f31a8341300c4f8213b1fb1d534d40d4dfee78d7bf9d4ff2ef6796bd3d699R46 for for upgrades will work

src/canisters/telemetry/src/telemetry_backend/main.mo
```diff
@@ -0,0 +1,208 @@
...
+        let periodsToAdd =
+          (now - data.thirtyDayAggregationPeriodStart) / nsPer30Days;
+        data.thirtyDayAggregationPeriodStart += periodsToAdd * nsPer30Days;
+      };
+      data.thirtyDayAggregationPeriodEnd :=
+        data.thirtyDayAggregationPeriodStart + nsPer30Days - 1;
+    };
+  };
+
+  public func reportCommandResult(args: ReportCommandArgs) : async () {
```
chenyan-dfinity, line 71: I understand ACL is a separate ticket, but how do we plan to prevent non-dfx client sending data to this endpoint?

src/canisters/telemetry/src/telemetry_backend/Command.mo
```diff
@@ -0,0 +1,227 @@
...
+    #dfxWalletName;
+    #dfxWalletRedeemFaucetCoupon;
+    #dfxWalletRemoveController;
+    #dfxWalletSend;
+    #dfxWalletSetName;
+    #dfxWalletUpgrade;
+  };
+
+  public func encodeForHash(c : Command) : Nat32 {
+    switch (c) {
```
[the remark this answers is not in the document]
ericswanson-dfinity: Yeah, an earlier version had precomputed values of, for example, `Text.hash("#dfxCanisterInstall")`.  But see `CommandDetails.hash()` - which computes a hash of the combination of `Command` and `?Parameters`, so it's convenient if this fits in a Nat8.  And as you noted, the constants are also useful for upgrades.

src/canisters/telemetry/src/telemetry_backend/CommandDetails.mo
```diff
@@ -0,0 +1,40 @@
...
+  public func hash(v : CommandDetails) : Hash {
+    let key : [Hash] = [
+      Command.encodeForHash(v.command),
+      Option.getMapped(v.parameters, Parameters.encodeForHash, 0 : Hash)
+    ];
+    Hash.hashNat8(key)
+  };
+
+  public func equal(a : CommandDetails, b : CommandDetails) : Bool {
+    a.command == b.command and a.parameters == b.parameters
```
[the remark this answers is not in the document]
ericswanson-dfinity: Thank you, this is exactly the sort of thing I was hoping to learn about

src/canisters/telemetry/src/telemetry_backend/main.mo
ericswanson-dfinity: We aren't going to do anything to prevent it.  It's a public update method on the canister.  For the first iteration, dfx is probably going to actually spawn  `dfx canister call --network ic --identity anonymous ...`, since it's reporting success/failure of the command, in order not to slow down developer flow while waiting for the call to go through.

The only real measure taken in this regard is to store enums rather than strings (`#dfxCanisterCall` vs `"dfx canister call"`).  The only Text value stored is the dfx version, and we can at least verify that it's some subset of semver.

chenyan-dfinity: Sounds like you want to make this a one-way function, i.e. change the return type from `async ()` to `()`. This way canister call will return immediately and don't care about whether the update call succeed or not.

My concern with ACL is that one malicious user, e.g., calling this function all the time, can skew the whole aggregate result. But I don't have a good solution either.

ericswanson-dfinity: Thanks!  I didn't know about that distinction.

The references to ACL in the PR description are about controlling access to the methods used for testing (testSetTime and testGetCommandResults).  As for the rest, I agree about the potential problem and the lack of a good solution.

e2e/tests-telemetry/canister.bash
```diff
@@ -0,0 +1,188 @@
...
+    assert_command getInvocationDetailsForReportingPeriod 60
+    assert_eq '(
+  vec {
+    record {
+      dfxVersion = "0.12.1";
+      network = variant { localProject };
+      platform = variant { linux };
+    };
+  },
+)'
```
sesi200, line 187: I don't see anything that tests where `getInvocationDetailsForReportingPeriod` produces a vec with length >1, but I don't see why it wouldn't work.

sesi200 (commented):
Happy to approve if you think it's reviewed enough

e2e/tests-telemetry/canister.bash
ericswanson-dfinity: Oh thanks, I meant to include that

PR #2852, opened 2023-01-06
feat: add dfx autocomplete command

ericswanson-dfinity:
# Description

Adds `dfx beta generate-autocompletion-script`. By default generates a bash script, but can generate scripts for zsh, fish, elvish, and powershell with the `--shell` parameter.

Of the features listed in SDK-599, it does the following, replacing "most similar to" with "that starts with":
> As a user begins typing a command a dfx command and then presses the tab key, the command most similar to the phrase that is being typed should auto-complete. This should work for flags as well if possible.

It does not do this, and I do not foresee a path that would provide for it:
>  If there is no phrase that is currently being typed, dfx should autocomplete with the most recently used command.

See https://dfinity.atlassian.net/browse/SDK-599

# How Has This Been Tested?

I've tested this with zsh on OSX, using the bash script output, and it seems to work:

```
    $ dfx beta generate-autocompletion-script >x.sh
    $ source x.sh
    $ dfx can<tab>
```
I tried the `--shell zsh` output and went through some steps to configure zsh to use that, without success.

# Checklist:

### DIFF
--- a/src/dfx/src/commands/beta/generate_autocompletion_script.rs
+++ b/src/dfx/src/commands/beta/generate_autocompletion_script.rs
@@ -1,0 +1,44 @@
+use crate::lib::error::DfxResult;
+use crate::CliOpts;
+use std::fs::File;
+use std::io::Write;
+use std::path::PathBuf;
+
+use anyhow::Context;
+use clap::CommandFactory;
+use clap::Parser;
+use clap_complete::generate;
+use clap_complete::Shell;
+
+/// Generate a shell autocompletion script.
+#[derive(Parser)]
+pub struct AutocompleteOpts {
+    /// The shell for which to generate autocomplete scripts
+    #[clap(long, default_value("bash"))]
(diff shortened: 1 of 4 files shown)


src/dfx/src/commands/beta/generate_autocompletion_script.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub fn exec(opts: AutocompleteOpts) -> DfxResult {
+    let mut output: Box<dyn Write> =
+        if opts.output_file == PathBuf::from("-") {
+            Box::new(std::io::stdout())
+        } else {
+            Box::new(File::create(&opts.output_file).with_context(|| {
+                format!("Unable to open {} for output", opts.output_file.display())
+            })?)
+        };
```
ericswanson-dfinity, line 36: I don't think it's worth doing this, rather than letting the shell take care of it, because the behavior is the same anyway.

PR #2896, opened 2023-01-19
fix: Use an anonymous agent for `dfx generate`

### LINKED ISSUE #2872: `dfx deploy` halts when running `npm run generate` using a protected identity
I tried the following:

`dfx deploy` with password protected identity

I expected to see this happen: *explanation*

Successful deployment 

Instead, this happened: *explanation*

Halts with `Building frontend...`

Reason:

You need to enter the passphrase when running `npm run generate` .

adamspofford-dfinity:
This changes `dfx generate` to not require the current identity, as it does not use it but does use the rest of the AgentEnvironment machinery. 

Fixes #2872, SDK-916

### DIFF
--- a/e2e/tests-dfx/generate.bash
+++ b/e2e/tests-dfx/generate.bash
@@ -113,3 +113,13 @@ teardown() {
     assert_file_not_exists "src/declarations/hello_backend/index.js"
     assert_file_not_exists "src/declarations/hello_backend/index.d.ts"
 }
+
+@test "dfx generate succeeds with an encrypted identity without input" {
+    dfx_start
+    dfx canister create --all
+
+    assert_command "${BATS_TEST_DIRNAME}/../assets/expect_scripts/init_alice_with_pw.exp"
+    assert_command dfx identity use alice
+    
+    assert_command timeout 30s dfx generate
+}
--- a/src/dfx/src/commands/generate.rs
+++ b/src/dfx/src/commands/generate.rs
@@ -3,7 +3,7 @@ use crate::lib::environment::Environment;
 use crate::lib::error::DfxResult;
 use crate::lib::models::canister::CanisterPool;
 use crate::lib::models::canister_id_store::CanisterIdStore;
-use crate::lib::provider::create_agent_environment;
+use crate::lib::provider::create_anonymous_agent_environment;
 use crate::NetworkOpt;
 
 use clap::Parser;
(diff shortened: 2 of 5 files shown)


e2e/tests-dfx/generate.bash
```diff
@@ -113,3 +113,13 @@ teardown() {
...
 }
+
+@test "dfx generate succeeds with an encrypted identity without input" {
+    dfx_start
+    dfx canister create --all
+
+    assert_command "${BATS_TEST_DIRNAME}/../assets/expect_scripts/init_alice_with_pw.exp"
+    assert_command dfx identity use alice
+    
+    assert_command timeout 30s dfx generate
```
ericswanson-dfinity, line 124: If you're not checking the output of the commands (assert_eq, assert_match and so forth), but only checking for success/failure, there is no need for `assert_command`:
```suggestion
    "${BATS_TEST_DIRNAME}/../assets/expect_scripts/init_alice_with_pw.exp"
    dfx identity use alice
    
    timeout 30s dfx generate
```

src/dfx/src/commands/generate.rs
```diff
@@ -22,7 +22,7 @@ pub struct GenerateOpts {
 }
 
 pub fn exec(env: &dyn Environment, opts: GenerateOpts) -> DfxResult {
-    let env = create_agent_environment(env, opts.network.network)?;
+    let env = create_anonymous_agent_environment(env, opts.network.network)?;
```
ericswanson-dfinity, line 25: Could this have been achieved like so?
```suggestion
    let env = create_agent_environment(env.with_identity_override(ANONYMOUS_IDENTITY_NAME), opts.network.network)?;
```

e2e/tests-dfx/generate.bash
adamspofford-dfinity: I put assert_command in front of any command that the test could be considered to be testing, because I've run into several issues where I couldn't tell what command was failing because bats was eating the output. assert_command's output is much better.

src/dfx/src/lib/environment.rs
```diff
@@ -243,24 +244,41 @@ pub struct AgentEnvironment<'a> {
...
 impl<'a> AgentEnvironment<'a> {
-    #[context("Failed to create AgentEnvironment for network '{}'.", network_descriptor.name)]
     pub fn new(
         backend: &'a dyn Environment,
         network_descriptor: NetworkDescriptor,
         timeout: Duration,
     ) -> DfxResult<Self> {
-        let logger = backend.get_logger().clone();
         let mut identity_manager = IdentityManager::new(backend)?;
         let identity = identity_manager.instantiate_selected_identity(backend.get_logger())?;
```
ericswanson-dfinity, line 257: Suggestion: Reduce special-casing for the anonymous identity by adding a parameter `use_identity: Option<String>` here, and then either calling `identity_manager.instantiate_selected_identity` or `identity_manager.instantiate_identity_from_name`.

ericswanson-dfinity (approved):

PR #2906, opened 2023-01-21
feat: Write environment variables to .env file

adamspofford-dfinity:
dfx already passes several environment variables to build processes on `dfx build` or `dfx deploy`. This PR outputs these variables to a `.env` file (by default), editing any existing .env file without overwriting it. Variables containing the 'current' canister's info such as `CANISTER_ID` will be correct for the most recent build process.

Implements SDK-888.


e2e/tests-dfx/dotenv.bash
```diff
@@ -0,0 +1,75 @@
...
+    assert_contains "DFX_NETWORK='local'" "$env"
+    assert_contains "foo=bar" "$env"
+    
+    echo 'baz=quux' >> .env
+    assert_command dfx build
+    env=$(< .env)
+    assert_contains "DFX_NETWORK='local'" "$env"
+    assert_contains "foo=bar" "$env"
+    assert_contains "baz=quux" "$env"
+
```
ericswanson-dfinity, line 65: Can you make a test that shows what would happen if some information in the dfx canister environment variables section changed (say, by passing `--network <something else>`, or adding a canister)?  The test would demonstrate that those values changed, but other vars like `foo=bar` and `baz=quuz` here would not.

src/dfx/src/config/dfinity.rs
```diff
@@ -575,6 +575,9 @@ pub struct ConfigInterface {
     /// Mapping between network names and their configurations.
     /// Networks 'ic' and 'local' are implicitly defined.
     pub networks: Option<BTreeMap<String, ConfigNetwork>>,
+
+    /// If set, environment variables will be output to this file (without overwriting it).
```
ericswanson-dfinity, line 579: > (without overwriting it)
Would it be more accurate to say that dfx does overwrite the file (changing the section between `# DFX CANISTER ENVIRONMENT VARIABLES` and `# END DFX CANISTER ENVIRONMENT VARIABLES`) without overwriting other values?

src/dfx/src/lib/builders/assets.rs
```diff
@@ -126,13 +126,19 @@ impl CanisterBuilder for AssetsBuilder {
...
             .collect::<DfxResult<Vec<CanisterId>>>().with_context( || format!("Failed to collect dependencies (canister ids) of canister {}.", info.get_name()))?;
 
-        let vars = super::environment_variables(info, &config.network_name, pool, &dependencies);
+        let vars = super::get_and_write_environment_variables(
+            info,
+            &config.network_name,
+            pool,
+            &dependencies,
+            config.env_file.as_deref(),
+        );
```
ericswanson-dfinity, line 135: Just a little surprised the error is handled at the subsequent call site, rather than here
```suggestion
        )?;
```

ericswanson-dfinity (changes requested):

src/dfx/src/config/dfinity.rs
ericswanson-dfinity: I understand the distinction that you are bringing up.  However, given the previous clause, "will be output to this file," the word "it" will be understood to refer to "this file."  That leaves us with "... environment variables will be output to this file (without overwriting this file)"

The impression I got while reviewing this was that the documentation writer hadn't proofread his/her own documentation.  It's a small thing, but at best it seemed sloppy to me.  Please find an alternate wording.

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #2929, opened 2023-01-27
feat(frontend-canister): add take_ownership method

ericswanson-dfinity:
# Description

The SNS should be able to take ownership of the the Asset Canister’s by, with a single call, removing every principal from the list of authorized principals and adding itself as the only authorized principal.

Fixes https://dfinity.atlassian.net/browse/SDK-928

# How Has This Been Tested?

Added an e2e test

# Checklist:

- [ ] (n/a) I have made corresponding changes to the documentation.


krpeacock:
Just a naming suggestion - what about `clear_authorized`, since that's the language you reached for when describing it?

dfx-json (approved):

ericswanson-dfinity:
> Just a naming suggestion - what about `clear_authorized`, since that's the language you reached for when describing it?

Because it doesn't leave the authorized list empty.  The method adds the caller, a controller, to the list.  I updated the description in the PR.

PR #2931, opened 2023-01-30
refactor: always fetch did file from canister when making canister calls

chenyan-dfinity:
# Description

This refactor enables `dfx canister call` to always fetch did file from the canister metadata. This is especially helpful for remote canister.

Note that `dfx canister sign` is still following the old logic, because `sign` is running in offline mode and cannot access the network.

This is also a pre-requisite of the candid assist feature.

# How Has This Been Tested?

Existing tests should cover this refactor.

# Checklist:


src/dfx/src/commands/canister/call.rs
```diff
@@ -209,29 +212,34 @@ pub async fn exec(
...
             let canister_id = canister_id_store.get(callee_canister)?;
-            get_local_cid_and_candid_path(env, callee_canister, Some(canister_id))?
+            get_local_cid_and_candid_path(env, callee_canister, Some(canister_id))?.0
         }
     };
-    let maybe_candid_path = opts.candid.or(maybe_candid_path);
+    let method_type = if let Some(path) = opts.candid {
+        get_candid_type(CandidSource::Path(&path), method_name)
+    } else {
+        read_module_metadata(agent, canister_id, "candid:service")
```
ericswanson-dfinity, line 234: The idea is we will never use the local candid file?  What if the canister does not have candid:service metadata, or if it's not public?  I would have expected this to fall back to reading the candid:service metadata if the local candid file was not found, not to skip it entirely.

chenyan-dfinity: Yes, the assumption is that `dfx build` or other build tool have already put the local did file into the metadata. Unless you are the author of the canister, you won't have the local did file anyway, so reading the local did file is redundant. 

If the metadata is not there, dfx emits a warning, and hopefully the developer can fix that. If we read the local did file first, the canister owner will have less incentive to add metadata. So this is also a gentle push into a better ecosystem.

e2e/tests-dfx/assetscanister.bash
```diff
@@ -510,24 +510,24 @@ check_permission_failure() {
     dfx canister install e2e_project_frontend
 
     assert_command dfx canister call --query e2e_project_frontend retrieve '("/binary/noise.txt")' --output idl
-    assert_eq '(blob "\b8\01 \80\0aw12 \00xy\0aKL\0b\0ajk")'
+    assert_contains '(blob "\b8\01 \80\0aw12 \00xy\0aKL\0b\0ajk")'
```
ericswanson-dfinity, line 513: Let's instead have these tests show that stdout has only the expected output:
```suggestion
    # shellcheck disable=SC2154
    assert_eq '(blob "\b8\01 \80\0aw12 \00xy\0aKL\0b\0ajk")' "$stdout"
```

ericswanson-dfinity (changes requested):

src/dfx/src/commands/canister/call.rs
THLO: I discussed this with @sesi200. The conclusion of our discussion is that reading the candid:service metadata is a good approach to ensure that the current version is used. For the sake of backward compatibility, the local version should not be ignored, though. We feel that the following flow would make sense:

If a path is provided on the command line, the path is used to read the candid information.
If no path is provided, `dfx` tries to get the candid information from the metadata. If it succeeds to retrieve this information, it is used. If the information could not be retrieved, `dfx` tries to use the local `dfx.json` to get the candid information.

Does that make sense to you?

chenyan-dfinity: @THLO This PR implements the workflow you mentioned except for the last item.

> If the information could not be retrieved, dfx tries to use the local dfx.json to get the candid information.

It's a tech debt from the beginning due to the lack of metadata in the canister. We should aim to remove this reliance, instead of maintaining the tech debt. This change also won't break one's workflow unless they didn't put the candid info as metadata. In which case, we would like to nudge them to do so, instead of silently fallback.

Conceptually, everything under `.dfx` is a build artifact and they shouldn't be used when calling canisters. This ensures that the canister owner and other users have the same experience. A build process can fail, and you may end up with an out-of-sync local did file. We don't want to rely on the build artifact to call canisters. Happy to schedule a meeting to discuss this further.

THLO: I see your point. @sesi200 convinced me that maintaining backward compatibility would be nice but maybe it's not appropriate here. I think a quick call would be helpful.

e2e/tests-dfx/call.bash
```diff
@@ -26,9 +26,9 @@ teardown() {
   CANISTER_ID=$(dfx canister id hello_backend)
   rm .dfx/local/canister_ids.json
 
-  # if no candid file known, then no field names
-  assert_command dfx canister call "$CANISTER_ID" make_struct '("A", "B")'
-  assert_eq '(record { 99 = "A"; 100 = "B" })'
+    # given a canister id, fetch the did file from metadata
+    assert_command dfx canister call "$CANISTER_ID" make_struct '("A", "B")'
+    assert_eq '(record { c = "A"; d = "B" })'
```
ericswanson-dfinity, line 31:
```suggestion
  # given a canister id, fetch the did file from metadata
  assert_command dfx canister call "$CANISTER_ID" make_struct '("A", "B")'
  assert_eq '(record { c = "A"; d = "B" })'
```

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ Now they are stored with nanosecond precision on Windows too.
 
 Previously, this would fail with a PKCS#11: CKR_CRYPTOKI_ALREADY_INITIALIZED error.
 
+### fix!: always fetch did file from canister when making canister calls
```
ericswanson-dfinity, line 40: This will need to move to the unreleased section after https://github.com/dfinity/sdk/pull/3522 is merged

chenyan-dfinity: Done

ericswanson-dfinity (approved):

PR #2940, opened 2023-02-03
feat: dfx pull can download wasm

lwshang:
# Description

Close SDK-793

# How Has This Been Tested?

e2e

# Checklist:


e2e/tests-dfx/pull.bash
```diff
@@ -110,3 +112,74 @@ WARN: \`dfx:deps\` metadata not found in canister rrkah-fqaaa-aaaaa-aaaaq-cai."
...
+    cp src/onchain_b/main.wasm ../www/a.wasm 
+
+    cd ../app
+    assert_command dfx pull dep1
+    assert_contains "The hash of downloaded wasm doesn't match the canister on chain."
+
+    # sad path 2: url server doesn't have the file
+    rm ../www/a.wasm
+
+    assert_command dfx pull dep1
```
ericswanson-dfinity, line 183: should we expect a nonzero exit code from `dfx pull` here?
```suggestion
    assert_command_fail dfx pull dep1
```

src/dfx/src/commands/pull.rs
```diff
@@ -114,3 +130,87 @@ async fn fetch_deps_to_pull(
         },
     }
 }
+
+#[context("Failed while download wasm of canister {canister_id}.")]
```
ericswanson-dfinity, line 134: or `Failed while downloading`
```suggestion
#[context("Failed to download wasm of canister {canister_id}.")]
```

src/dfx/src/commands/pull.rs
```diff
@@ -114,3 +130,87 @@ async fn fetch_deps_to_pull(
...
+        }
+        None => {
+            warn!(
+                logger,
+                "Canister {canister_id} doesn't have module hash. Perhaps it's not installed."
+            );
+        }
+    }
+
+    wasm_file.write_all(&content)?;
```
ericswanson-dfinity, line 213: It's safer to download (and first write) to a temporary file, and then atomically rename into place.  Otherwise it's possible to have a file with only part of the contents written.

src/dfx/src/commands/pull.rs
```diff
@@ -114,3 +130,87 @@ async fn fetch_deps_to_pull(
...
+
+    let canister_status =
+        get_canister_status(agent_env, canister_id, &CallSender::SelectedId).await?;
+    match canister_status.module_hash {
+        Some(hash_on_chain) => {
+            let mut hasher = Sha256::new();
+            hasher.update(&content);
+            let hash_wasm = hasher.finalize();
+            if hash_wasm.as_slice() != hash_on_chain {
+                warn!(
```
ericswanson-dfinity, line 195: Do we have some design document that describes the tradeoffs between error/warning cases?  This seems like an error case to me.

ericswanson-dfinity (changes requested):

src/dfx/src/commands/pull.rs
```diff
@@ -95,22 +111,139 @@ async fn fetch_deps_to_pull(
...
+            );
+        }
+        Err(e) => {
+            bail!(e);
+        }
+    }
+    Ok(())
+}
+
+#[context("Failed to download wasm of canister rrkah-fqaaa-aaaaa-aaaaq-cai.")]
```
adamspofford-dfinity, line 128: Shouldn't this refer to the supplied principal?

src/dfx/src/commands/pull.rs
```diff
@@ -95,22 +111,139 @@ async fn fetch_deps_to_pull(
...
+
+    // skip download if cache hit
+    if wasm_path.exists() {
+        let bytes = std::fs::read(&wasm_path)?;
+        let mut hasher = Sha256::new();
+        hasher.update(&bytes);
+        let hash_cache = hasher.finalize();
+
+        if hash_cache.as_slice() == hash_on_chain {
+            info!(logger, "The canister wasm found in cache.");
```
adamspofford-dfinity, line 179:
```suggestion
            info!(logger, "The canister wasm was found in the cache.");
```

e2e/tests-dfx/pull.bash
```diff
@@ -95,18 +97,151 @@ WARN: \`dfx:deps\` metadata not found in canister rrkah-fqaaa-aaaaa-aaaaq-cai."
...
+    assert_file_exists "$WASM_CACHE/r7inp-6aaaa-aaaaa-aaabq-cai/canister.wasm"
+
+    # sad path 1: wasm hash doesn't match on chain
+    rm -r "${WASM_CACHE:?}/"
+    cd ../onchain
+    cp src/onchain_b/main.wasm ../www/a.wasm 
+
+    cd ../app
+    assert_command_fail dfx pull dep1
+    assert_contains "Failed to download wasm of canister rrkah-fqaaa-aaaaa-aaaaq-cai."
```
adamspofford-dfinity, line 180: Can we check for the specific error message?

adamspofford-dfinity (approved):

e2e/tests-dfx/pull.bash
```diff
@@ -68,13 +70,13 @@ teardown() {
 
     ## 1.2. pull onchain canisters in "app" project
     cd ../app
-    assert_command dfx pull dep1
-    assert_match "Pulling canister ryjl3-tyaaa-aaaaa-aaaba-cai...
+    assert_command_fail dfx pull dep1
```
ericswanson-dfinity, line 73: Since this is now `assert_command_fail`, will there now be `ERROR:` log output that this test can check for?

Looking at this test now, not knowing a whole lot about what's going on, it's not clear to me why it would be expected to fail.

e2e/tests-dfx/pull.bash
```diff
@@ -68,13 +70,13 @@ teardown() {
...
+    assert_contains "Pulling canister ryjl3-tyaaa-aaaaa-aaaba-cai...
 Pulling canister rrkah-fqaaa-aaaaa-aaaaq-cai...
 WARN: \`dfx:deps\` metadata not found in canister rrkah-fqaaa-aaaaa-aaaaq-cai."
 
-    assert_command dfx pull # if not specify canister name, all pull type canisters (dep1, dep2) will be pulled
-    assert_match "Pulling canister ryjl3-tyaaa-aaaaa-aaaba-cai...
+    assert_command_fail dfx pull # if not specify canister name, all pull type canisters (dep1, dep2) will be pulled
```
ericswanson-dfinity, line 78: Same question here about having the test show the reason for the expected failure

e2e/tests-dfx/pull.bash
```diff
@@ -95,18 +97,153 @@ WARN: \`dfx:deps\` metadata not found in canister rrkah-fqaaa-aaaaa-aaaaq-cai."
...
+
+    # prepare "onchain" canisters
+    cd onchain
+
+    echo -n -e \\x00asm\\x01\\x00\\x00\\x00 > src/onchain_a/main.wasm
+    ic-wasm src/onchain_a/main.wasm -o src/onchain_a/main.wasm metadata "dfx:wasm_url" -d "http://localhost:$E2E_WEB_SERVER_PORT/a.wasm" -v public
+
+    echo -n -e \\x00asm\\x01\\x00\\x00\\x00 > src/onchain_b/main.wasm
+    ic-wasm src/onchain_b/main.wasm -o src/onchain_b/main.wasm metadata "dfx:wasm_url" -d "http://localhost:$E2E_WEB_SERVER_PORT/b.wasm" -v public
+    ic-wasm src/onchain_b/main.wasm -o src/onchain_b/main.wasm metadata "dfx:deps" -d "onchain_a:rrkah-fqaaa-aaaaa-aaaaq-cai;" -v public
```
ericswanson-dfinity, line 145: Would it make sense to `dfx deploy` and `dfx canister create --all`, then use `dfx canister id` to obtain canister IDs, rather than hardcoding these canister ids?

src/dfx/src/commands/pull.rs
```diff
@@ -64,26 +71,38 @@ pub fn exec(env: &dyn Environment, opts: PullOpts) -> DfxResult {
             }
         }
 
-        Ok(())
+        let mut any_download_fail = false;
+
+        for canister_id in pulled_canisters {
+            if let Err(e) = download_canister_wasm(&agent_env, logger, canister_id).await {
+                warn!(
```
ericswanson-dfinity, line 78: Should this be an `error!` now?

src/dfx/src/commands/pull.rs
```diff
@@ -95,22 +114,138 @@ async fn fetch_deps_to_pull(
...
+            "Hash mismatch.
+  on chain: {}
+  download: {}",
+            hex::encode(hash_on_chain),
+            hex::encode(hash_download.as_slice())
+        );
+    }
+
+    // write to a tempfile
+    let mut f = tempfile::NamedTempFile::new()?;
```
ericswanson-dfinity, line 216: This should use `new_in`, to create the temp file on the same filesystem where `wasm_path` will go, so that the `rename` below will take place with an atomic move rather than an interruptible copy.

e2e/tests-dfx/pull.bash
lwshang: This line now fails because `dfx pull` will fail if any following downloads fails. And in this test, we didn't write `dfx:wasm_url` metadata which definitely will fail.
I would love to add some comments here to clarify the circumstance.

e2e/tests-dfx/pull.bash
lwshang: I overlooked that it's possible to get canister id before deploy it. I will fix this in another PR later.

src/dfx/src/commands/pull.rs
lwshang: Yes. Thanks for pointing it out!

e2e/tests-dfx/pull.bash
ericswanson-dfinity: Put another way: if `dfx pull` is expected to fail here (which it is), dfx should print some information about the reason.  We should have this test check for that output that describes the failure reason.  Does that make sense?

PR #2948, opened 2023-02-06
fix(frontend-canister): include `.well-known` dir by defualt, unless explicitly ignored

smallstepman:
# Description

Since `.well-known` is a well-known hidden directory, we're adding a rule that asset canisters should upload it's content by default, even when not configured in `.ic-assets.json`. 

user report:
> People who haven’t created a .ic-assets.json before get tripped up with that

Fixes https://dfinity.atlassian.net/browse/SDK-1218

# How Has This Been Tested?

added unit tests for all three cases:
- is the content of `.well-known` dir included by default even when not configured?
- is the content of `.well-known` dir included by when configured? (current behavior)
- is the content of `.well-known` dir not included when the `.well-known` directory is configured to be ignored?
# Checklist:


ericswanson-dfinity (changes requested):
Before implementing a feature like this, we should first discuss whether or not to do so.

src/canisters/frontend/ic-asset/src/sync.rs
```diff
@@ -220,9 +220,17 @@ fn include_entry(entry: &walkdir::DirEntry, config: &AssetConfig) -> bool {
...
+            .file_name()
+            .to_str()
+            .map(|s| s.starts_with(".well-known"))
+            .unwrap_or(false);
+
+    match (starts_with_a_dot, config.ignore, well_known) {
+        (_, ignored, wk) if wk && !ignored.map_or(false, |v| v) => true,
+        (dot, None, _) => !dot,
+        (_dot, Some(ignored), _) => !ignored,
     }
```
ericswanson-dfinity, line 234: Given that the logic here is, in order:
1. If the `ignore` field is present, use that
2. If it's in a `.well-known` directory, include it
3. include files that do not begin with a dot

Three ideas to make this easier to understand:

1. Reorder the match parameters left-to-right, cases top-to-bottom, and streamline the expressions:
```suggestion
    match (config.ignore, well_known, starts_with_a_dot) {
        (Some(ignored), _, _) => !ignored,
        (None, true, _) => true,
        (None, _, starts_with_a_dot) => !starts_with_a_dot,
    }
```

2. In this case an if-else chain might be easiest to understand:
```suggestion
    if let Some(ignore) = config.ignore {
        !ignore
    } else if well_known {
        true
    } else {
        !starts_with_a_dot
    }
```

3. Finally, while this method returns true to `include_entry`, this piece might be easier to understand by inverting that logic here:
```suggestion
    let ignore = if let Some(ignore) = config.ignore {
        ignore
    } else if well_known {
        false
    } else {
        starts_with_a_dot
    };
    !ignore
```

src/canisters/frontend/ic-asset/src/sync.rs
```diff
@@ -856,4 +864,82 @@ mod test_gathering_asset_descriptors_with_tempdir {
...
+        let mut asset_descriptors = dbg!(gather_asset_descriptors(&[&assets_dir]).unwrap());
+
+        let mut expected_asset_descriptors = vec![
+            AssetDescriptor::default_from_path(&assets_dir, "file"),
+            AssetDescriptor::default_from_path(&assets_dir, ".well-known/file"),
+        ];
+
+        expected_asset_descriptors.sort_by_key(|v| v.source.clone());
+        asset_descriptors.sort_by_key(|v| v.source.clone());
+        assert_eq!(asset_descriptors, expected_asset_descriptors);
```
ericswanson-dfinity, line 887: Since this is tracking inclusion/exclusion of specific files, do you think it would be more clear here, rather than creating asset descriptors and sorting them, to instead check that asset_descriptors contains (or does not contain) assets by key?

In other words, something like this
```
        let asset_keys = asset_descriptors.iter().map(|v| v.key.clone()).collect::<Vec<_>>();
        assert!(asset_keys.contains(".well-known/file".into()));
```

ericswanson-dfinity (changes requested):

sesi200:
Replaced by #3759

PR #2960, opened 2023-02-09
feat(frontend-canister): Add support for certification v2 

sesi200:
# Description

Implements the new asset certification spec that allows response headers to be certified.

Closes [SDK-927](https://dfinity.atlassian.net/browse/SDK-927), 

# How Has This Been Tested?

Added unit tests that check some basics, plus one test with an exact output that was tested e2e against the new service worker

# Checklist:

- [x] Security review approves of this PR
- [x] Asset canister is updated and changelog contains new canister hash
- [x] Verify the presence/absence of the certificate for v1 in unit tests as described in the changelog

[SDK-927]: https://dfinity.atlassian.net/browse/SDK-927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sesi200:
Notes to self:
- [ ]: add custom wasm section for supported versions
- [ ]: add empty node for request hash according to spec change

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -773,11 +931,14 @@ fn on_asset_change(
         }
     }
 
+    // Clean up pre-existing paths for this asset
+    let mut keys_to_remove = dependent_keys.clone();
+    keys_to_remove.push(key.to_string());
+    for key in keys_to_remove {
+        let key_path = AssetPath::from(&key);
+        asset_hashes.delete(key_path.asset_hash_path_root_v2().as_vec());
```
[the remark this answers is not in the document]
sesi200: They get added at the bottom in the iterator over `keys_to_insert_hash_for`. That list includes all dependent keys.

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -788,84 +949,96 @@ fn on_asset_change(
...
+    let keys_to_insert_hash_for: Vec<_> = keys_to_insert_hash_for
+        .into_iter()
+        .map(|key| {
+            let v1_hash_path = AssetPath::from(&key).asset_hash_path_v1();
+            (key, v1_hash_path)
+        })
+        .collect();
+
+    // Insert certified response values into hash_tree
+    for enc_name in encoding_order.iter() {
```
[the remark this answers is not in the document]
sesi200: implemented, thanks

src/canisters/frontend/ic-certified-assets/src/certification_types.rs
```diff
@@ -0,0 +1,137 @@
...
+/// E.g. `["http_expr", "foo", "index.html", "<$>", "<expr_hash>", "<response_hash>"]`
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct HashTreePath(pub Vec<NestedTreeKey>);
+
+impl<T> From<T> for AssetPath
+where
+    T: AsRef<str>,
+{
+    fn from(key: T) -> Self {
+        let mut iter = key.as_ref().split("/").peekable();
```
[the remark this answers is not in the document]
sesi200: `.split("/")` will do exactly that. See [this snippet](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=fn+main%28%29+%7B%0A++++let+a%3A+Vec%3C%26str%3E+%3D+%22%2Fa%2Fb%2F%22.split%28%22%2F%22%29.collect%28%29%3B%0A++++println%21%28%22%7Ba%3A%3F%7D%22%29%3B%0A%7D) as proof.

src/canisters/frontend/ic-certified-assets/src/certification_types.rs
```diff
@@ -0,0 +1,137 @@
...
+        Self(vec)
+    }
+}
+
+impl HashTreePath {
+    pub fn as_vec(&self) -> &Vec<NestedTreeKey> {
+        &self.0
+    }
+
+    pub fn expr_path(&self) -> String {
```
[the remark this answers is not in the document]
sesi200: We decided to alter the spec and allow it to start with `http_expr` because it makes building the tree significantly easier. I added it as a comment to the spec since we forgot to apply the change to the spec

nathanosdev:
Just an FYI, that the publishing the `response-verification` package is currently dependent on getting [this issue](https://dfinity.slack.com/archives/CGJND92DA/p1678988079143969) sorted.

src/canisters/frontend/ic-certified-assets/src/certification_types.rs
```diff
@@ -0,0 +1,138 @@
...
+pub struct HashTreePath(pub Vec<NestedTreeKey>);
+
+impl<T> From<T> for AssetPath
+where
+    T: AsRef<str>,
+{
+    fn from(key: T) -> Self {
+        let mut iter = key.as_ref().split("/").peekable();
+        if let Some(first_segment) = iter.peek() {
+            // "/path/to/asset".split("/") produces an emtpy node before "path", therefore we need to skip it
```
ericswanson-dfinity, line 33:
```suggestion
            // "/path/to/asset".split("/") produces an empty node before "path", therefore we need to skip it
```

src/canisters/frontend/ic-certified-assets/src/lib.rs
```diff
@@ -27,6 +29,9 @@ use ic_cdk_macros::{query, update};
 use serde_bytes::ByteBuf;
 use std::cell::RefCell;
 
+// #[link_section = "icp:public supported_certificate_versions"]
+// pub static SUPPORTED_CERTIFICATE_VERSIONS: [u8; 3] = *b"1,2";
```
ericswanson-dfinity, line 33: Is this unresolved?

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -830,117 +999,158 @@ impl From<StableState> for State {
...
+                    sha2::Sha256::digest(&[header_hash.as_ref(), enc.sha256.as_ref()].concat())
+                        .into();
+                response_hashes.insert(200, response_hash_200);
+
+                // add HTTP 304
+                encoding_headers
+                    .get_mut(0)
+                    .map(|elem| *elem = (":ic-cert-status".to_string(), Value::Number(304)));
+                let header_hash = representation_independent_hash(&encoding_headers);
+                let emtpy_body_hash: [u8; 32] = sha2::Sha256::digest(&[]).into();
```
ericswanson-dfinity, line 1115:
```suggestion
                let empty_body_hash: [u8; 32] = sha2::Sha256::digest(&[]).into();
```

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -830,117 +999,158 @@ impl From<StableState> for State {
...
+                max_age,
+                &*content_type,
+                enc_name,
+                enc.ic_ce.as_ref().map(|ce| &ce.ic_certificate_expression),
+            )
+            .into_iter()
+            .map(|(k, v)| (k, Value::String(v)))
+            .collect();
+
+            let new_response_hashes = {
```
ericswanson-dfinity, line 1099: Can we decompose this method into some smaller ones?  This block might be a starting point.

src/canisters/frontend/ic-certified-assets/src/tests.rs
```diff
@@ -232,6 +240,59 @@ fn can_create_assets_using_batch_api() {
...
+            .with_header("Accept-Encoding", "gzip")
+            .with_certificate_version(2)
+            .build(),
+        &[],
+        unused_callback(),
+    );
+    assert_eq!(gzip_response.status_code, 200);
+    assert_eq!(gzip_response.body.as_ref(), GZIP_BODY);
+
+    // V1 serves only the most important encoding - in this case "identity"
```
ericswanson-dfinity, line 283: AFAIK this is a change in behavior.  Given `Accept-Encoding: gzip`, and an asset with both identity and gzip encodings, the current asset canister returns the gzip encoding, but with the certification for the identity encoding 🤦‍♂️.  Is this change in behavior expected?

krpeacock:
CC @nomeata - relevant to the Motoko certification you're working on

src/canisters/frontend/ic-certified-assets/src/tests.rs
sesi200: Expected given the code changes, yes. I think it's ok to make a breaking change because this only affects users that 1) only accept `gzip` and nothing else, and 2) do not check for the certificate (over raw), and 3) use the HTTP interface, no other method, and 4) can't handle `identity` instead.

I can also offer these alternatives:
- Do not fix this bug in this PR and talk about that some more in e.g. an architecture meeting
- Revert to old behaviour if request is through `raw.ic0.app`

ericswanson-dfinity: Let's start with a description of the change in behavior in the changelog, with rationale.

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
sesi200: I was able to extract a lot. It is much easier to read now

src/canisters/frontend/ic-certified-assets/src/tests.rs
sesi200: I modified the behaviour a bit, and documented it properly in the changelog. This should not break any clients, unless they check for the presence of a cert but don't validate the contents properly

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -830,117 +1018,119 @@ impl From<StableState> for State {
...
-            if l != r {
-                panic!("merge_hash_trees: inconsistent leaves");
+        for status_code in STATUS_CODES_TO_CERTIFY {
+            if let Some(hash_path) = enc.asset_hash_path_v2(&key_path, status_code) {
+                asset_hashes.insert(hash_path.as_vec(), Vec::new());
+            } else {
+                unreachable!(
+                    "Could not create a hash path for a status code {} and key {} - did you forget to add a response hash for this status code?",
+                    status_code, &key
+                );
```
ericswanson-dfinity, line 1121: I guess this really is supposed to be unreachable.  A small suggestion to tie this to the likely culprit (compute_response_hashes).  When I read the verb "add" I thought maybe it had to be defined in the asset configuration somewhere.
```suggestion
                unreachable!(
                    "Could not create a hash path for a status code {} and key {} - did you forget to compute a response hash for this status code?",
                    status_code, &key
                );
```

src/canisters/frontend/ic-certified-assets/src/tests.rs
```diff
@@ -232,6 +240,67 @@ fn can_create_assets_using_batch_api() {
...
+
+    create_assets(
+        &mut state,
+        time_now,
+        vec![
+            AssetBuilder::new("/contents.html", "text/html")
+                .with_encoding("identity", vec![IDENTITY_BODY]),
+            AssetBuilder::new("/contents.html", "text/html").with_encoding("gzip", vec![GZIP_BODY]),
+        ],
+    );
```
ericswanson-dfinity, line 259: It would also be good to also show what happens if there is only an `identity` encoding (ic-asset only stores a gzip encoding if there is a size reduction), and only `gzip` is requested.  The identity encoding should be returned in this case.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

src/canisters/frontend/ic-certified-assets/src/tests.rs
```diff
@@ -1166,6 +1235,218 @@ mod allow_raw_access {
...
+#[cfg(test)]
+mod certificate_expression {
+    use crate::http::build_ic_certificate_expression_from_headers_and_encoding;
+
+    use super::*;
+
+    #[test]
+    fn ic_certificate_expression_value_from_headers() {
+        let h = ["a", "b", "c"].to_vec();
+        let c = build_ic_certificate_expression_from_headers_and_encoding(&h, "not identity");
```
ericswanson-dfinity, line 1247: would be good to have a version that shows what happens when encoding is `identity` (`content-encoding` disappears from the cert expr)

ericswanson-dfinity (changes requested):

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -48,6 +74,102 @@ pub struct AssetEncoding {
...
+                path.push(ce.expression_hash.as_slice().into());
+                path.push("".into()); // no request certification - use empty node
+                path.push(response_hash.as_slice().into());
+                Some(path.into())
+            } else {
+                None
+            }
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
sesi200: I like your suggestion better, thanks

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -173,6 +179,77 @@ impl HttpResponse {
...
+                        chunk_index,
+                        certificate_header,
+                        callback,
+                        etags,
+                        cert_version,
+                    );
+                }
+            }
+        }
+        unreachable!("No available encodings for requested asset.")
```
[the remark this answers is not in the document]
sesi200: You were right. Added a test case, and turned return type into an `Option<HttpResponse>`

src/canisters/frontend/ic-certified-assets/src/tests.rs
sesi200: test case added

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -15,6 +19,7 @@ pub struct HttpRequest {
     pub url: String,
     pub headers: Vec<HeaderField>,
     pub body: ByteBuf,
+    pub certificate_version: Option<Nat>,
```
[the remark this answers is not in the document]
sesi200: Spec and code have been changed to `u16`, it makes the service worker side better as well

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -173,6 +182,89 @@ impl HttpResponse {
...
 
+    #[allow(clippy::too_many_arguments)]
+    pub fn build_ok_from_requested_encodings(
+        asset: &Asset,
+        requested_encodings: &Vec<String>,
+        key: &str,
+        chunk_index: usize,
+        certificate_header: Option<&HeaderField>,
+        callback: &Func,
+        etags: &Vec<Hash>,
```
ericswanson-dfinity, line 193: Is it more common to pass this way?
```suggestion
        etags: &[Hash],
```

src/canisters/frontend/ic-certified-assets/src/http.rs
```diff
@@ -200,3 +292,58 @@ impl HttpResponse {
...
+    }
+
+    let ic_certificate_expression = IC_CERTIFICATE_EXPRESSION_VALUE.replace("{headers}", &headers);
+    let expression_hash = sha2::Sha256::digest(ic_certificate_expression.as_bytes())
+        .into_iter()
+        .collect();
+    CertificateExpression {
+        expression: ic_certificate_expression,
+        hash: expression_hash,
+    }
```
ericswanson-dfinity, line 316:
```suggestion
    let expression = IC_CERTIFICATE_EXPRESSION_VALUE.replace("{headers}", &headers);
    let hash = sha2::Sha256::digest(expression.as_bytes())
        .into_iter()
        .collect();
    CertificateExpression {
        expression,
        hash,
    }
```

src/canisters/frontend/ic-certified-assets/src/state_machine.rs
```diff
@@ -29,25 +35,152 @@ pub const BATCH_EXPIRY_NANOS: u64 = 300_000_000_000;
...
+                .and_then(|hashes| hashes.get(&200))
+                .and_then(|response_hash| {
+                    Some(HashTreePath::from(Vec::<NestedTreeKey>::from([
+                        "http_expr".into(),
+                        "<*>".into(), // 404 not found wildcard segment
+                        ce.hash.as_slice().into(),
+                        "".into(), // no request certification - use empty node
+                        response_hash.as_slice().into(),
+                    ])))
+                })
```
ericswanson-dfinity, line 119: Maybe we need to investigate why clippy isn't catching this, or maybe Marcin just fixed this in [#3051](https://github.com/dfinity/sdk/pull/3051), but isn't `.and_then(|_|Some(..))` equivalent to `.map()`?
```suggestion
                .map(|response_hash| {
                    HashTreePath::from(Vec::<NestedTreeKey>::from([
                        "http_expr".into(),
                        "<*>".into(), // 404 not found wildcard segment
                        ce.hash.as_slice().into(),
                        "".into(), // no request certification - use empty node
                        response_hash.as_slice().into(),
                    ]))
                })
```

ericswanson-dfinity (approved):
A couple very minor suggestions, up to you, and nothing that should prevent merging

PR #2972, opened 2023-02-16
feat: Allow specifying assets canister path in environment variable

adamspofford-dfinity:
This PR adds the DFX_ASSETS_WASM environment variable, which works like DFX_WALLET_WASM, but for the assets canister. In addition to being useful this makes SDK-672 testable.


smallstepman (approved):
LGTM!

PR #2975, opened 2023-02-16
fix: Fix schema for canister .build property

### LINKED ISSUE #2908: vscode gives false warning in json schema
Ik have the following prebuild canister configuration in vscode:

"Token": {
      "type": "custom",
      "wasm": "ledger/ic-icrc1-ledger.wasm",
      "candid": "ledger/icrc1.did"
}

Warning given in the json file (with the yellow squiggly line under "Token") 

Missing property "build". The canister seems to deploy fine and does not need the "build" field in the config

adamspofford-dfinity:
The `#[serde(default)]` nature of `CanisterTypeProperties::Custom.build` was preserved when `Deserialize` was manually implemented, but without an attribute, `derive(JsonSchema)` can't see it. This fixes the schema to make it look optional again.

Fixes #2908, SDK-932


sesi200:
Can you add a changelog entry? And I think the schema in the docs (`docs/dfx-json-schema.json`) should be updated too

sesi200 (approved):
Thank you!

PR #2934, opened 2023-01-30
feat: Add dfx sns download

mraszyk:
Adding a new command to download SNS canister WASMs. This is needed for SNS testflight without installing NNS, i.e., without running `dfx nns install`.

PR #2979, opened 2023-02-22
feat: Get identity from env var if present

bitdivine:
# Description
`dfx identity use` changes global state.  This means that one script that sets a default identity for its own purposes can interfere with another running concurrently.

This PR proposes a mechanism of scoping a default identity to the current shell.  The identity for a command is:
- as specified by `--identity XXX`, if present, else
- as specified by the environment variable `DFX_IDENTITY=XXX`, if present, else
- as specified by `dfx identity use`
The second of these is scoped to the local shell.

# How Has This Been Tested?
- I have run `DFX_IDENTITY=default cargo run -- identity whoami` and similar commands locally.
- I have included a CI test in this PR.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ By specifying the `--no-asset-upgrade` flag in `dfx deploy` or `dfx canister ins
 
 ### feat: Get identity from env var if present
 
-The identity may be specified using the environment variable `DFX_DENTITY`.
+The identity may be specified using the environment variable `DFX_IDENTITY`.
 
 ### feat: Add DFX_ASSETS_WASM
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 By specifying the `--no-asset-upgrade` flag in `dfx deploy` or `dfx canister install`, you can ensure that the asset canister itself is not upgraded, but instead only the assets themselves are installed.
 
+### feat: Get identity from env var if present
+
+The identity may be specified using the environment variable `DFX_DENTITY`.
+
 ### feat: Add DFX_ASSETS_WASM
 
 Added the ability to configure the WASM module used for assets canisters through the environment variable `DFX_ASSETS_WASM`.
--- a/e2e/tests-dfx/identity_command.bash
+++ b/e2e/tests-dfx/identity_command.bash
@@ -389,6 +389,19 @@ default'
     assert_eq '2vxsx-fae'
 }
 
+@test "identity use: is overridden by env var DFX_IDENTITY" {
+    assert_command dfx identity new dan
+    assert_command dfx identity new frank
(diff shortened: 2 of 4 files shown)


src/dfx/src/main.rs
```diff
@@ -180,7 +180,9 @@ fn print_error_and_diagnosis(err: Error, error_diagnosis: Diagnosis) {
 fn main() {
     let cli_opts = CliOpts::parse();
     let (verbose_level, log) = setup_logging(&cli_opts);
-    let identity = cli_opts.identity;
+    let identity = cli_opts
+        .identity
+        .or_else(|| std::env::var("DFX_IDENTITY").ok());
```
ericswanson-dfinity, line 185: clap has a feature for this, see https://github.com/dfinity/sdk/pull/2934/files#diff-525ab7c425ed736af9f5d76feaa51616197f59fee5db465772ee27c534a57eccR15

e2e/tests-dfx/identity_command.bash
```diff
@@ -389,6 +389,19 @@ default'
...
+    assert_command dfx identity new dan
+    assert_command dfx identity new frank
+    assert_command dfx identity use dan
+    assert_command dfx identity whoami
+    assert_eq 'dan'
+    DFX_IDENTITY=frank
+    export DFX_IDENTITY
+    assert_command dfx identity whoami
+    assert_eq 'frank'
+}
```
ericswanson-dfinity, line 402: Let's show that the `--identity` parameter overrides the env var as expected, something like this:
```suggestion
    assert_command dfx identity whoami --identity dan
    assert_eq 'dan'
}
```
though it might be nicer to add a third identity, since dan is also selected by config

ericswanson-dfinity (changes requested):

e2e/tests-dfx/identity_command.bash
bitdivine: Good idea.  Doing... Dan.

src/dfx/src/main.rs
bitdivine: Nice.  Does that mean that the env var is documented automatically?  Yes it does.  Cool.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 By specifying the `--no-asset-upgrade` flag in `dfx deploy` or `dfx canister install`, you can ensure that the asset canister itself is not upgraded, but instead only the assets themselves are installed.
 
+### feat: Get identity from env var if present
+
+The identity may be specified using the environment variable `DFX_DENTITY`.
```
ericswanson-dfinity, line 13: typo
```suggestion
The identity may be specified using the environment variable `DFX_IDENTITY`.
```

ericswanson-dfinity (approved):

PR #2999, opened 2023-03-03
style(frontend-canister): rename `container_asset{s}` to `canister_asset{s}`

smallstepman:
# Description

I found `container` to be a bit confusing (as if it would refer to `Containerization`). I believe `canister` will reflect intent better.


ericswanson-dfinity (approved):
Good eye, thanks!

PR #3002, opened 2023-03-04
feat(frontend-canister): update asset properties during `ic_asset::sync` (via batch)

smallstepman:
# Description

Allows to synchronize the properties of assets also after assets have already been deployed.
 
Closes https://dfinity.atlassian.net/browse/SDK-819
Closes https://dfinity.atlassian.net/browse/SDK-565
Closes https://github.com/dfinity/sdk/issues/3053

# How Has This Been Tested?

- [x] e2e test
- [x] unit tests in `ic-certified-assets`

# Checklist:


src/canisters/frontend/ic-asset/src/asset/config.rs
```diff
@@ -937,3 +939,18 @@ mod with_tempdir {
         Ok(())
     }
 }
+
+impl std::cmp::PartialEq<Option<&AssetProperties>> for AssetConfig {
```
ericswanson-dfinity, line 943: Is it really necessary to write this by hand?

src/canisters/frontend/ic-asset/src/asset/config.rs
```diff
@@ -937,3 +939,18 @@ mod with_tempdir {
...
+    fn eq(&self, other: &Option<&AssetProperties>) -> bool {
+        if let Some(other) = other {
+            self.cache
+                .clone()
+                .map_or(CacheConfig::default().max_age, |v| v.max_age)
+                == other.max_age
+                && self.allow_raw_access == other.allow_raw_access
+                && self.headers == other.headers
+        } else {
+            false
```
ericswanson-dfinity, line 953: Is `None PartialEq None` really `false`?

src/canisters/frontend/ic-asset/src/asset/config.rs
ericswanson-dfinity: I mean, does derive(PartialEq) on AssetProperties not work?

smallstepman: this impl compares AssetConfig vs AssetProperties (two different structs)

src/canisters/frontend/ic-asset/src/batch_upload/operations.rs
```diff
@@ -167,3 +175,54 @@ pub(crate) fn set_encodings(
...
+) {
+    for (key, project_asset) in project_assets {
+        let project_asset_properties = project_asset.asset_descriptor.config.clone();
+        let canister_asset_properties = canister_asset_properties.get(key);
+        // skip if the asset is not already in the canister, because
+        // proporties gonna be created during create_new_assets call
+        if canister_asset_properties.is_none() {
+            continue;
+        }
+        let canister_asset_properties = canister_asset_properties.unwrap();
```
ericswanson-dfinity, line 192: Isn't this all the same, but without the is_none/unwrap?
```suggestion
        // skip if the asset is not already in the canister, because
        // properties gonna be created during create_new_assets call
        let Some(canister_asset_properties) = canister_asset_properties.get(key) {
            ....
        }
```

src/canisters/frontend/ic-asset/src/batch_upload/operations.rs
```diff
@@ -167,3 +175,54 @@ pub(crate) fn set_encodings(
...
+                        hm.iter()
+                            .map(|(k, v)| (k.clone(), v.clone()))
+                            .collect::<Vec<_>>()
+                    })),
+                    is_aliased: Some(project_asset_properties.enable_aliasing),
+                    allow_raw_access: Some(project_asset_properties.allow_raw_access),
+                },
+            ));
+        }
+    }
```
ericswanson-dfinity, line 227: Let's set these individually.  They are all `opt opt` in order to support this.  Meaning all these values start as None, and then:
- set max_age to Some(new max age) if max age is different
- set headers to Some(new headers) if headers are different
- set is_aliased to Some(new value) if enable_aliasing is different
- set allow_raw_access to Some(new value) if allow_raw_access is different

This is a little more efficient, but more importantly, notice that the check on line 207 didn't check for `enable_aliasing` to change, so if it was the only thing that changed, no SetAssetProperties operation would be pushed.

Please start by updating a test, because this code should have failed at least four assertions.

src/canisters/frontend/ic-asset/src/canister_api/types/batch_upload/v1.rs
```diff
@@ -0,0 +1,41 @@
...
+        ClearArguments, CreateAssetArguments, DeleteAssetArguments, SetAssetContentArguments,
+        UnsetAssetContentArguments,
+    },
+};
+use candid::{CandidType, Nat};
+
+/// Batch operations that can be applied atomically.
+#[derive(CandidType, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub enum BatchOperationKind {
+    /// Create a new asset.
```
ericswanson-dfinity, line 13: Please see https://github.com/dfinity/sdk/pull/3085 and reorder these accordingly

src/canisters/frontend/ic-asset/src/evidence/mod.rs
```diff
@@ -37,11 +40,19 @@ pub async fn compute_evidence(
     let asset_descriptors = gather_asset_descriptors(dirs, logger)?;
 
     let canister_assets = list_assets(canister).await?;
+    let canister_asset_properties = get_assets_properties(canister, &canister_assets)
```
ericswanson-dfinity, line 43: Let's add log output before `list_assets` and before and after `get_assets_properties`.  I ran a test locally with 1000 assets and `get_assets_properties` took five seconds, which isn't excessive, but it's long enough that we should give some feedback.

src/canisters/frontend/ic-asset/src/evidence/mod.rs
```diff
@@ -37,11 +40,19 @@ pub async fn compute_evidence(
...
+        .await
+        .unwrap_or_default(); // older canisters don't have get_assets_properties method
```
ericswanson-dfinity, line 45: If there is an error retrieving asset properties, do we really want to set all properties for every asset?

src/canisters/frontend/ic-asset/src/sync.rs
```diff
@@ -32,6 +36,9 @@ pub async fn upload_content_and_assemble_sync_operations(
     let asset_descriptors = gather_asset_descriptors(dirs, logger)?;
 
     let canister_assets = list_assets(canister).await?;
+    let canister_asset_properties = get_assets_properties(canister, &canister_assets)
+        .await
+        .unwrap_or_default(); // older canisters don't have get_assets_properties method
```
ericswanson-dfinity, line 41: Same here about differentiating between an i/o error and method not existing.

src/canisters/frontend/ic-asset/src/sync.rs
```diff
@@ -32,6 +36,9 @@ pub async fn upload_content_and_assemble_sync_operations(
     let asset_descriptors = gather_asset_descriptors(dirs, logger)?;
 
     let canister_assets = list_assets(canister).await?;
+    let canister_asset_properties = get_assets_properties(canister, &canister_assets)
```
ericswanson-dfinity, line 39: Same here about logging what's going on as it happens

ericswanson-dfinity (changes requested):

src/canisters/frontend/ic-asset/src/canister_api/types/batch_upload/v1.rs
smallstepman: I just learned that when ord is derived for enum, variants are ordered by their top-to-bottom discriminant order.
rust is so awesome

src/canisters/frontend/ic-asset/src/evidence/mod.rs
smallstepman: thanks for benchmarking this!

e2e/tests-dfx/assetscanister.bash
```diff
@@ -131,6 +131,34 @@ check_permission_failure() {
   assert_command curl --fail -vv -H "Accept-Encoding: gzip" http://localhost:"$PORT"/notreally.js?canisterId="$ID"
   assert_not_contains "content-encoding"
   assert_eq "will not compress" "$stdout"
+
+  # show SetAssetProperties works as expected
+  echo '[
+    {
+      "match": "new_file.txt",
```
ericswanson-dfinity, line 138: What would you think about setting these properties on one of the files deployed on line 74?  This way we wouldn't need additional deploy/propose/commit steps

smallstepman: Ah, right, I haven't registered the first `dfx deploy` (the one that doesn't go through the proposal).

But just to make sure we're on the same page: the properties will have to be attached to `notreally.js` instead of `new_file.txt`, cause `new_file.txt` won't have any `BatchOperationKind::SetAssetProperties` cause its properties will be attached when calling `create_new_asset`. But this should work fine for `notreally.js` cause it's already in the canister

src/canisters/frontend/ic-asset/src/canister_api/methods/asset_properties.rs
```diff
@@ -0,0 +1,53 @@
...
+                reject_message,
+            }) if reject_code == 3
+                && reject_message
+                    .contains(&format!("has no query method '{GET_ASSET_PROPERTIES}'")) =>
+            {
+                break;
+            }
+            _ => {
+                return Err(anyhow::anyhow!("Failed to get asset properties"));
+            }
```
ericswanson-dfinity, line 35: Better to add context to an underlying error, rather than replacing it.

src/canisters/frontend/ic-asset/src/evidence/mod.rs
```diff
@@ -37,11 +40,21 @@ pub async fn compute_evidence(
     let asset_descriptors = gather_asset_descriptors(dirs, logger)?;
 
     let canister_assets = list_assets(canister).await?;
+    info!(
+        logger,
+        "Fetching properties for all assets in the canister."
+    );
+    let canister_asset_properties = get_assets_properties(canister, &canister_assets).await?;
```
ericswanson-dfinity, line 48: Please display another info! line here, because the next step takes some time too

src/canisters/frontend/ic-asset/src/sync.rs
```diff
@@ -72,18 +82,16 @@ pub async fn sync(canister: &Canister<'_>, dirs: &[&Path], logger: &Logger) -> a
...
+        0 => {
+            let commit_batch_args_v0 = v0::CommitBatchArguments::try_from(commit_batch_args)
+                .map_err(|e| anyhow!("Failed to downgrade from v1::CommitBatchArguments to v0::CommitBatchArguments: {}. Please upgrade your asset canister, or use older tooling (dfx<=v-0.13.1 or icx-asset<=0.20.0)", e))?;
+            warn!(logger, "The asset canister is running an old version of the API. It will not be able to set assets properties.");
+            commit_batch(canister, commit_batch_args_v0).await
         }
-    }
+        1.. => commit_batch(canister, commit_batch_args).await,
+    };
+    response.map_err(|e| anyhow!("The API version of the canister (v{}) is incompatible with batch upload version (v{}): {}", canister_api_version, BATCH_UPLOAD_API_VERSION, e))?;
```
ericswanson-dfinity, line 94: Is this the only possible explanation for `response` to be an Err?

src/canisters/frontend/ic-certified-assets/src/evidence.rs
```diff
@@ -22,6 +22,7 @@ const TAG_CREATE_ASSET: [u8; 1] = [4];
 const TAG_SET_ASSET_CONTENT: [u8; 1] = [5];
 const TAG_UNSET_ASSET_CONTENT: [u8; 1] = [6];
 const TAG_DELETE_ASSET: [u8; 1] = [7];
+const TAG_SET_ASSET_PROPERTIES: [u8; 1] = [9];
 const TAG_CLEAR: [u8; 1] = [8];
```
ericswanson-dfinity, line 26:
```suggestion
const TAG_CLEAR: [u8; 1] = [8];
const TAG_SET_ASSET_PROPERTIES: [u8; 1] = [9];
```

e2e/tests-dfx/assetscanister.bash
ericswanson-dfinity: That's right, there are a couple assets that will already be in the canister.  But even better, how about adding another asset specifically to be able to set its properties in the deploy?

src/canisters/frontend/ic-asset/src/batch_upload/operations.rs
```diff
@@ -183,6 +183,8 @@ pub(crate) fn update_properties(
 ) {
     for (key, project_asset) in project_assets {
         let project_asset_properties = project_asset.asset_descriptor.config.clone();
+        // skip if the asset is not already in the canister, because
+        // proporties gonna be created during create_new_assets call
```
ericswanson-dfinity, line 187:
```suggestion
        // properties gonna be created during create_new_assets call
```

src/canisters/frontend/ic-asset/src/sync.rs
```diff
@@ -91,7 +91,8 @@ pub async fn sync(canister: &Canister<'_>, dirs: &[&Path], logger: &Logger) -> a
         }
         1.. => commit_batch(canister, commit_batch_args).await,
     };
-    response.map_err(|e| anyhow!("The API version of the canister (v{}) is incompatible with batch upload version (v{}): {}", canister_api_version, BATCH_UPLOAD_API_VERSION, e))?;
+    response
+        .map_err(|e| anyhow!("Failed to synchronize frontend canister with project assets: {e}"))?;
```
ericswanson-dfinity, line 95:
```suggestion
    response.context("Failed to synchronize frontend canister with project assets")?;
```

src/canisters/frontend/ic-certified-assets/src/tests.rs
```diff
@@ -2874,6 +2874,117 @@ mod evidence_computation {
...
+        let time_now = 100_000_000_000;
+
+        let batch_1 = state.create_batch(time_now);
+        assert!(state
+            .propose_commit_batch(CommitBatchArguments {
+                batch_id: batch_1.clone(),
+                operations: vec![BatchOperation::SetAssetProperties(
+                    SetAssetPropertiesArguments {
+                        key: "/1".to_string(),
+                        max_age: Some(Some(100)),
```
ericswanson-dfinity, line 2944: What we're going for in these unit tests is that in each test, the batches differ by only one field.  So ideally, here we would add tests that show that changing any one field (key, max_age, headers, allow_raw_access, is_aliased) changes the evidence.  Since these are all `opt opt`, there would actually be a lot of variations.

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

ericswanson-dfinity (approved):

PR #3018, opened 2023-03-14
feat: ledger transfer optional set created_at_time field

levifeldman:
Useful when controlling transaction-deduplication. https://internetcomputer.org/docs/current/developer-docs/integrations/icrc-1/#transaction-deduplication-

# Description

Adds an optional `--created-at-time <timestamp-nanoseconds>`  into the ledger functions: `transfer`, `create-canister`, and `top-up`.

# Checklist:


src/dfx/src/commands/ledger/create_canister.rs
```diff
@@ -74,13 +78,21 @@ pub async fn exec(env: &dyn Environment, opts: CreateCanisterOpts) -> DfxResult
...
     })?;
 
+    let created_at_time = opts
+        .created_at_time
+        .as_ref()
+        .map_or(Ok(None), |v| {
+            u64::from_str_radix(v, 10).map(|n| Some(n)).map_err(|err| anyhow!(err))
+        })
+        .context("Failed to parse nanoseconds created_at_time.")?;
+
```
ericswanson-dfinity, line 88: If `Option<String>` is required, please DRY this (same code block exists in three places)

src/dfx/src/commands/ledger/create_canister.rs
```diff
@@ -47,6 +47,10 @@ pub struct CreateCanisterOpts {
     #[clap(long, validator(icpts_amount_validator))]
     max_fee: Option<String>,
 
+    /// Transaction timestamp, in nanoseconds, for use in controlling transaction-deduplication, default is system-time. // https://internetcomputer.org/docs/current/developer-docs/integrations/icrc-1/#transaction-deduplication-
+    #[clap(long)]
+    created_at_time: Option<String>,
```
ericswanson-dfinity, line 52: Would `Option<u64>` work here, and eliminate the need for the manual conversion?

src/dfx/src/commands/ledger/mod.rs
```diff
@@ -135,7 +136,7 @@ pub async fn transfer(
                     fee,
                     from_subaccount,
                     to,
-                    created_at_time: Some(TimeStamp { timestamp_nanos }),
+                    created_at_time: Some(TimeStamp { timestamp_nanos: created_at_time.unwrap_or(timestamp_nanos) }),
```
ericswanson-dfinity, line 139: Recommend setting this at the beginning of this method, `let timestamp_nanos = `, to make it clear that the SystemTime::now()... expression isn't used at all if created_at_time is Some.

ericswanson-dfinity (changes requested):
Please add appropriate e2e tests to https://github.com/dfinity/sdk/blob/master/e2e/tests-dfx/ledger.bash

src/dfx/src/commands/ledger/create_canister.rs
levifeldman: String is not required. Changed to Option<u64> and removed the conversions.

src/dfx/src/commands/ledger/create_canister.rs
levifeldman: Yes Option<u64> works. Changed in this commit https://github.com/dfinity/sdk/pull/3018/commits/803dfdadc37efea6a9232fbec52f60d2af93878e .

src/dfx/src/commands/ledger/mod.rs
levifeldman: Sure! See https://github.com/dfinity/sdk/pull/3018/commits/803dfdadc37efea6a9232fbec52f60d2af93878e.

levifeldman: Do you rather a match statement for clarity or is unwrap_or() good?

levifeldman:
@ericswanson-dfinity When I do cargo run ledger <command> , it gives this error: 
Running `target/debug/dfx ledger show-subnet-types`
thread 'main' panicked at 'Cannot parse version.: Error("unexpected character '+' after major version number")', src/dfx/src/config/mod.rs:9:51

How can I fix this so I can run and test?

levifeldman:
@ericswanson-dfinity When I do `cargo run ledger <command>`, it gives this error:
Running target/debug/dfx ledger show-subnet-types
thread 'main' panicked at 'Cannot parse version.: Error("unexpected character '+' after major version number")', src/dfx/src/config/mod.rs:9:51

How can I fix this so I can run and test?

levifeldman:
Looks like macos does not support the $EPOCHSECONDS variable. The tests pass on ubuntu.

levifeldman:
Commit https://github.com/dfinity/sdk/pull/3018/commits/35fbc84dcfae7572be32e751a1fc85c34bb95784 switches `$EPOCHSECONDS` for `$(date +%s)` to support the macos e2e tests.

src/dfx/src/commands/ledger/mod.rs
```diff
@@ -147,7 +150,6 @@ pub async fn transfer(
                     .context("Failed to decode transfer response.")?;
                 match result {
                     Ok(block_height) => break block_height,
-                    Err(TransferError::TxDuplicate { duplicate_of }) => break duplicate_of,
```
ericswanson-dfinity, line 150: Please leave this in: it makes the call idempotent in the presence of fields that detect duplicates.

ericswanson-dfinity (changes requested):

src/dfx/src/commands/ledger/mod.rs
ericswanson-dfinity: This does mean that the tests will have to go through some extra work to make sure that there were no duplicate transfers.

levifeldman: The call is idempotent either way, this just makes sure the user knows that it is idempotent. Otherwise the user can think that 2 transfers are made when only 1 went through.

levifeldman: What is the benefit of swallowing the duplicate-transaction-error?

ericswanson-dfinity: Two benefits, one general, one specific:
1. General: If you're retrying, it's because the result of the first call was a failure.  There's a class of failures where an operation succeeds on the server, but communication fails before the caller finds out that the operation succeeded.

In this case, the advantage of "swallowing" the error is that you can make the call a second time, and check for success just as you did the first time.

Put another way, one can loop until `dfx ledger transfer --icp 1 --memo 1 --created-at-time "$t"` returns 0, and know that the operation succeeded exactly once.  Without this "swallowing", one must loop until the call either succeeds, or until it fails with a duplicate error.

2. Specific: `dfx ledger create-canister` [calls transfer_cmc()](https://github.com/dfinity/sdk/blob/master/src/dfx/src/commands/ledger/create_canister.rs#L83) (which in turn [calls transfer()](https://github.com/dfinity/sdk/blob/master/src/dfx/src/commands/ledger/mod.rs#L181)), and then calls notify_create() with the block height in order to obtain and display the canister id. If `transfer()` returns an error, then as currently written, `dfx ledger create-canister` will not call `notify` and will not display the principal of the newly-created canister.

levifeldman: Thanks. 

Commit https://github.com/dfinity/sdk/pull/3018/commits/b276bcf8d0eb2c9560e893ba33de5f7615744365 solves number 2.

About number 1, 

> Put another way, one can loop until dfx ledger transfer --icp 1 --memo 1 --created-at-time "$t" returns 0, and know that the operation succeeded exactly once. Without this "swallowing", one must loop until the call either succeeds, or until it fails with a duplicate error.

It may be easier in some scenarios, but can cause confusion in other scenarios. These are two different outcomes that would be blurred into one. A user can send multiple transactions and all would show success but only 1 would go through on the ledger. 

There is a reason the ledger canister returns Err(TransferError::TransactionDuplicate) on a duplicate-transaction. If the ledger would return Ok(height) even on a duplicate transaction it would cause confusion.

ericswanson-dfinity: Yes, there are tradeoffs.  Will discuss.

src/dfx/src/commands/ledger/mod.rs
```diff
@@ -117,15 +114,16 @@ pub async fn transfer(
...
-
-    let block_height: BlockHeight = loop {
+    let timestamp_nanos = created_at_time.unwrap_or(
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_nanos() as u64,
+    );
+
+    let block_height: BlockHeight = {
```
ericswanson-dfinity, line 126: This `loop` should remain, as well as the retry policy and retry handling in this method.  Notice that even in the absence of a `--created-at-time` CLI parameter, dfx passes passes a `created_at` timestamp and attempts retries in an idempotent manner.

```suggestion
    );

    let mut retry_policy = ExponentialBackoff::default();

    let block_height: BlockHeight = loop {
```

CHANGELOG.md
```diff
@@ -56,6 +56,12 @@ Added the ability to configure the WASM module used for assets canisters through
 
 ### fix: dfx deploy and icx-asset no longer retry on permission failure
 
+### feat: --created-at-time for the ledger functions: transfer, create-canister, and top-up
+
+### fix: ledger transfer duplicate transaction now returns an error.
+
+Before it was possible that a user could send 2 ledger transfers and both would show success but there would have been only 1 ledger transfer.
```
ericswanson-dfinity, line 63: > Before it was possible that a user could send 2 ledger transfers and both would show success but there would have been only 1 ledger transfer.

What's the scenario in which this description would be the case?  Two calls to `dfx ledger transfer` with identical `timestamp_nanos`?

ericswanson-dfinity (changes requested):
Regarding `dfx ledger transfer`, we should distinguish between two cases:
1. dfx knows about a successful completion (which includes the `TxDuplicate` "error")
2. dfx doesn't

In the first case, dfx should print success with additional information, in particular the block height and, if it is a duplicate, information that the transaction was processed in the past.  It should return exit code 0, indicating the successful completion.
This way, we effectively get idempotency because all calls are successful - but the additional metadata can still be used to determine how many transactions were actually executed.

In the second case, dfx should retry to hopefully end up in the first case.  If that doesn't happen, dfx should print an error message and return an appropriate error code.

CHANGELOG.md
levifeldman: > What's the scenario in which this description would be the case? Two calls to dfx ledger transfer with identical timestamp_nanos? 

Yes.

levifeldman:
> In the first case, dfx should print success with additional information, in particular the block height and, if it is a duplicate, information that the transaction was processed in the past. It should return exit code 0, indicating the successful completion.
This way, we effectively get idempotency because all calls are successful - but the additional metadata can still be used to determine how many transactions were actually executed.

If there is a loop in the transfer call logic, then if there is a network error on the first try, the second try will return a transaction duplicate response and if dfx returns exit code 0 but prints that this is a transaction of the past, the user will be confused because it is the transaction from the same dfx ledger transfer call.

ericswanson-dfinity:
> If there is a loop in the transfer call logic, then if there is a network error in the first loop try, the second loop try will return a transaction duplicate response and if dfx prints that this is a transaction of the past, the user can think the dfx ledger transfer call did not create a new transaction when in fact it did.

Understood.  It would have been better if I had written something like "dfx prints that this transaction completed successfully, although through a TxDuplicate response."  dfx can also generate output informing the caller of the errors that occur as it retries.

This is a result of displaying the TxDuplicate result at all.

If there were no loop, then it would be possible for the caller to call `dfx ledger transfer`, get an error, and then call `dfx ledger transfer` a second time, with `TxDuplicate` result (although successful).

CHANGELOG.md
ericswanson-dfinity: Okay.  It's fine to display that the result from the replica was `TxDuplicate`, but even in that case both `dfx ledger transfer` calls should return success, indicating that the transfer completed exactly once.

Note that according to https://internetcomputer.org/docs/current/developer-docs/integrations/icrc-1/#transaction-deduplication-, this would mean that `created_at_time`, `memo`, the transfer argument fields, and the caller all have the same values.

```
If the ledger observed a structurally equal transfer payload (i.e., all the transfer argument fields and the caller have the same values) at transaction with index i, it should return variant { Duplicate = record { duplicate_of = i } }.
```

levifeldman:
Ok, now the loop is back, with dfx printing different messages if the response is a TxDuplicate or an Ok(block) so the user can differentiate, and both responses complete with exit code 0. The transfer function prints "Waiting to retry after error: {agent_err}" and then "Sending duplicate transaction" before retrying the loop, so the user knows what dfx is doing.

e2e/tests-dfx/ledger.bash
```diff
@@ -117,6 +137,18 @@ tc_to_num() {
...
+    # Transaction Deduplication
+    t=$(current_time_nanoseconds)
+
+    assert_command dfx ledger top-up "$wallet" --icp 5 --created-at-time "$t"
+    assert_match "Canister was topped up with"
+
+    assert_command dfx ledger top-up "$wallet" --icp 5 --created-at-time $((t+1))
+    assert_match "Canister was topped up with"
+
+    run dfx ledger top-up "$wallet" --icp 5 --created-at-time "$t"
```
ericswanson-dfinity, line 150: what is `run`?

```suggestion
    assert_command dfx ledger top-up "$wallet" --icp 5 --created-at-time "$t"
```

e2e/tests-dfx/ledger.bash
```diff
@@ -125,6 +157,18 @@ tc_to_num() {
...
+    # Transaction Deduplication
+    t=$(current_time_nanoseconds)
+
+    assert_command dfx ledger create-canister --amount=100 --created-at-time "$t" "$(dfx identity get-principal)"
+    assert_match "Transfer sent at block height"
+
+    assert_command dfx ledger create-canister --amount=100 --created-at-time $((t+1)) "$(dfx identity get-principal)"
+    assert_match "Transfer sent at block height"
+
+    run dfx ledger create-canister --amount=100 --created-at-time "$t" "$(dfx identity get-principal)"
```
ericswanson-dfinity, line 170:
```suggestion
    assert_command dfx ledger create-canister --amount=100 --created-at-time "$t" "$(dfx identity get-principal)"
```

e2e/tests-dfx/ledger.bash
smallstepman: `run` is the way you’d run the command in up-to-date `bats`. With, e.g. bats 1.9, instead of assert_match, you’d split it two lines
```bash
run CMD
assert_success  # imported from bats-assert
```
I did try to see how difficult would it be to move us over to new version of bats here: https://github.com/smallstepman/sdk/pull/46 but the work was left unfinished, though I am planning to pick it up again at some point

ericswanson-dfinity: > I thought the command would fail because it is a duplicate transaction, do you know why it succeeds anyway?

From the perspective of the `dfx ledger top-up` command on line 150, it (the transfer portion, at least) is a duplicate of the `dfx ledger top-up` command on line 144.  They share the same `created_at_time`, `memo`, `amount`, and so forth.

Also from the perspective of the second invocation of `dfx ledger top-up` with the same parameters, we don't know that the previous call succeeded.  A plausible reason for running `dfx ledger top-up` with the exact same parameters would be that we are unsure whether the first command succeeded, or that part of it succeeded (`transfer_cmc`), and part of it failed (`notify_top_up`).

So in the context of the second call to `dfx ledger top-up` with the exact same parameters, dfx sees the `TxDuplicate` result, continues and reports success.

ericswanson-dfinity:
> @ericswanson-dfinity When I do `cargo run ledger <command>`, it gives this error: Running target/debug/dfx ledger show-subnet-types thread 'main' panicked at 'Cannot parse version.: Error("unexpected character '+' after major version number")', src/dfx/src/config/mod.rs:9:51
> 
> How can I fix this so I can run and test?

I guess you found a way around this.  I ran into the same when trying to run tests.  This happens when the branch is on a fork.  It seems that our code that figures out what to use as the dfx version generates a version string that doesn't parse as valid semver.  The workaround is to export `DFX_VERSION=0.14.0` or some other valid version.

e2e/tests-dfx/ledger.bash
```diff
@@ -60,12 +64,28 @@ teardown() {
...
+    t=$(current_time_nanoseconds)
+
+    assert_command dfx ledger transfer --icp 1 --memo 1 --created-at-time "$t" 345f723e9e619934daac6ae0f4be13a7b0ba57d6a608e511a00fd0ded5866752
+    assert_match "Transfer sent at block height"
+
+    assert_command dfx ledger transfer --icp 1 --memo 1 --created-at-time $((t+1)) 345f723e9e619934daac6ae0f4be13a7b0ba57d6a608e511a00fd0ded5866752
+    assert_match "Transfer sent at block height"
+
+    assert_command dfx ledger transfer --icp 1 --memo 1 --created-at-time "$t" 345f723e9e619934daac6ae0f4be13a7b0ba57d6a608e511a00fd0ded5866752
+    assert_match "transaction is a duplicate of another transaction in block"
```
ericswanson-dfinity, line 84:
```suggestion
    assert_match "transaction is a duplicate of another transaction in block"
    assert_match "Transfer sent at block height"
```

e2e/tests-dfx/ledger.bash
```diff
@@ -125,6 +157,18 @@ tc_to_num() {
...
+    t=$(current_time_nanoseconds)
+
+    assert_command dfx ledger create-canister --amount=100 --created-at-time "$t" "$(dfx identity get-principal)"
+    assert_match "Transfer sent at block height"
+
+    assert_command dfx ledger create-canister --amount=100 --created-at-time $((t+1)) "$(dfx identity get-principal)"
+    assert_match "Transfer sent at block height"
+
+    assert_command dfx ledger create-canister --amount=100 --created-at-time "$t" "$(dfx identity get-principal)"
+    assert_match "transaction is a duplicate of another transaction in block"
```
ericswanson-dfinity, line 171: same here with noting all of the outputs

e2e/tests-dfx/ledger.bash
```diff
@@ -125,6 +157,18 @@ tc_to_num() {
     assert_match "Transfer sent at block height"
     assert_match "Refunded at block height"
     assert_match "with message: Provided subnet type type1 does not exist"
+    
+    # Transaction Deduplication
+    t=$(current_time_nanoseconds)
+
+    assert_command dfx ledger create-canister --amount=100 --created-at-time "$t" "$(dfx identity get-principal)"
+    assert_match "Transfer sent at block height"
```
ericswanson-dfinity, line 165: Since this PR adds output, we should account for that here.
```suggestion
    # shellcheck disable=SC2154
    assert_contains "Transfer sent at block height" "$stdout"
    # shellcheck disable=SC2154
    assert_contains "Using transfer at block height" "$stdout"
```

ericswanson-dfinity (changes requested):

e2e/tests-dfx/ledger.bash
levifeldman: @smallstepman Either there or a stackoverflow post. The `assert_command_fail` from the utils file is a better fit for what I was trying to do.

levifeldman: > So in the context of the second call to dfx ledger top-up with the exact same parameters, dfx sees the TxDuplicate result, continues and reports success.

@ericswanson-dfinity Why is the notify part of the command succeeding? 
https://github.com/dfinity/ic/blob/3323f0fea730cef34233f6da48bbf043af5f0da4/rs/nns/cmc/src/main.rs#L1026

levifeldman: > So in the context of the second call to dfx ledger top-up with the exact same parameters, dfx sees the TxDuplicate result, continues and reports success.

@ericswanson-dfinity I was wondering why the notify part succeeds even if it's the second time calling notify with the same block height but I see [here](https://github.com/dfinity/ic/blob/3323f0fea730cef34233f6da48bbf043af5f0da4/rs/nns/cmc/src/main.rs#L1026) that the cmc returns the cached result of the first notification of that block height which is a success. Thanks.

e2e/tests-dfx/ledger.bash
levifeldman: On a duplicate transaction the output does not contain the line "Transfer sent at block height". 
Is it good to include the line "Transfer sent at block height" even on a duplicate transaction response?

levifeldman: Good call, done.

ericswanson-dfinity (approved):

levifeldman:
🎉🍾Awesome collaboration!

levifeldman:
My pleasure.

PR #3023, opened 2023-03-15
feat(frontend-canister): canister-side evidence computation

ericswanson-dfinity:
# Description

Add `compute_evidence` asset canister method, callable with Prepare permission.  It must be called over and over until it completes, which it will indicate by returning Some(evidence).

Fixes https://dfinity.atlassian.net/browse/SDK-966

# How Has This Been Tested?

Added an e2e test to verify ACL requirements, and unit tests.

# Checklist:


dfx-json (approved):
LGTM. Good set of tests. What does changing the number of iterations do? I didn't notice a test that alters this variable.

ericswanson-dfinity:
> LGTM. Good set of tests. What does changing the number of iterations do? I didn't notice a test that alters this variable.

Changing `max_iterations` limits how many times `compute_evidence()` will advance the evidence computation before saving its place and waiting for a subsequent call.  For example, [evidence_with_set_single_chunk_asset_content()](https://github.com/dfinity/sdk/pull/3023/files#diff-a0921620f2c0475cedcf421dcb3ff17fc2214d4cb2e38c0049729b07dc2e9a7aR1224) (go to tests.rs and "Load Diff" to jump to line) first calls `compute_evidence` with `max_iterations=Some(3)`, and gets a result of `None`, meaning the computation did not complete.  It then calls `compute_evidence` again with `max_iterations=Some(1)`, which is just enough to complete the computation, returning `Some(_)`.

PR #3030, opened 2023-03-20
feat(frontend-canister): `ic-asset` makes use of frontend canister api versioning 

smallstepman:
# Description

Summary of changes:
 
- added `fn api_version` canister call 
- `fn commit_batch` is now generic over `<T: CandidType + Sync>` for the `arg` parameter (previously: `arg: CommitBatchArguments`)
- `/canister_api/types/batch_upload.rs` is moved to the `/canister_api/types/batch_upload/` directory and further split into `common.rs` and `v0.rs`
- added logic to `fn upload` and `fn sync`, which checks the `api_version` of the canister. 

In preparation for:
- https://dfinity.atlassian.net/browse/SDK-819
- https://dfinity.atlassian.net/browse/SDK-565
- https://github.com/dfinity/sdk/pull/3002

Continuation of:
- #3022 

# How Has This Been Tested?

Covered by CI

# Checklist:


src/canisters/frontend/ic-asset/src/upload.rs
```diff
@@ -42,20 +47,22 @@ pub async fn upload(
...
-
-    let args = CommitBatchArguments {
-        batch_id,
-        operations,
-    };
-
-    commit_batch(canister, args).await?;
+    match api_version(canister).await {
+        0 => {
+            let operations = batch_upload::operations::v0::assemble_batch_operations(
```
ericswanson-dfinity, line 52: I recommend against versioning the "assemble_batch_operations" method and its dependencies.

Rather:
- unversioned `assemble_batch_operations` method returns the latest version of CommitBatchArguments
- this method checks `api_version()` and determines if it can express the CommitBatchArguments in a compatible version.  In the first case that we have to handle, that means: `v1::CommitBatchArguments` can be trivially converted to `v0::CommitBatchArguments` if there are no operations that set properties.

This will reduce greatly the volume of code duplication and version handling.

src/canisters/frontend/ic-asset/src/upload.rs
```diff
@@ -42,20 +47,22 @@ pub async fn upload(
...
+                AssetDeletionReason::Incompatible,
+            );
+            info!(logger, "Committing batch.");
+            let args = canister_api::types::batch_upload::v0::CommitBatchArguments {
+                batch_id,
+                operations,
+            };
+            commit_batch(canister, args).await?;
+        }
+        _ => bail!("unsupported API version"),
```
ericswanson-dfinity, line 64: We shouldn't need to bail here.  A v1 api canister should be able to accept a v0 CommitBatchArguments, because v0::BatchOperationKind is a subset of v1::BatchOperationKind.
```suggestion
        _ => bail!("unsupported API version"),
```

ericswanson-dfinity (changes requested):

src/canisters/frontend/ic-asset/src/upload.rs
smallstepman: I like where this is going.

One thing that is unclear to me is what should happen when `v1::CommitBatchArguments` contains `SetAssetProperties`, but the `api_version()` returns `0`? Should we just "force-convert" from v1 to v0 by dropping all of the `SetAssetProperties` from `v1::CommitBatchArguments`?

src/canisters/frontend/ic-asset/src/upload.rs
smallstepman: Thanks, I haven't thought about it from a _subset_-perspective, I will keep that in mind.
  
Currently, there is only one version of the API, which is `0`. This PR won't introduce API version `1`. I believe it's ok to `bail` for the unknown version, no?

ericswanson-dfinity: Consider if this code were used to deploy to a canister that had been upgraded to an api version 1 canister.  What should happen?

smallstepman: deploying to v1 canister with v0 tooling should succeed, I agree, but what happens when someone wants to deploy to v22 canister with v0 tooling? Are we planning to always change the API in a way that v0 is the subset of the new version?

smallstepman: I also feel that in this case, it's ok to expect the users to upgrade the tooling (under the condition the new version of the tooling supports all api versions)

smallstepman: as discussed, instead of bailing, we will rely on candid to return error for the incompatible payload

ericswanson-dfinity (approved):

PR #3036, opened 2023-03-21
refactor(cache): move code to `dfx-core`

smallstepman:
# Description

Moves most of the `cache`-related logic and errors to `dfx-core` 

# How Has This Been Tested?

covered by CI

# Checklist:


src/dfx/src/config/cache.rs
```diff
@@ -65,66 +58,11 @@ impl Cache for DiskBasedCache {
     }
 
     fn get_binary_command(&self, binary_name: &str) -> Result<std::process::Command, CacheError> {
+        Self::install(&self.version_str())?;
```
ericswanson-dfinity, line 61: Was this not happening before?

smallstepman: it was happening here: https://github.com/dfinity/sdk/pull/3036/files#diff-0c4331b4ea6b0fad92e885ffcb7a58541bef24817e1d495cdc56b8266f1319b2L219
but `fn get_binary_path_from_version` got moved to `dfx-core` so it can no longer access `fn install()`, so I had to reshuffle

ericswanson-dfinity: Okay.  Do you think `get_binary_command_path()` in this file (5 lines above) also needs to call Self::install?

smallstepman: 🙈 thanks!!

ericswanson-dfinity (approved):

PR #3059, opened 2023-03-30
refactor(dfx): move `build_wallet_canister` to `dfx-core` + concrete error types

smallstepman:
-------

# Description

Moves `build_wallet_canister` to `dfx-core`.
Introduces concrete error type.
In preparation to move `install_canister_wasm` to dfx-core

Fixes https://dfinity.atlassian.net/browse/SDK-1055

# How Has This Been Tested?

covered by C

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](httnps://www.conventionalcommits.org/en/v1.0.0/).


src/dfx-core/src/canister/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+pub async fn build_wallet_canister(
+    id: Principal,
+    agent: &Agent,
+) -> Result<WalletCanister<'_>, CanisterBuilderError> {
+    Ok(WalletCanister::from_canister(
+        ic_utils::Canister::builder()
+            .with_agent(agent)
+            .with_canister_id(id)
+            .build()
+            .unwrap(),
```
smallstepman, line 16: is this alright?

ericswanson-dfinity: fn_context made it very easy to add context at every level of the call stack.  I'm not sure we're adding much value here by adding an error type that mostly identifies the method being called.

What would you think about having this method's error type be `AgentError`?

ericswanson-dfinity (approved):

PR #3077, opened 2023-04-07
fix!: require `--clean` when network config changes

adamspofford-dfinity:
This PR ensures that any updated network configuration requires `dfx start` to be passed `--clean`, unless the user passes `--force`.

Implements SDK-519.

### DIFF
--- a/src/dfx/src/commands/replica.rs
+++ b/src/dfx/src/commands/replica.rs
@@ -12,16 +12,20 @@ use crate::lib::network::id::write_network_id;
 use crate::lib::replica_config::{HttpHandlerConfig, ReplicaConfig};
 use dfx_core::config::model::dfinity::DEFAULT_REPLICA_PORT;
 use dfx_core::config::model::local_server_descriptor::LocalServerDescriptor;
+use dfx_core::json::{load_json_file, save_json_file};
 use dfx_core::network::provider::{create_network_descriptor, LocalBindDetermination};
 
-use anyhow::Context;
+use anyhow::{bail, Context};
 use clap::Parser;
 use fn_error_context::context;
+use std::borrow::Cow;
 use std::default::Default;
 use std::fs;
 use std::fs::create_dir_all;
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
+
+use super::start::{CachedConfig, CachedReplicaConfig};
 
 /// Starts a local Internet Computer replica.
 #[derive(Parser)]
@@ -49,6 +53,10 @@ pub struct ReplicaOpts {
     /// The delay (in milliseconds) an update call should take. Lower values may be expedient in CI.
     #[clap(long, conflicts_with("emulator"), default_value = "600")]
     artificial_delay: u32,
+
+    /// Start even if the network config was modified.
+    #[clap(long)]
+    force: bool,
 }
 
 /// Gets the configuration options for the Internet Computer replica.
@@ -91,6 +99,7 @@ pub fn exec(
         enable_bitcoin,
         enable_canister_http,
         artificial_delay,
+        force,
     }: ReplicaOpts,
 ) -> DfxResult {
     let system = actix::System::new();
@@ -138,6 +147,8 @@ pub fn exec(
     let canister_http_adapter_socket_holder_path =
         local_server_descriptor.canister_http_adapter_socket_holder_path();
 
+    let previous_config_path = temp_dir.join("replica-effective-config.json");
+
     // dfx bootstrap will read these port files to find out which port to use,
     // so we need to make sure only one has a valid port in it.
     let replica_config_dir = local_server_descriptor.replica_configuration_dir();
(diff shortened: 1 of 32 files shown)


src/dfx/src/commands/start.rs
```diff
@@ -320,6 +325,26 @@ pub fn exec(
...
+                        .context(
+                        "Failed to read previous replica configuration. Rerun with `--clean`.",
+                    )?;
+                    if previous_config != replica_config {
+                        bail!("The network configuration was changed. Rerun with `--clean`.");
+                    }
+                }
+            }
+
+            fs::write(
```
ericswanson-dfinity, line 342: Please use dfx_core `save_json_file`, which includes the filename as context

src/dfx/src/commands/start.rs
```diff
@@ -320,6 +325,26 @@ pub fn exec(
                 }
             }
 
+            let previous_config_path = replica_config_dir.join("replica-effective-config.json");
```
ericswanson-dfinity, line 328: Let's store this in the network data directory, not the replica configuration directory.  Two reasons:
1. The replica configuration directory isn't especially useful, as it only stores the port and pid, but not actually the replica configuration.  We should probably instead store those two files in the network data directory
2. We need this "require clean" functionality on more than just the replica configuration, specifically to also include: replica/emulator mode, and replica commit hash

src/dfx/src/commands/start.rs
```diff
@@ -320,6 +325,26 @@ pub fn exec(
...
+
+            if !clean && !force {
+                if let Ok(previous_config) = fs::read(&previous_config_path) {
```
ericswanson-dfinity, line 331: Please use dfx_core load_json_path, which includes path information when there are errors.

ericswanson-dfinity (changes requested):
We need to save a little more information than just the ReplicaConfig, currently: if `--emulator` mode was used, and the replica commit hash.

ericswanson-dfinity (changes requested):
`dfx replica` will need similar treatment.

src/dfx/src/commands/replica.rs
```diff
@@ -178,8 +189,39 @@ pub fn exec(
...
+            }
+            save_json_file(&previous_config_path, &effective_config)
+                .context("Failed to write replica configuration")?;
             start_emulator_actor(env, shutdown_controller, emulator_port_path)?;
         } else {
+            let effective_config = CachedConfig {
+                config: CachedReplicaConfig::Replica {
+                    config: Cow::Borrowed(&replica_config),
+                },
+                replica_rev: env!("DFX_ASSET_REPLICA_REV").into(),
```
ericswanson-dfinity, line 212: Also available from `crate::lib::info::replica_rev`

src/dfx/src/commands/replica.rs
```diff
@@ -178,8 +189,39 @@ pub fn exec(
     system.block_on(async move {
         let shutdown_controller = start_shutdown_controller(env)?;
         if emulator {
+            let effective_config = CachedConfig {
+                replica_rev: env!("DFX_ASSET_REPLICA_REV").into(),
+                config: CachedReplicaConfig::Emulator,
+            };
```
ericswanson-dfinity, line 195: There's some opportunity for DRY here.
```suggestion
            let effective_config = CachedConfig::emulator();
```

src/dfx/src/commands/replica.rs
```diff
@@ -178,8 +189,39 @@ pub fn exec(
...
+                )?;
+                if effective_config != previous_config {
+                    bail!("The network configuration was changed. Run `dfx start` with `--clean`.")
+                }
+            }
+            save_json_file(&previous_config_path, &effective_config)
+                .context("Failed to write replica configuration")?;
             start_emulator_actor(env, shutdown_controller, emulator_port_path)?;
         } else {
+            let effective_config = CachedConfig {
```
ericswanson-dfinity, line 208: suggestion for DRY
```suggestion
            let effective_config = CachedConfig::replica(&replica_config);
```

src/dfx/src/commands/start.rs
```diff
@@ -209,6 +218,8 @@ pub fn exec(
         empty_writable_path(local_server_descriptor.icx_proxy_pid_path())?;
     let webserver_port_path = empty_writable_path(local_server_descriptor.webserver_port_path())?;
 
+    let previous_config_path = network_temp_dir.join("replica-effective-config.json");
```
ericswanson-dfinity, line 221: DRY + "one of these things is not like the other"
```suggestion
    let previous_config_path = local_server_descriptor.replica_effective_config_path();
```

src/dfx/src/commands/replica.rs
```diff
@@ -138,6 +147,8 @@ pub fn exec(
     let canister_http_adapter_socket_holder_path =
         local_server_descriptor.canister_http_adapter_socket_holder_path();
 
+    let previous_config_path = temp_dir.join("replica-effective-config.json");
```
ericswanson-dfinity, line 150:
```suggestion
    let previous_config_path = local_server_descriptor.replica_effective_config_path();
```

ericswanson-dfinity (changes requested):
There are a few opportunities for DRY here.  Can you find more?

ericswanson-dfinity (changes requested):

src/dfx/src/commands/replica.rs
```diff
@@ -178,8 +188,31 @@ pub fn exec(
...
+                let previous_config = load_json_file(&previous_config_path).context(
+                    "Failed to read replica configuration. Run `dfx start` with `--clean`.",
+                )?;
+                if effective_config != previous_config {
+                    bail!("The network configuration was changed. Run `dfx start` with `--clean`.")
+                }
+            }
+            save_json_file(&previous_config_path, &effective_config)
+                .context("Failed to write replica configuration")?;
+
```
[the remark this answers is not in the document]
adamspofford-dfinity: I wanted it to look basically the same as the code in start.rs, which *does* need to stay that way, so both can easily be updated at the same time. But if you say so.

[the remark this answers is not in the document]
ericswanson-dfinity: Understood.  My intent was to follow up with a change to start.rs to move that configuration out of the block, since I didn't want to ask you to do that here as it's adjacent work.

ericswanson-dfinity (dismissed):

src/dfx/src/commands/replica.rs
```diff
@@ -175,6 +185,21 @@ pub fn exec(
...
+    if !force && previous_config_path.exists() {
+        let previous_config = load_json_file(&previous_config_path)
+            .context("Failed to read replica configuration. Run `dfx start` with `--clean`.")?;
+        if effective_config != previous_config {
+            bail!("The network configuration was changed. Run `dfx start` with `--clean`.")
+        }
+    }
+    save_json_file(&previous_config_path, &effective_config)
+        .context("Failed to write replica configuration")?;
+
```
ericswanson-dfinity, line 202: I missed something: this can't be moved here easily for the time being, because `replica_config` is altered below on lines 239-250 immediately before the call to `start_replica_actor`.

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #3083, opened 2023-04-12
docs: small typo in dfx-ledger.md

Web3NL:


smallstepman (approved):
many thanks @Web3NL !

smallstepman (changes requested):
I just found out, that both ` ``` bash ` and ` ```bash` will work(disregard space at the beginning in 2nd example), see below:
``` bash
echo "something" | cat XYZ
``` 

```bash
echo "something" | cat XYZ
``` 

Because of this, I think we would prefer to leave the ` ``` bash` as it is. The reason behind it is these markdown docs are also being used to [build our website](https://internetcomputer.org/docs/current/references/cli-reference), and I can't predict how changing the style of starting code block will affect the website. If you could revert that particular change that would be great.

But nevertheless, this PR still improves the visual aspect of looking at the tables in the code editor (i.e. columns are properly aligned when looking at the raw text), so I'll accept this PR once you'd revert to the other thing I mentioned.

Web3NL:
excuse me! What a pain

smallstepman:
oh wow, haha, yeah, I didn't even notice the missing hyphen when I was reviewing the first time ^^

smallstepman (approved):
o wow, there i

PR #3091, opened 2023-04-17
refactor: methods that read canister info from state tree

ericswanson-dfinity:
# Description

Upcoming work to support the bitcoin canister needs to determine if a canister exists, what its controllers are, and whether or not it has been installed (has a module hash).

Extracted methods from the `canister info` command for this purpose, while letting the command continue to handle formatting of output.  This turned out to simplify a few other places that also read the module hash.

# How Has This Been Tested?

Covered by CI

# Checklist:

### DIFF
--- a/src/dfx/src/commands/wallet/upgrade.rs
+++ b/src/dfx/src/commands/wallet/upgrade.rs
@@ -3,9 +3,10 @@ use crate::lib::error::DfxResult;
 use crate::lib::identity::wallet::wallet_canister_id;
 use crate::lib::operations::canister::install_wallet;
 use crate::lib::root_key::fetch_root_key_if_needed;
+use crate::lib::state_tree::canister_info::read_state_tree_canister_module_hash;
+
 use anyhow::{anyhow, bail};
 use clap::Parser;
-use ic_agent::AgentError;
 use ic_utils::interfaces::management_canister::builders::InstallMode;
 
 /// Upgrade the wallet's Wasm module to the current Wasm bundled with DFX.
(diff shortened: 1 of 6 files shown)


src/dfx/src/commands/wallet/upgrade.rs
```diff
@@ -36,19 +37,12 @@ pub async fn exec(env: &dyn Environment, _opts: UpgradeOpts) -> DfxResult {
...
-        // If the canister is empty, this path does not exist.
-        // The replica doesn't support negative lookups, therefore if the canister
-        // is empty, the replica will return lookup_path([], Pruned _) = Unknown
-        Err(AgentError::LookupPathUnknown(_)) | Err(AgentError::LookupPathAbsent(_)) => {
-            bail!("The cycles wallet canister is empty. Try running `dfx identity deploy-wallet` to install code for the cycles wallet in this canister.")
-        }
-        Err(x) => bail!(x),
-        _ => {}
-    };
+        bail!("The cycles wallet canister is empty. Try running `dfx identity deploy-wallet` to install code for the cycles wallet in this canister.")
```
dfx-json, line 44: would it be beneficial to guide users to `dfx quickstart` instead?

dfx-json (approved):

src/dfx/src/commands/wallet/upgrade.rs
ericswanson-dfinity: Possibly, but not in the context of this PR, which refactors existing code without changing functionality.  Please note that the error message is moved here from line 47 in the previous version.

If it's desired to answer this question and make some improvement here, then in my opinion it would be better to track that investigation and work separately, in order to avoid distractions from the task at hand.

ericswanson-dfinity: Also, `dfx quickstart` always operates on the `ic` network, while `dfx wallet upgrade` and `dfx identity deploy-wallet` accept a `--network` parameter.  But maybe `dfx diagnose` or `dfx fix`?

PR #3090, opened 2023-04-14
feat: expose `wasm-opt` optimizer in `ic-wasm` to users

kentosugama:
# Description
[LANG-124](https://dfinity.atlassian.net/browse/LANG-124)

[Profiling data](https://docs.google.com/document/d/1ICXF083-hfRZr2OfRUk8OxmZqYDQL4XINHQ0FXuDqTA/edit) suggests that it would be useful to take advantage of the performance and binary size wasm optimizations offered by [`wasm-opt`](https://github.com/WebAssembly/binaryen). 

See also results in the [canister profiling repo](https://github.com/dfinity/canister-profiling/pull/35).

This tool has been integrated with `ic-wasm` in this PR: https://github.com/dfinity/ic-wasm/pull/28. The tool is added under the `shrink` command that is currently used in `dfx` to perform binary size reduction of canisters.

The next step is to expose this feature to users by allowing users to opt into the optimizer and specify the optimization level in `dfx.json`. Then to invoke `ic-wasm` in `dfx` according to this argument, similarly to `shrink`.

Note: This tool has been used in the past to perform binary size reductions for canisters, but was replaced because it deletes the custom metadata sections from wasm modules. The `ic-wasm` feature invokes `wasm-opt` while preserving these sections.

Example usage in `dfx.json`:
```
"canisters" : {
  "backend" : {
    "optimize" : "cycles"
  }
}
```

# How Has This Been Tested?
The following are `build.bash` tests, ensuring that builds succeed with optimizations turned on, but that the feature is turned off by default, especially for custom canisters.

```
@test "build succeeds if enable optimize" {
    jq '.canisters.e2e_project_backend.optimize="cycles"' dfx.json | sponge dfx.json
    dfx_start
    dfx canister create --all
    assert_command dfx build
}

@test "build custom canister default no optimize" {
    install_asset custom_canister
    install_asset wasm/identity

    dfx_start
    dfx canister create --all
    assert_command dfx build custom
    assert_not_match "Optimize"

    jq '.canisters.custom.optimize="size"' dfx.json | sponge dfx.json
    assert_command dfx build custom
    assert_match "Optimize"
}
```

The following is a test that ensures that `dfx start` correctly rejects invalid optimization options in `dfx.json`.
```
@test "invalid optimization level" {
    jq '.canisters.bad_optimization_level.optimize="bad_level"' dfx.json | sponge dfx.json
    assert_command_fail dfx_start
    assert_match "expected one of "
}
```

# Checklist:

[...]

PR #3095, opened 2023-04-19
feat: dfx deps subcommands

lwshang:
Fixes SDK-794

# Description

This feature was named `dfx pull` before. To make a complete, intuitive user experience, we present a set of subcommands under `dfx deps`:
- `dfx deps pull`: pull the dependencies from mainnet and generate `deps/pulled.json`, the candid files of direct dependencies will also be put into `deps/`;
- `dfx deps init`: set the init arguments for the pulled dependencies and save the data in `deps/init.json`;
- `dfx deps deploy`: deploy the pulled dependencies on local replica with the init arguments recorded in `deps/init.json`;

# How Has This Been Tested?

e2e

# Checklist:


ericswanson-dfinity (changes requested):
Missing docs in docs/cli-reference

src/dfx-core/src/config/model/dfinity.rs
```diff
@@ -769,6 +768,27 @@ impl ConfigInterface {
...
+                if let CanisterTypeProperties::Pull { id } = v.type_specific {
+                    match id_to_name.get(&id) {
+                        Some(other_name) => {
+                            return Err(PullCanistersSameId(other_name.to_string(), k.clone(), id));
+                        }
+                        None => {
+                            res.insert(k.clone(), id);
+                            id_to_name.insert(id, k);
+                        }
+                    }
```
ericswanson-dfinity, line 786:
```suggestion
                    if let Some(other_name) = id_to_name.get(&id) {
                        return Err(PullCanistersSameId(other_name.to_string(), k.clone(), id));
                    }
                    res.insert(k.clone(), id);
                    id_to_name.insert(id, k);

```

src/dfx-core/src/error/canister_id_store.rs
```diff
@@ -32,6 +35,9 @@ pub enum CanisterIdStoreError {
         canister_id: String,
         cause: UnifiedIoError,
     },
+
+    #[error(transparent)]
+    PullIdsError(#[from] DfxConfigError),
```
ericswanson-dfinity, line 40: A transparent error just passes through the underlying error, without an assumption that it is about some other specific thing
```suggestion
    DfxConfigError(#[from] DfxConfigError),
```

lwshang:
> Missing docs in docs/cli-reference

I created another Jira ticket (SDK-1086) for the docs work so that I can focus on feature implementation in this PR.

src/dfx/src/commands/deps/delete.rs
```diff
@@ -0,0 +1,59 @@
...
+use slog::{info, Logger};
+
+/// Install pulled canisters.
+#[derive(Parser)]
+pub struct DepsDeleteOpts {
+    /// Specify the canister to delete. You can specify its name (as defined in dfx.json) or Principal.
+    canister: String,
+}
+
+pub async fn exec(env: &dyn Environment, opts: DepsDeleteOpts) -> DfxResult {
```
ericswanson-dfinity, line 21: I'm not sure I understand the usefulness of this command.   From the changelog:
> delete the pulled dependencies;
I would have expected this to delete files in `deps/` previously pulled from `dfx deps pull`.

After understanding that this is about canisters, I would have expected this to delete pulled canisters and their dependencies.

But this command stops and deletes a single pulled canister, without taking pulled dependencies into account.  Wouldn't we expect `dfx canister stop` and `dfx canister delete` to work for this purpose?

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+use ic_agent::{Agent, AgentError};
+use ic_wasm::metadata::get_metadata;
+use sha2::{Digest, Sha256};
+use slog::{error, info, trace, warn, Logger};
+
+/// Pull canisters upon which the project depends
+#[derive(Parser)]
+pub struct DepsPullOpts {}
+
+pub async fn exec(env: &dyn Environment, _opts: DepsPullOpts) -> DfxResult {
```
ericswanson-dfinity, line 34:
```
$ dfx-r deps pull                 
Found 0 dependencies:
Error: Failed to write pulled.json
Caused by: Failed to write pulled.json
  Failed to write JSON file: Failed to write to /Users/ericswanson/w/deps/proj/deps/pulled.json: No such file or directory (os error 2)
```

ericswanson-dfinity (changes requested):
I created a new project and tried this:

```
$ dfx-r deps deploy
Error: Failed to read pulled.json. Please (re)run `dfx deps pull`.
Caused by: Failed to read pulled.json. Please (re)run `dfx deps pull`.
  Failed to read JSON file: Failed to read /Users/ericswanson/w/deps/xxx/deps/pulled.json: No such file or directory (os error 2)

$ dfx-r deps pull                 
Found 0 dependencies:
Error: Failed to write pulled.json
Caused by: Failed to write pulled.json
  Failed to write JSON file: Failed to write to /Users/ericswanson/w/deps/xxx/deps/pulled.json: No such file or directory (os error 2)

$ dfx-r deps init
Error: Failed to read pulled.json. Please (re)run `dfx deps pull`.
Caused by: Failed to read pulled.json. Please (re)run `dfx deps pull`.
  Failed to read JSON file: Failed to read /Users/ericswanson/w/deps/xxx/deps/pulled.json: No such file or directory (os error 2)
```

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,105 @@
...
+    Ok(())
+}
+
+// not use operations::canister::create_canister because we don't want to modify canister_id_store
+#[context("Failed to create canster {}", canister_id)]
+async fn try_create_canister(agent: &Agent, logger: &Logger, canister_id: &Principal) -> DfxResult {
+    info!(logger, "Creating canister: {canister_id}");
+    let mgr = ManagementCanister::create(agent);
+    // ignore the error that the canister is already installed
+    let _res = mgr
```
ericswanson-dfinity, line 79: Is "canister is already installed" really the only possible error here?

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,105 @@
...
+    let project_root = env.get_config_or_anyhow()?.get_project_root().to_path_buf();
+    let pulled_json = load_pulled_json(&project_root)?;
+    let pull_canisters_in_config = get_pull_canisters_in_config(env)?;
+    validate_pulled(&pulled_json, &pull_canisters_in_config)?;
+
+    let init_json = load_init_json(&project_root)?;
+
+    fetch_root_key_if_needed(env).await?;
+    let agent = env
+        .get_agent()
```
ericswanson-dfinity, line 37: If the intent is for dfx to deploy pulled canisters with the anonymous identity as the controller, then we can't use the environment's agent here.

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,105 @@
...
+#[context("Failed to install canster {}", canister_id)]
+async fn install_pulled_canister(
+    agent: &Agent,
+    logger: &Logger,
+    canister_id: &Principal,
+    install_args: Vec<u8>,
+) -> DfxResult {
+    info!(logger, "Installing canister: {canister_id}");
+    let pulled_canister_path = get_pulled_wasm_path(*canister_id)?;
+    let wasm = std::fs::read(pulled_canister_path)?;
```
ericswanson-dfinity, line 96: Please use dfx_core::fs::read in order to include path context
```suggestion
    let wasm = dfx_core::fs::read(pulled_canister_path)?;
```

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,105 @@
...
+pub async fn exec(env: &dyn Environment, opts: DepsDeployOpts) -> DfxResult {
+    let logger = env.get_logger();
+    let project_root = env.get_config_or_anyhow()?.get_project_root().to_path_buf();
+    let pulled_json = load_pulled_json(&project_root)?;
+    let pull_canisters_in_config = get_pull_canisters_in_config(env)?;
+    validate_pulled(&pulled_json, &pull_canisters_in_config)?;
+
+    let init_json = load_init_json(&project_root)?;
+
+    fetch_root_key_if_needed(env).await?;
```
ericswanson-dfinity, line 35: Maybe there's nothing to be done here, but `dfx deps deploy` should never be run against mainnet, meaning we should always expect to fetch the root key here.

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,105 @@
...
+    let agent = env
+        .get_agent()
+        .ok_or_else(|| anyhow!("Cannot get HTTP client from environment."))?;
+
+    match opts.canister {
+        Some(canister) => {
+            let canister_id = match pull_canisters_in_config.get(&canister) {
+                Some(canister_id) => *canister_id,
+                None => Principal::from_text(canister).with_context(|| {
+                    "The canister is not a valid Principal nor a name specified in dfx.json"
```
ericswanson-dfinity, line 45: More specifically, the name of a `type: pull` canister in dfx.json

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,105 @@
...
+    // ignore the error that the canister is already installed
+    let _res = mgr
+        .create_canister()
+        .as_provisional_create_with_specified_id(*canister_id)
+        .call_and_wait()
+        .await;
+    Ok(())
+}
+
+#[context("Failed to install canster {}", canister_id)]
```
ericswanson-dfinity, line 87:
```suggestion
#[context("Failed to install canister {}", canister_id)]
```

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    Ok(())
+}
+
+#[context("Failed to create and install canster {}", canister_id)]
+async fn create_and_install(
+    agent: &Agent,
+    logger: &Logger,
+    canister_id: &Principal,
```
ericswanson-dfinity, line 64: Should this somewhere also install dependencies of the pulled canister?  Or is that for later work?

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,154 @@
...
+                        types.push(env.ast_to_type(ty)?);
+                    }
+                    match types.is_empty() {
+                        true => init_json.set_empty_init(*canister_id),
+                        false => canisters_require_init.push(*canister_id),
+                    }
+                }
+            }
+            if !canisters_require_init.is_empty() {
+                let mut message = "Following canister(s) require init argument, please run `dfx deps init <PRINCIPAL>` to set them individually:".to_string();
```
ericswanson-dfinity, line 116: Grammar/run-on sentence
```suggestion
                let mut message = "The following canister(s) require an init argument.  Please run `dfx deps init <PRINCIPAL>` to set them individually:".to_string();
```

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,154 @@
...
+                    } else {
+                        let bytes = args_to_bytes(arg_str, &env, &types)?;
+                        init_json.set_init_arg(canister_id, Some(arg_str.to_string()), &bytes);
+                    }
+                }
+                (Some(_), true) => {
+                    bail!("Canister {canister_id} takes no init argument. Please rerun without `--argument`");
+                }
+                (None, false) => {
+                    let mut message = format!("Canister {canister_id} requires init argument, following info might be helpful.");
```
ericswanson-dfinity, line 81: Run-on sentence
```suggestion
                    let mut message = format!("Canister {canister_id} requires an init argument.  The following info might be helpful:");
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    match any_download_fail {
+        true => bail!("Failed when pulling canisters."),
+        false => {
+            for (name, canister_id) in &pull_canisters_in_config {
+                copy_service_candid_to_project(&project_root, name, *canister_id)?;
+            }
+        }
+    }
```
ericswanson-dfinity, line 92:
```suggestion
    if any_download_fail {
        bail!("Failed when pulling canisters.");
    }

    for (name, canister_id) in &pull_canisters_in_config {
        copy_service_candid_to_project(&project_root, name, *canister_id)?;
    }
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+    canisters_to_pull: &mut VecDeque<Principal>,
+    pulled_json: &mut PulledJson,
+) -> DfxResult {
+    info!(
+        logger,
+        "Resolving dependencies of canister {canister_id}..."
+    );
+
+    match fetch_metatdata(agent, canister_id, DFX_DEPS).await {
+        Ok(Some(deps_raw)) => {
```
ericswanson-dfinity, line 112: Typo, and no need to match on the Result
```suggestion
    match fetch_metadata(agent, canister_id, DFX_DEPS).await? {
        Some(deps_raw) => {
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+                        let dep_id = Principal::from_text(p)
+                            .with_context(|| format!("`{p}` is not a valid Principal."))?;
+                        canisters_to_pull.push_back(dep_id);
+                        deps.push(dep_id);
+                    }
+                    None => bail!(
+                        "Failed to parse `dfx:deps` entry: {entry}. Expected `name:Principal`. "
+                    ),
+                }
+            }
```
ericswanson-dfinity, line 127: Please extract this to a method

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+) -> DfxResult {
+    info!(logger, "Pulling canister {canister_id}...");
+
+    let agent = env
+        .get_agent()
+        .ok_or_else(|| anyhow!("Cannot get HTTP client from environment."))?;
+
+    // try fetch `dfx:wasm_hash`. If not available, get the hash of the on chain canister.
+    let hash_on_chain = match fetch_metatdata(agent, canister_id, DFX_WASM_HASH).await {
+        Ok(Some(wasm_hash_raw)) => {
```
ericswanson-dfinity, line 167: Use of `?` rather than match on Result
```suggestion
    let hash_on_chain = match fetch_metatdata(agent, canister_id, DFX_WASM_HASH).await? {
        Some(wasm_hash_raw) => {
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+            match canister_status.module_hash {
+                Some(hash_on_chain) => hash_on_chain,
+                None => {
+                    bail!("Canister {canister_id} doesn't have module hash. Perhaps it's not installed.");
+                }
+            }
+        }
+        Err(e) => {
+            bail!(e);
+        }
```
ericswanson-dfinity, line 187: Almost anytime Err(e) => bail!(e) means might as well check `?` outside of the match
```suggestion
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+    };
+
+    pulled_canister.wasm_hash = hex::encode(&hash_on_chain);
+
+    // will save wasm and candid in $(cache_root)/pulled/{canister_id}/
+    let canister_dir = get_cache_root()?
+        .join("pulled")
+        .join(canister_id.to_string());
+    std::fs::create_dir_all(&canister_dir)
+        .with_context(|| format!("Failed to create dir at {:?}", &canister_dir))?;
```
ericswanson-dfinity, line 197:
```suggestion
    dfx_core::fs::create_dir_all(&canister_dir)?;
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    let wasm_path = get_pulled_wasm_path(canister_id)?;
+
+    // skip download if cache hit
+    let mut cache_hit = false;
+    let wasm_url_txt_path = get_wasm_url_txt_path(canister_id)?;
+    if wasm_path.exists() {
+        let bytes = std::fs::read(&wasm_path)?;
```
ericswanson-dfinity, line 205: dfx_core::fs to include context (path info)
```suggestion
        let bytes = dfx_core::fs::read(&wasm_path)?;
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+        if status.is_client_error() || status.is_server_error() {
+            bail!("Failed to download wasm from url: {wasm_url}.\n  StatusCode: {status}");
+        }
+        let content = response.bytes().await?;
+
+        // hash check
+        let mut hasher = Sha256::new();
+        hasher.update(&content);
+        let hash_download = hasher.finalize();
+        if hash_download.as_slice() != hash_on_chain {
```
ericswanson-dfinity, line 239: Should this take into account `DFX_WASM_URL` if present?

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+    // try get `dfx:init` from downloaded wasm
+    match get_metadata(&wasm, DFX_INIT) {
+        Ok(dfx_init) => pulled_canister.dfx_init = Some(dfx_init),
+        Err(_e) => trace!(
+            logger,
+            "Failed to get {} metadata from {:?}",
+            DFX_INIT,
+            &wasm_path
+        ),
+    }
```
ericswanson-dfinity, line 281: Suggestion: update to a newer version of `ic-wasm`, which changes this method to operate on a Module rather than a wasm blob, and return an `Option<String>` rather than a `Result<String>`.

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,340 @@
...
+        Err(agent_error) => match agent_error {
+            // replica returns such error
+            AgentError::HttpError(ref e) => {
+                let content = String::from_utf8(e.content.clone())?;
+                if content.starts_with("Custom section") {
+                    Ok(None)
+                } else {
+                    bail!(agent_error);
+                }
+            }
```
ericswanson-dfinity, line 306: The replica returns 404 in this case, which is pretty intuitive.  I would suggest both of the following:
1. check for status 404
2. check for `.starts_with("Custom section {metadata} not found")`
```
The replica returned an HTTP Error: Http Error: status 404 Not Found, content type "application/cbor", content: Custom section a:b:c not found.
```

src/dfx/src/commands/deps/delete.rs
lwshang: That's a great point. I'm going to remove this subcommand.

src/dfx/src/commands/deps/deploy.rs
lwshang: Of course there can be other possible error. I'm not sure if there are better approach to handle them.
If other error happens, the following "canister install" will fail and give some information there.

src/dfx/src/lib/builders/pull.rs
```diff
@@ -1,16 +1,18 @@
 use crate::lib::builders::{
     BuildConfig, BuildOutput, CanisterBuilder, IdlBuildOutput, WasmBuildOutput,
 };
+use crate::lib::canister_info::pull::PullCanisterInfo;
 use crate::lib::canister_info::CanisterInfo;
 use crate::lib::environment::Environment;
 use crate::lib::error::DfxResult;
 use crate::lib::models::canister::CanisterPool;
 
+use std::path::PathBuf;
```
ericswanson-dfinity, line 10: Please move this (back) into the group of imports beginning below on line 12

src/dfx/src/commands/deps/deploy.rs
ericswanson-dfinity: Suggestion: add a method like this to state_tree/canister_info.rs and call it

```
pub async fn read_state_tree_canister_exists(
    agent: &Agent,
    canister_id: Principal,
) -> DfxResult<bool> {
    matches!(read_state_tree_canister_controllers(agent, canister_id).await, Ok(Some(_)))
}
```

src/dfx/src/commands/deps/deploy.rs
lwshang: IIUC, `fetch_root_key_if_needed` will fetch the root key when it is not the mainnet so this is necessary and it do make the call every time.

Do yo mean replacing this line with `agent.fetch_root_key()?`?

Both work and `fetch_root_key_if_needed` is more common in dfx command implementation.

src/dfx/src/commands/deps/deploy.rs
lwshang: `dfx deps pull` and `dfx deps init` fetches all direct and indirect dependencies and set their init arguments.
Then `dfx deps deploy` can deploy them in arbitrary order. There is no requirement that the dependency should be deployed before the dependent.

It's possible that the user forget to deploy some indirect dependencies. When they make calls that reach the dependencies, the replica will error that the canister is not installed.

src/dfx/src/commands/deps/pull.rs
lwshang: Could you elaborate this a bit?

The code just downloaded from `DFX_WASM_URL` above.

src/dfx/src/commands/deps/pull.rs
lwshang: I would like to make the change once this PR https://github.com/dfinity/sdk/pull/3090 merged.

src/dfx/src/commands/deps/pull.rs
lwshang: https://github.com/dfinity/sdk/blob/6e033c4f17deae2ed659d6d98fedc515eac31d83/src/dfx/src/commands/deps/pull.rs#L153-L171

`hash_on_chain` is from above code which took `DFX_WASM_HASH` into account.

docs/concepts/canister-metadata.md
```diff
@@ -75,50 +75,68 @@ A string containing the path of a file containing the wasm section contents. Con
...
-A list of `name:ID` pairs of direct dependencies separated by semicolon.
+The "production" canister running on the mainnet should has following public metadata:
 
-This section must not be set in `dfx.json`. `dfx` will generate the content.
+- `dfx:deps`
+- `dfx:wasm_url`
+- `dfx:wasm_hash` (optional)
 
-### `dfx:init`
+The canister wasm downloaded from `dfx:wasm_url` should has following metadata (public or private):
```
ericswanson-dfinity, line 131:
```suggestion
A canister wasm downloaded from `dfx:wasm_url` should have the following metadata (public or private):
```

docs/concepts/canister-metadata.md
```diff
@@ -75,50 +75,68 @@ A string containing the path of a file containing the wasm section contents. Con
...
 
-A message to guide consumers how to initialize the canister.
+- `candid:service`
+- `candid:args`
+- `dfx:init`
 
-This section must be set explicitly.
+All the metadata sections with `dfx:` prefix must be set in `dfx.json` explicitly.
 
+Those two `candid:` metadata sections are handled by `dfx` during canister building.
```
ericswanson-dfinity, line 139:
```suggestion
The two `candid:` metadata sections are handled by `dfx` during canister building.
```

docs/concepts/canister-metadata.md
```diff
@@ -75,50 +75,68 @@ A string containing the path of a file containing the wasm section contents. Con
...
-### `dfx:init`
+The canister wasm downloaded from `dfx:wasm_url` should has following metadata (public or private):
 
-A message to guide consumers how to initialize the canister.
+- `candid:service`
+- `candid:args`
+- `dfx:init`
 
-This section must be set explicitly.
+All the metadata sections with `dfx:` prefix must be set in `dfx.json` explicitly.
```
ericswanson-dfinity, line 137:
```suggestion
All the metadata sections with the `dfx:` prefix must be set in `dfx.json` explicitly.
```

docs/concepts/canister-metadata.md
```diff
@@ -75,50 +75,68 @@ A string containing the path of a file containing the wasm section contents. Con
...
 A string containing the wasm section content directly. Conflicts with "path".
 
-## The candid:service metadata
+
+
+
+## Canister Metadata Standard
+
+This section describes each metadata field (_in alphabetical order_)
+
```
ericswanson-dfinity, line 84: These were meant as PR comments
```suggestion
```

src/dfx/src/commands/deps/pull.rs
ericswanson-dfinity: Ah, now I see.  I confused by this:
> get the hash of the on chain canister
Suggestion: rename `hash_on_chain` to `expected_hash`.

While technically they are both "on chain" (either the `dfx:wasm_hash` hash from metadata, or the `module_hash`), they are "on chain" in different ways.

src/dfx/src/commands/deps/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Parser)]
+enum SubCommand {
+    Pull(pull::DepsPullOpts),
+    Init(init::DepsInitOpts),
+    Deploy(deploy::DepsDeployOpts),
+}
+
+/// Executes `dfx deps` and its subcommands.
+pub fn exec(env: &dyn Environment, opts: DepsOpts) -> DfxResult {
+    let agent_env = create_agent_environment(env, opts.network.network)?;
```
ericswanson-dfinity, line 34: Recommend creating the anonymous agent environment here, since all of these commands should use the anonymous identity.  This will avoid prompting for credentials (for pem decryption) that aren't needed, as well as avoid holding onto an HSM session if so configured.

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,121 @@
...
+    // ignore the error that the canister is already installed
+    let _res = mgr
+        .create_canister()
+        .as_provisional_create_with_specified_id(*canister_id)
+        .call_and_wait()
+        .await;
+    match read_state_tree_canister_controllers(agent, *canister_id).await? {
+        Some(cs) if cs.len() == 1 && cs[0] == Principal::anonymous() => Ok(()),
+        Some(_) => bail!("Canister {canister_id} has been created before and its controller is not anonymous identity. Please stop and delete it and then deploy again."),
+        None => bail!("Canister {canister_id} has no controllers."),
```
ericswanson-dfinity, line 99: This would mean if there is an i/o or other unexpected error in the `create_canister` call, then this method would hide that error and display "Canister {canister_id} has no controllers."  What `None` here means is that the canister does not exist. Some(empty vec) means the canister exists but has no controllers.

```suggestion
        None => mgr.....create_canister()...,
```

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Specifies the data type of the init argument.
+    #[clap(long, requires("argument"), possible_values(&["idl", "raw"]))]
+    argument_type: Option<String>,
+}
+
+pub async fn exec(env: &dyn Environment, opts: DepsInitOpts) -> DfxResult {
+    let logger = env.get_logger();
+    let pull_canisters_in_config = get_pull_canisters_in_config(env)?;
+    if pull_canisters_in_config.is_empty() {
+        info!(logger, "There is no pull dependencies defined in dfx.json");
```
ericswanson-dfinity, line 38:
```suggestion
        info!(logger, "There are no pull dependencies defined in dfx.json");
```

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,158 @@
...
+                types.push(env.ast_to_type(ty)?);
+            }
+
+            let arguments = opts.argument.as_deref();
+            let arg_type = opts.argument_type.as_deref();
+
+            match (arguments, types.is_empty()) {
+                (Some(arg_str), false) => {
+                    if arg_type == Some("raw") {
+                        let bytes = hex::decode(arguments.unwrap_or(""))
```
ericswanson-dfinity, line 73: I don't see any e2e test that passes --argument-type raw

Also:
```suggestion
                        let bytes = hex::decode(arg_str)
```

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,158 @@
...
+                    info!(logger, "{canister_id} already set init argument.");
+                } else {
+                    let idl_path = get_service_candid_path(*canister_id)?;
+                    let (env, _) = check_candid_file(&idl_path)?;
+                    let candid_args = pulled_json.get_candid_args(canister_id)?;
+                    let candid_args_idl_types: IDLTypes = candid_args.parse()?;
+                    let mut types = vec![];
+                    for ty in candid_args_idl_types.args.iter() {
+                        types.push(env.ast_to_type(ty)?);
+                    }
```
ericswanson-dfinity, line 112: can this be simplified to the following?
```suggestion
                    let requires_init = !candid_args_idl_types.args.is_empty();
```

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,158 @@
...
+                    match types.is_empty() {
+                        true => init_json.set_empty_init(*canister_id),
+                        false => canisters_require_init.push(*canister_id),
+                    }
```
ericswanson-dfinity, line 116: Recommend `if/else` vs `match { true =>, false => }`
```suggestion
                    if requires_init {
                        canisters_require_init.push(*canister_id);         
                    } else {
                        init_json.set_empty_init(*canister_id);
                    }
```

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Specifies the init argument.
+    #[clap(long, requires("canister"))]
+    argument: Option<String>,
+
+    /// Specifies the data type of the init argument.
+    #[clap(long, requires("argument"), possible_values(&["idl", "raw"]))]
+    argument_type: Option<String>,
+}
+
+pub async fn exec(env: &dyn Environment, opts: DepsInitOpts) -> DfxResult {
```
ericswanson-dfinity, line 34: This method is pretty large.  Can you think of ways to break it up into smaller pieces?

Also, I think it would help the review process to have a description under docs/cli-reference that describes the intended functionality of the command, with examples.

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,338 @@
...
+
+/// Pull canisters upon which the project depends
+#[derive(Parser)]
+pub struct DepsPullOpts {}
+
+pub async fn exec(env: &dyn Environment, _opts: DepsPullOpts) -> DfxResult {
+    let logger = env.get_logger();
+    let pull_canisters_in_config = get_pull_canisters_in_config(env)?;
+    if pull_canisters_in_config.is_empty() {
+        info!(logger, "There is no pull dependencies defined in dfx.json");
```
ericswanson-dfinity, line 38:
```suggestion
        info!(logger, "There are no pull dependencies defined in dfx.json");
```

ericswanson-dfinity (changes requested):
Still more review to come, but in addition to the comments I have left, I'd like to see:
- Documentation in docs/cli-reference and possible a dedicated file in docs/concepts
- use `dfx_core::fs` rather than `std::fs` everywhere, in order to easily include pathname context

src/dfx/src/lib/deps.rs
```diff
@@ -0,0 +1,239 @@
...
+use candid::Principal;
+use fn_error_context::context;
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct PulledJson {
+    pub named: BTreeMap<String, Principal>,
+    pub canisters: BTreeMap<Principal, PulledCanister>,
+}
```
ericswanson-dfinity, line 23: Also: what is the rationale for/purpose of the `named` field in pulled.json?

ericswanson-dfinity (changes requested):
Review part 2 (more coming)

e2e/tests-dfx/deps.bash
```diff
@@ -0,0 +1,445 @@
...
+    assert_contains "The following canister(s) require an init argument. Please run \`dfx deps init <PRINCIPAL>\` to set them individually:"
+    assert_contains "$CANISTER_ID_A"
+    assert_contains "$CANISTER_ID_C"
+
+    assert_command dfx deps init "$CANISTER_ID_A" --argument 11
+    assert_command dfx deps init "$CANISTER_ID_C" --argument 33
+
+    # The argument is the hex string of '("abc")' which doesn't type check
+    # However, passing raw argument will bypass the type check so following command succeed
+    assert_command dfx deps init "$CANISTER_ID_A" --argument "4449444c00017103616263" --argument-type raw
```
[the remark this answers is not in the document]
lwshang: This is covered in the next e2e test "dfx deps deploy works"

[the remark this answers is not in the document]
ericswanson-dfinity: I should have been more specific:
 - with `--argument-type raw`
 - where the init succeeds
 - and shows the raw parameter value passed through as expected
 

The following test does have a `dfx deps init` with `--argument-type raw`, but it is expected to fail

src/dfx/src/lib/deps.rs
lwshang: I refactored pulled.json by removing `named` field` and add `name: Option<String>` to each `PulledCanister`.
The name will be displayed alongside the canister_id in `dfx deps init` and `dfx deps deploy`, e.g. "yhgn4-myaaa-aaaaa-aabta-cai (dep_b)"

src/dfx/src/lib/operations/canister/deploy_canisters.rs
```diff
@@ -99,7 +115,7 @@ pub async fn deploy_canisters(
     let pool = build_canisters(
         env,
         &canisters_to_load,
-        &canisters_to_deploy,
+        &canisters_to_build,
```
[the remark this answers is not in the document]
lwshang: Let me use the e2e project as an example.
The `app` canister depends on `dep_b` and `dep_c`.
When `dfx deploy app`, both `dep_b` and `dep_c` must be built so that their candid files can be copied to `.dfx/local/...` and the path of candids are required when building `app` canister.

`canisters_to_build` includes `dep_b`, `dep_c`, `app`.
`canisters_to_install` contains only `app`.

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,342 @@
...
+use std::collections::VecDeque;
+use std::io::Write;
+use std::path::Path;
+
+use anyhow::{anyhow, bail, Context};
+use candid::Principal;
+use clap::Parser;
+use fn_error_context::context;
+use ic_agent::{Agent, AgentError};
+// TODO: update the usage of this method once ic-wasm bump (#3090)
```
ericswanson-dfinity, line 25: 3090 is now merged

src/dfx/src/commands/deps/init.rs
```diff
@@ -0,0 +1,154 @@
...
+    Ok(())
+}
+
+#[context("Failed to validate argument against type defined in candid:args")]
+fn args_to_bytes(
+    arg_str: &str,
+    env: &TypeEnv,
+    types: &[Type], // types has been checked to not be empty
+) -> DfxResult<Vec<u8>> {
+    let first_char = arg_str.chars().next();
```
ericswanson-dfinity, line 135: This method looks like it was copy-pasted from [blob_from_arguments](https://github.com/dfinity/sdk/blob/master/src/dfx/src/util/mod.rs#L188).  If so, I'd suggest to extract the method (which may be this exact method) and call it from both places.

ericswanson-dfinity (changes requested):
Higher-level view: The commands are doing a lot, inlined in their `exec` methods.  I encourage you to see if you can think of a way to organize the code into components that provide higher-level functionality, and have the commands use those.

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
+# Pull Dependencies
+
+## Overview
+
+The interoperability of canisters on the Internet Computer (IC) is an important feature. 
+
+`dfx` provides a consistent developer workflow for integrating thrid party canisters.
```
ericswanson-dfinity, line 7:
```suggestion
`dfx` provides a consistent developer workflow for integrating third-party canisters.
```

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+
+A service provider attaches [necessary metadata](canister-metadata.md) to the canister wasm.
+
+A service consumer then can pull dependencies directly from the IC mainnet and easily deploy them on a local replica.
+
+This document describes the consumer workflow and explains what happens behind the scene.
```
ericswanson-dfinity, line 13:
```suggestion
This document describes the consumer workflow and explains what happens behind the scenes.
```

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+}
+```
+
+### 2. Pull the dependencies using `dfx deps pull`
+
+Running `dfx deps pull --network ic` will:
+
+1. resolve the dependency graph by fetching `dfx:deps` metadata recuresively;
```
ericswanson-dfinity, line 50:
```suggestion
1. resolve the dependency graph by fetching `dfx:deps` metadata recursively;
```

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+            "type": "pull",
+            "id": "yahli-baaaa-aaaaa-aabtq-cai"
+        }
+    }
+}
+```
+
+### 2. Pull the dependencies using `dfx deps pull`
+
+Running `dfx deps pull --network ic` will:
```
ericswanson-dfinity, line 48: Shouldn't `--network ic` be the default for `dfx deps pull`?  This would be the normal use case.  The other use cases would be `--network local` for CI, and `--network <some test net>` for test networks.

```suggestion
Running `dfx deps pull` will:
```

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+
+1. resolve the dependency graph by fetching `dfx:deps` metadata recuresively;
+2. download wasm of all direct and indirect dependencies from `dfx:wasm_url` into shared cache;
+3. hash check the downloaded wasm against `dfx:wasm_hash` metadata or the hash of the mainnet running canister;
```
ericswanson-dfinity, line 52:
```suggestion
3. verify the hash of the downloaded wasm against `dfx:wasm_hash` metadata or the hash of the canister deployed on mainnet;
```

src/dfx/src/lib/deps.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    Ok(())
+}
+
+fn get_deps_dir(project_root: &Path) -> PathBuf {
+    project_root.join("deps")
+}
+
+pub fn get_candid_path_in_project(project_root: &Path, name: &str) -> PathBuf {
+    get_deps_dir(project_root).join(name).with_extension("did")
```
ericswanson-dfinity, line 180: Since the canister id of a pulled canister won't change, but the name might (because it's specified by the developer in dfx.json), do you think it would make sense to store the candid file in `deps/<canister id>.did`?

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+For the example project, you will find following files in `deps/`:
+
+- `dep_b.did` and `dep_c.did`: candid files that can be imported by "app";
+- `pulled.json` which has follwing content:
+
+```
+{
+  "canisters": {
+    "yofga-2qaaa-aaaaa-aabsq-cai": {
+      "name": null,
```
ericswanson-dfinity, line 67: Please make it so the serializer omits this field if None, rather than writing it as `null` in the JSON.

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,342 @@
...
+    let mut cache_hit = false;
+    let wasm_url_txt_path = get_pulled_wasm_url_txt_path(canister_id)?;
+    if wasm_path.exists() {
+        let bytes = dfx_core::fs::read(&wasm_path)?;
+        let hash_cache = Sha256::digest(bytes);
+        if hash_cache.as_slice() == hash_on_chain {
+            cache_hit = true;
+            trace!(logger, "The canister wasm was found in the cache.");
+            let wasm_url_str = dfx_core::fs::read_to_string(&wasm_url_txt_path)
+                .with_context(|| format!("Failed to read {:?}", &wasm_url_txt_path))?;
```
ericswanson-dfinity, line 198: Context already provided
```suggestion
            let wasm_url_str = dfx_core::fs::read_to_string(&wasm_url_txt_path)?;
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,342 @@
...
+    let canister_dir = get_cache_root()?
+        .join("pulled")
+        .join(canister_id.to_string());
+    dfx_core::fs::create_dir_all(&canister_dir)?;
+
+    let wasm_path = get_pulled_wasm_path(canister_id)?;
+
+    // skip download if cache hit
+    let mut cache_hit = false;
+    let wasm_url_txt_path = get_pulled_wasm_url_txt_path(canister_id)?;
```
ericswanson-dfinity, line 190: I'm not sure it's worthwhile to store `dfx:wasm_url` in this wasm_url.txt file.  Nothing reads it, but we have to go to some trouble in the "cache hit" case to copy it from wasm_url.txt into pulled.json.

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,342 @@
...
+    if !cache_hit {
+        // fetch `dfx:wasm_url`
+        let wasm_url_raw = fetch_metadata(agent, canister_id, DFX_WASM_URL)
+            .await?
+            .ok_or_else(|| {
+                anyhow!("`{DFX_WASM_URL}` metadata not found in canister {canister_id}.")
+            })?;
+        let wasm_url_str = String::from_utf8(wasm_url_raw)?;
+        let wasm_url = reqwest::Url::parse(&wasm_url_str)?;
+        write_to_tempfile_then_rename(wasm_url_str.as_bytes(), &wasm_url_txt_path)?;
```
ericswanson-dfinity, line 211: This is the only place that uses `dfx:wasm_url`, and it uses it right away, so do we really need to store it in pulled.json?

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,342 @@
...
+        pulled_canister.wasm_url = Some(wasm_url_str);
+
+        // download
+        let response = reqwest::get(wasm_url.clone()).await?;
+        let status = response.status();
+        if status.is_client_error() || status.is_server_error() {
+            bail!("Failed to download wasm from url: {wasm_url}.\n  StatusCode: {status}");
+        }
```
ericswanson-dfinity, line 219: Let's have some retry logic here

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,342 @@
...
+    Ok(())
+}
+
+// download canister wasm, candid, init
+async fn download_canister_files(
+    env: &dyn Environment,
+    logger: &Logger,
+    canister_id: Principal,
+    pulled_canister: &mut PulledCanister,
+) -> DfxResult {
```
ericswanson-dfinity, line 151: This method has some tricky semantics.

First, it does more than `download canister files`: it also sets some, but not all, fields in `pulled_canister`.
Second, it has one of the following:
  - an unstated precondition that `pulled_canister` is mostly empty
  - an unstated promise that `download_canister_files` will always set any of the fields that it sometimes sets

It does happen that coming in to this method, `pulled_canister` will only have the `deps` field set.  If this were not the case, then the code fragment at `match get_metadata(&wasm, DFX_INIT)` would have a bug: if there were no `dfx:init` field, then it would leave the previous `dfx:init` value.

I'd like to suggest a different approach: use the canister wasm in the cache as the source of truth.  read `candid:args`, `dfx:init`, `dfx:deps` and so forth from it as needed.  Store only the expected module hash with the project, in order to be able to find out if the canister wasm changed from the perspective of this project.

And change this method signature to:
```suggestion
// download canister wasm, candid, init
async fn download_canister_files(
    env: &dyn Environment,
    logger: &Logger,
    canister_id: Principal,
) -> DfxResult<PulledCanister> {
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,342 @@
...
+            }
+        }
+    };
+
+    pulled_canister.wasm_hash = hex::encode(&hash_on_chain);
+
+    // will save wasm and candid in $(cache_root)/pulled/{canister_id}/
+    let canister_dir = get_cache_root()?
+        .join("pulled")
+        .join(canister_id.to_string());
```
ericswanson-dfinity, line 183: There are three methods (`get_pulled_wasm_path`, `get_pulled_service_candid_path`, and `get_pulled_wasm_url_txt_path`) that share this logic.  I'd extract this to a method and have all of them call it.

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+
+There are three dependencies:
+
+- "yhgn4-myaaa-aaaaa-aabta-cai": "dep_b" in `dfx.json`;
+- "yahli-baaaa-aaaaa-aabtq-cai": "dep_c" in `dfx.json`;
+- "yofga-2qaaa-aaaaa-aabsq-cai": an indirect dependency that both "dep_b" and "dep_c" depend on;
+
+### 3. Set init arguments using `dfx deps init`
+
+Running `dfx deps init` will iterate all dependencies in `pulled.json`, set empty argument for the ones need no init argument and print the list of dependencies that require init argument.
```
ericswanson-dfinity, line 106:
```suggestion
Running `dfx deps init` will iterate over all dependencies in `pulled.json`, set an empty argument for the ones that need no init argument and print the list of dependencies that do require an init argument.
```

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+Then running `dfx deps init <CANISTER> --argument <ARGUMENT>` to set init argument for individual dependency.
+
+The init arguments will be recorded in `deps/init.json`.
+
+For our example, we should run:
+
+```
+> dfx deps init
+WARN: The following canister(s) require an init argument. Please run `dfx deps init <NAME/PRINCIPAL>` to set them individually:
+yofga-2qaaa-aaaaa-aabsq-cai
```
ericswanson-dfinity, line 117: Note that the `dfx:init` metadata in the canister that referenced this indirect dependency does have a name for this dependency, but we discarded it.  Can display it here in an unambiguous way?

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+yahli-baaaa-aaaaa-aabtq-cai (dep_c)
+> dfx deps init yofga-2qaaa-aaaaa-aabsq-cai --argument 10
```
ericswanson-dfinity, line 119: It feels like we're skipping a step here: how does the developer know what to pass for the `--argument`? 

Would they first run `dfx deps init yofga-2qaaa-aaaaa-aabsq-cai` (not specifying an argument) which would display the `dfx:init` message and/or the `candid:args` ?

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+Creating canister: yofga-2qaaa-aaaaa-aabsq-cai
+Installing canister: yofga-2qaaa-aaaaa-aabsq-cai
+> dfx deps deploy dep_b
+Creating canister: yhgn4-myaaa-aaaaa-aabta-cai (dep_b)
+Installing canister: yhgn4-myaaa-aaaaa-aabta-cai (dep_b)
+```
+
+**Note**
+
+- `dfx deps deploy` always create the canister with the anonymous identity so that dependencies and application canisters will have different controllers;
```
ericswanson-dfinity, line 172:
```suggestion
- `dfx deps deploy` always creates the canister with the anonymous identity so that dependencies and application canisters will have different controllers;
```

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,202 @@
...
+- `dfx deps deploy` always install the canister in "reinstall" mode so that the canister status will be discarded;
```
ericswanson-dfinity, line 173:
```suggestion
- `dfx deps deploy` always installs the canister in "reinstall" mode so that the canister status will be discarded;
```

docs/concepts/pull-dependencies.md
lwshang: In my e2e example project, both `b` and `c` depends on `a`. It's possible that `b` and `c` set different names for `a` in their `dfx:deps`. Which name should we use here?

Another idea is to remove the `name` from `dfx:deps` so that it only contains a list of principals.

ericswanson-dfinity: I wondered what to do about the potentially different names too, but I don't have a clear answer either.

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,357 @@
...
+                hex::encode(hash_download.as_slice())
+            );
+        }
+
+        write_to_tempfile_then_rename(&content, &wasm_path)?;
+    }
+
+    // extract `candid:service` and save as candid file in shared cache
+    let wasm = dfx_core::fs::read(&wasm_path).context("Failed to read wasm")?;
+    let module = ic_wasm::utils::parse_wasm(&wasm, true)?;
```
ericswanson-dfinity, line 236: This fails with the following error if `wasm:url` references a .wasm.gz file:

```
Could not parse the data as WASM module. Bad magic number (at offset 0)
```

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,121 @@
...
+    if pull_canisters_in_config.is_empty() {
+        info!(logger, "There are no pull dependencies defined in dfx.json");
+        return Ok(());
+    }
+
+    let project_root = env.get_config_or_anyhow()?.get_project_root().to_path_buf();
+    let pulled_json = load_pulled_json(&project_root)?;
+    validate_pulled(&pulled_json, &pull_canisters_in_config)?;
+
+    let init_json = load_init_json(&project_root)?;
```
ericswanson-dfinity, line 38: If you run `dfx deps pull` and then `dfx deps deploy`, the error looks like below.  This should instead instruct the developer to run `dfx deps init`.

```
Trace mode enabled. Lots of logs coming up.
Error: Failed to read init.json
Caused by: Failed to read init.json
  Failed to read JSON file: Failed to read /Users/ericswanson/w/pulls/deps/init.json: No such file or directory (os error 2)
```

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,121 @@
...
+        Some(canister) => {
+            let canister_id = get_pull_canister_or_principal(canister, &pull_canisters_in_config)?;
+            vec![canister_id]
+        }
+        None => pulled_json.canisters.keys().copied().collect(),
+    };
+    for canister_id in canister_ids {
+        // Safe to unwrap:
+        // caniseter_ids are guranteed to exist in pulled.json
+        let pulled_canister = pulled_json.canisters.get(&canister_id).unwrap();
```
ericswanson-dfinity, line 55: It turns out this unwrap isn't safe.

Repro steps:
1. Set up a pullable canister
2. `dfx deps pull && dfx deps init`
3. `dfx deps deploy <some other canister id>`

I used this metadata, for example:
```
      "metadata": [
          {
              "name": "dfx:wasm_url",
              "content": "https://github.com/dfinity/sdk/raw/master/e2e/assets/wasm/identity/main.wasm"
          },
          {
              "name": "dfx:wasm_hash",
              "content": "bd27eb59219b5a810621d577c6d8a56ea68529fa53bbd855fdb8bd9ea328f2c3"
          },
          {
              "name": "dfx:deps",
              "content": ""
          }
      ]
```

Outcome:
```
$ dfx-r -v -v deps deploy zrrcc-3aaaa-aaaai-abzkq-cai
Trace mode enabled. Lots of logs coming up.
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/dfx/src/commands/deps/deploy.rs:55:71
```

src/dfx/src/commands/deps/deploy.rs
```diff
@@ -0,0 +1,121 @@
...
+    agent: &Agent,
+    logger: &Logger,
+    canister_id: &Principal,
+    canister_prompt: &str,
+) -> DfxResult {
+    info!(logger, "Creating canister: {canister_prompt}");
+    match read_state_tree_canister_controllers(agent, *canister_id).await? {
+        Some(cs) if cs.len() == 1 && cs[0] == Principal::anonymous() => Ok(()),
+        Some(_) => {
+            bail!("Canister {canister_id} has been created before and its controller is not anonymous identity. Please stop and delete it and then deploy again.");
```
ericswanson-dfinity, line 89:
```suggestion
            bail!("Canister {canister_id} has been created before and its controller is not the anonymous identity. Please stop and delete it and then deploy again.");
```

ericswanson-dfinity (changes requested):
Review in progress, but here is the next round

src/dfx/src/lib/metadata/dfx.rs
```diff
@@ -0,0 +1,18 @@
...
+//!
+//! The cli tool dfx should consolidate its usage of canister metadata into this single section
+//! It's originally for pulling dependencies. But open to extend for other usage.
+
+use dfx_core::config::model::dfinity::PullReady;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct DfxMetadata {
+    pub pull_ready: PullReady,
```
ericswanson-dfinity, line 11: Since pull support isn't a requirement for all canisters, I suggest making this an option now.  Otherwise the first step anyone will have to do to add a field is to make this an Option.
This will also let us specify the error message ("canister {} is not ...") rather than starting from a serde error.
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub pull_ready: Option<PullReady>,
```

src/dfx-core/src/config/model/dfinity.rs
```diff
@@ -231,12 +248,11 @@ pub struct ConfigCanistersCanister {
...
 
-    /// # Ready for dfx Pull
-    /// Whether or not to make this canister ready for dfx pull by other project.
+    /// # Ready to be pulled
+    /// Whether or not to make this canister ready for `dfx deps pull` by other project.
     /// If true, several required metadata fields must be also set with the correct format.
-    // TODO: Add a link to `dfx pull` document.
     #[serde(default)]
-    pub pull_ready: bool,
+    pub pull_ready: Option<PullReady>,
```
ericswanson-dfinity, line 255: Suggestion to use this synonym:

```suggestion
    pub pullable: Option<PullReady>,
```

src/dfx/src/commands/deps/pull.rs
```diff
@@ -0,0 +1,323 @@
...
+
+#[context("Failed to get dependencies of canister {canister_id}.")]
+async fn get_dependencies(
+    agent: &Agent,
+    logger: &Logger,
+    canister_id: &Principal,
+) -> DfxResult<Vec<Principal>> {
+    info!(
+        logger,
+        "Resolving dependencies of canister {canister_id}..."
```
ericswanson-dfinity, line 113:
```suggestion
        "Fetching dependencies of canister {canister_id}..."
```

src/dfx/src/util/mod.rs
```diff
@@ -1,18 +1,25 @@
...
 
-use anyhow::Context;
+use anyhow::{bail, Context};
+use backoff::backoff::Backoff;
+use backoff::ExponentialBackoff;
+use bytes::Bytes;
 use candid::parser::typing::{pretty_check_file, TypeEnv};
 use candid::types::{Function, Type};
 use candid::{parser::value::IDLValue, IDLArgs};
+use dfx_core::fs::create_dir_all;
```
ericswanson-dfinity, line 11: our convention elsewhere has been to group dfx_core imports with `crate` dependencies

ericswanson-dfinity (changes requested):

e2e/tests-dfx/deps.bash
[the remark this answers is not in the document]
lwshang: The following lines
```
assert_command jq -r '.canisters."'"$CANISTER_ID_A"'".arg_raw' deps/init.json
```
check that `init.json` stores the raw argument in `arg_raw` field.

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,270 @@
...
+
+### `wasm_url`
+
+A URL to download canister wasm module which will be deployed locally.
+
+### `wasm_hash` 
+
+SHA256 hash of the wasm module located at `wasm_url`.
+
+This section is optional.
```
ericswanson-dfinity, line 50:
```suggestion
This field is optional.
```

docs/concepts/pull-dependencies.md
```diff
@@ -0,0 +1,270 @@
...
+**Note**
+
+-  `dfx deps pull` connects to the IC mainnet by default (`--network ic`).
+You can choose other network as usual, e.g. `--network local`.
+
+### 3. Set init arguments using `dfx deps init`
+
+Running `dfx deps init` will iterate over all dependencies in `pulled.json`, set an empty argument for the ones that need no init argument and print the list of dependencies that do require an init argument.
+
+Then running `dfx deps init <CANISTER> --argument <ARGUMENT>` to set init argument for individual dependency.
```
ericswanson-dfinity, line 170:
```suggestion
Then running `dfx deps init <CANISTER> --argument <ARGUMENT>` will set the init argument for an individual dependency.
```

src/dfx/src/lib/deps.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    pub dependencies: Vec<Principal>,
+    // the hash on chain
+    // wasm_hash if defined
+    // or get from canister_status
+    pub wasm_hash: String,
+
+    pub init_guide: String,
+    // candid:args
+    pub candid_args: Option<String>,
```
ericswanson-dfinity, line 38:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub candid_args: Option<String>,
```

src/dfx/src/lib/deps.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub canisters: BTreeMap<Principal, PulledCanister>,
+}
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct PulledCanister {
+    // name of `type: pull` in dfx.json. None if indirect dependency.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
+
+    pub dependencies: Vec<Principal>,
```
ericswanson-dfinity, line 30:
```suggestion
    #[serde(skip_serializing_if = "Vec::is_empty")]
    pub dependencies: Vec<Principal>,
```

ericswanson-dfinity (changes requested):

src/dfx/src/lib/deps.rs
lwshang: `candid:args` is a required field. I will turn it into non-optional string.
```rust
pub candid_args: String
```

ericswanson-dfinity (approved):

PR #3096, opened 2023-04-19
feat(extensions): `dfx --help` lists installed extensions

smallstepman:
# Description
- Installed extensions are printed when `dfx help` is issued (or -h or --help). 
  image
- Parses extensions subcommands (and their args) from `extension.json` manifest into `dfx`'s command tree, allowing for calling `help`/`-h`/`--help` on any of the extensions subcommands without actually running them
  image

Closes https://dfinity.atlassian.net/browse/SDK-1088

# How Has This Been Tested?

e2e + unit + manually

# Checklist:


smallstepman:
@ericswanson-dfinity seems like `dfx cache install` is needed to make e2e tests pass. Not sure what's the root issue... let me know if you have any ideas

ericswanson-dfinity (changes requested):

src/dfx/src/lib/error/extension.rs
```diff
@@ -62,8 +62,14 @@ pub enum ExtensionError {
     #[error("Cannot list extensions: {0}")]
     ExtensionsDirectoryIsNotReadable(dfx_core::error::fs::FsError),
 
-    #[error("Malformed extension manifest: {0}")]
-    ExtensionManifestIsNotValidJson(dfx_core::error::structured_file::StructuredFileError),
+    // TODO: this error should be `#[from dfx_core::...::StructureFileError]`, however,
+    // adding new error type to enum StructureFileError causes a cascade of problems,
+    // and they should be addressed separately: https://dfinity.atlassian.net/browse/SDK-1096
+    #[error("Malformed extension manifest: Failed to parse contents of {0} as toml: {1}")]
+    ExtensionManifestIsNotValid(Box<std::path::PathBuf>, toml::de::Error),
```
ericswanson-dfinity, line 69: Box<toml::de::Error) in StructuredFileError didn't work?

src/dfx/src/lib/extension/manager/mod.rs
```diff
@@ -22,20 +16,19 @@ pub struct ExtensionManager {
...
-            get_bin_cache(env.get_version().to_string().as_str()).map_err(|e| {
-                ExtensionError::FindCacheDirectoryFailed(
-                    get_cache_root().unwrap_or_default().join("versions"),
-                    e,
-                )
-            })?;
+    pub fn new(version: &Version) -> Result<Self, ExtensionError> {
+        let versioned_cache_dir = get_bin_cache(version.to_string().as_str()).map_err(|e| {
+            ExtensionError::FindCacheDirectoryFailed(
+                get_cache_root().unwrap_or_default().join("versions"),
```
ericswanson-dfinity, line 22: `version` would be a key piece of information to include here

src/dfx/src/lib/extension/manifest/extension.rs
```diff
@@ -14,15 +21,36 @@ pub struct ExtensionManifest {
...
             return Err(std::fmt::Error)
         };
-        write!(f, "{}", json)
+        write!(f, "{}", s)
+    }
+}
+
+impl ExtensionManifest {
+    pub fn from_extension_directory(path: PathBuf) -> Result<Self, ExtensionError> {
+        let manifest_path = path.join(MANIFEST_FILE_NAME);
```
ericswanson-dfinity, line 39: Given that the only caller is:
```
                let text = match ExtensionManifest::from_extension_directory(
                    self.get_extension_directory(&extension.name),
                ) {
```

Recommend something like this:

```suggestion
    pub fn new(name: String, extensions_root_dir: &Path) -> Result<Self, ExtensionError> {
        let manifest_path = extensions_root_dir.join(name).join(MANIFEST_FILE_NAME);
```
and call with 
```
ExtensionManifest::new(&extension.name, self.dir)
```

src/dfx/src/lib/extension/manifest/extension.rs
```diff
@@ -14,15 +21,36 @@ pub struct ExtensionManifest {
...
+        if !manifest_path.exists() {
+            return Err(ExtensionError::ExtensionManifestMissing(
+                path.components()
+                    .last()
+                    .unwrap() // safe to unwrap - the `path` parameter is guaranteed not to be root (`/`)
+                    .as_os_str()
+                    .to_string_lossy()
+                    .to_string(),
```
ericswanson-dfinity, line 47: The suggestion above will make it easier to get the extension name here.

src/dfx/src/lib/extension/manifest/extension.rs
```diff
@@ -14,15 +21,36 @@ pub struct ExtensionManifest {
...
+            ));
+        }
+        let ext: ExtensionManifestWrapper =
+            toml::from_str(&dfx_core::fs::read_to_string(&manifest_path)?).map_err(|e| {
+                ExtensionError::ExtensionManifestIsNotValid(Box::new(manifest_path), e)
+            })?;
```
ericswanson-dfinity, line 53:
```
dfx_core::load_toml_file(manifest_path).map_err(ExtensionManifestIsNotValid)?;
```

src/dfx/src/lib/error/extension.rs
smallstepman: it generates the warning described in Jira ticket, because `CanisterIdStoreError` imports `UnifiedError` which imports `StructuredFileError`

ericswanson-dfinity: Again, please try again with `Box`.  I have tried that here, and indeed without Box it fails with the warnings you describe, but with Box clippy does not report any warnings.

```
    #[error("Failed to parse contents of {0} as toml: {1}")]
    DeserializeTomlFileFailed(Box<PathBuf>, Box<toml::de::Error>),
```

src/dfx/src/lib/extension/manifest/extension.rs
smallstepman: resolving conversation without changes since we're moving to `json`

src/dfx/src/lib/extension/manifest/extension.rs
```diff
@@ -14,15 +16,33 @@ pub struct ExtensionManifest {
...
+pub struct ExtensionSubcommand {
+    pub subcommands: Vec<ExtensionSubcommand>,
+    pub key: String,
+    pub summary: String,
+}
+
 impl Display for ExtensionManifest {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let Ok(json) = serde_json::to_string_pretty(self) else {
+        let Ok(s) = serde_json::to_string_pretty(self) else {
```
ericswanson-dfinity, line 32: how did you smuggle let Ok(x) else error?  This should be `.map_err()?`.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/extension/manifest/extension.rs
smallstepman: I was surprised too when I found that last week

src/dfx/src/commands/mod.rs
```diff
@@ -71,43 +73,249 @@ pub enum Command {
...
+    fn augment_subcommands(cmd: clap::Command) -> clap::Command {
+        let extension_subcmds = ExtensionManager::new(dfx_version(), false).map_or(vec![], |mgr| {
+            mgr.list_installed_extensions()
+                .unwrap_or_default()
+                .into_iter()
+                .map(|ext| ext.into_clap_command(&mgr))
+                .collect::<Vec<Command>>()
+        });
+
+        let mut cmd = cmd
```
ericswanson-dfinity, line 127: Does this help?
```
fn DfxCommand_with_extensions() -> DfxResult<Command> {

    let cli = Command::new("dfx with extensions")
        .subcommand(
            Command::new("nns").about("about nns")
        )
        .subcommand(
            Command::new("sns").about("about sns"));
    let cli = DfxCommand::augment_subcommands(cli);
    Ok(cli)
}
```

smallstepman: I did something similar but not this exactly.

I tried this approach but couldn't make it work because I couldn't plug the `Command` back into `CliOpts` (because it results in an error: `duplicate command "beta"` (it duplicates all subcommand)). The example in the link parses from an enum, whereas dfx parses from a struct that contains an enum, therefore we cannot just call `parse` on enum, because we won't be able to get top level args (verbose,version,etc).

src/dfx/src/commands/extension/run.rs
```diff
@@ -26,8 +25,6 @@ impl From<Vec<OsString>> for RunOpts {
 
 pub fn exec(env: &dyn Environment, opts: RunOpts) -> DfxResult<()> {
     let mgr = env.new_extension_manager()?;
-    let dfx_version = &env.get_version().to_string();
-    let path_to_dfx_cache = get_bin_cache(dfx_version)?;
-    mgr.run_extension(path_to_dfx_cache, opts.name, opts.params)?;
+    mgr.run_extension(opts.name, opts.params)?;
```
smallstepman, line 28: unnecessary to pass the path to dfx cache, cause it can be derived (ExtensionManager already has a field `dfx_version`)

src/dfx/src/lib/extension/manager/mod.rs
```diff
@@ -21,21 +15,46 @@ pub struct ExtensionManager {
...
 }
 
-impl ExtensionManager {
-    pub fn new(env: &dyn Environment) -> Result<Self, ExtensionError> {
-        let versioned_cache_dir =
-            get_bin_cache(env.get_version().to_string().as_str()).map_err(|e| {
-                ExtensionError::FindCacheDirectoryFailed(
-                    get_cache_root().unwrap_or_default().join("versions"),
-                    e,
+impl CommandFactory for ExtensionManager {
```
smallstepman, line 18: perhaps its unnecessary to impl `clap::CommandFactory`, but I felt like it fit the description:
https://docs.rs/clap/latest/clap/trait.CommandFactory.html
> Create a [Command](https://docs.rs/clap/latest/clap/struct.Command.html) relevant for a user-defined container.

src/dfx/src/main.rs
```diff
@@ -170,7 +174,29 @@ fn print_error_and_diagnosis(err: Error, error_diagnosis: Diagnosis) {
...
+
+    let mut args = std::env::args_os().collect::<Vec<_>>();
+
+    if let Ok(em) = ExtensionManager::new(dfx_version(), false) {
+        // safe to unwrap because clap will display help if no subcommand is provided
+        // which happens when "app.get_matches()" is called above
+        let subcmd = matches.subcommand().unwrap().0;
+        if !DfxCommand::has_subcommand(subcmd) && em.is_extension_installed(subcmd) {
+            args.insert(1, "extension".into());
+            args.insert(2, "run".into());
```
smallstepman, line 195: The benefit of this approach is that we don't need to rewrite anything below (e.g. setting up logging works). The extension shorthand notation will get converted into `DfxCommand::Extension`, which would not be possible otherwise

ericswanson-dfinity (changes requested):
Can we have some e2e tests?  I understand they won't be able to use `dfx extension install`, but they should be able to populate the extensions directory in the cache with some extension.json files and executable shell scripts for the extension "binaries" themselves.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/extension/manifest/extension.rs
```diff
@@ -14,15 +22,177 @@ pub struct ExtensionManifest {
...
+    pub fn new(name: &str, extensions_root_dir: &Path) -> Result<Self, ExtensionError> {
+        let manifest_path = extensions_root_dir.join(name).join(MANIFEST_FILE_NAME);
+        if !manifest_path.exists() {
+            return Err(ExtensionError::ExtensionManifestMissing(
+                name.to_owned(),
+                manifest_path,
+            ));
+        }
+        let mut m: ExtensionManifest = dfx_core::json::load_json_file(&manifest_path)
+            .map_err(ExtensionError::ExtensionManifestIsNotValid)?;
```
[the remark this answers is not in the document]
smallstepman: right, makes sense

src/dfx/src/main.rs
```diff
@@ -170,12 +174,35 @@ fn print_error_and_diagnosis(err: Error, error_diagnosis: Diagnosis) {
 }
 
 fn main() {
-    let cli_opts = CliOpts::parse();
+    let mut args = std::env::args_os().collect::<Vec<OsString>>();
+    let mut error_diagnosis: Diagnosis = NULL_DIAGNOSIS;
+
+    if let Ok(em) = ExtensionManager::new(dfx_version(), false) {
```
smallstepman, line 180: @ericswanson-dfinity this line makes `e2e/tests-dfx/usage_env.bash` fail on this line https://github.com/dfinity/sdk/blob/41dc038704bc24cd2deebdb77365602a86148d0b/e2e/tests-dfx/usage_env.bash#L26
because `ExtensionManager::new` calls `pub fn get_bin_cache(v: &str)` which calls `get_cache_root` & `get_bin_cache_root` and they always create cache dir if it doesn't exist
https://github.com/dfinity/sdk/blob/41dc038704bc24cd2deebdb77365602a86148d0b/src/dfx-core/src/config/cache.rs#L34-L35
https://github.com/dfinity/sdk/blob/41dc038704bc24cd2deebdb77365602a86148d0b/src/dfx-core/src/config/cache.rs#L47-L48

Could you please advise how to solve it? I've been thinking about these options:
- extend function signatures for `fn get_bin_cache/get_cache_root/get_bin_cache_root` with param `initialize: bool` which when set to `false` guards the calls to `crate::fs::create_dir_all` 
- create additional function `fn get_bin_cache_unitialized() -> PathBuf`
- create additional function `fn is_cache_installed(v: Version) -> bool`, and make `ExtensionManager::new` return `Err` if cache not present

ericswanson-dfinity: If I understand it, `dfx identity new` creates the empty cache dir (as will any dfx command), rather than `dfx new`.

The purpose of this test is to show that DFX_CONFIG_ROOT and DFX_CACHE_ROOT control the locations of identities and the cache, respectively.

The failing line in question is verifying a known initial state, in order to prove that `dfx new` creates the cache directory under DFX_CACHE_ROOT and does not create it under HOME.

I would suggest deleting the cache directory before that failing line, either with `dfx cache delete` or through other means.

smallstepman: > dfx identity new creates the empty cache dir (as will any dfx command), rather than dfx new.

Yes, exactly. I thought this is not ok? with the current setup even `dfx --help` would create an empty cache dir

ericswanson-dfinity: Sure, how about see if you can find a way to skip all of this if there's no cache (or no extensions directory in the cache).

```
    if let Ok(em) = ExtensionManager::new(dfx_version(), false) {
        match em.installed_extensions_as_clap_commands() {
```

src/dfx-core/src/config/cache.rs
```diff
@@ -41,10 +41,10 @@ pub fn get_cache_root() -> Result<PathBuf, CacheError> {
 
 /// Return the binary cache root. It constructs it if not present
 /// already.
-pub fn get_bin_cache_root() -> Result<PathBuf, CacheError> {
-    let p = get_cache_root()?.join("versions");
+pub fn get_bin_cache_root(ensure_dir_exists: bool) -> Result<PathBuf, CacheError> {
+    let p = get_cache_root(ensure_dir_exists)?.join("versions");
 
-    if !p.exists() {
+    if !p.exists() && ensure_dir_exists {
```
smallstepman, line 47: @ericswanson-dfinity wdyt about this approach?

src/dfx-core/src/config/cache.rs
```diff
@@ -17,7 +17,7 @@ pub trait Cache {
     fn get_binary_command(&self, binary_name: &str) -> Result<std::process::Command, CacheError>;
 }
 
-pub fn get_cache_root() -> Result<PathBuf, CacheError> {
+pub fn get_cache_root(ensure_dir_exists: bool) -> Result<PathBuf, CacheError> {
```
ericswanson-dfinity, line 20: This approach doesn't work for at least these reasons:
- booleans aren't the most readable type to pass around as parameters
- this particular boolean is a lie at different levels:
  - get_cache_root(ensure_dir_exists = true): ensures `<cache root>` exists, returns `<cache root> ...  ok
  - get_bin_cache_root(ensure_dir_exists = true): ensures `<cache root>/versions` exists, returns `<cache root>/versions` .. ok
  - get_bin_cache(version: &str, ensure_dir_exists) = true: ensures `<cache root>/versions` exists, returns `<cache root>/versions/<dfx version>` .. oops

Luckily, `get_cache_root` only has two callers, and one of them adds two levels of subdirectory.

Let's try this:
- remove all these `ensure_dir_exists : bool` parameters
- change `get_cache_root` to construct and return the path without creating any directories
- add a new method 'get_cache_path_for_version(v: &str)` that constructs and returns `<cache root>/versions/<version>` without creating any directories
- use `get_cache_path_for_version(v: &str)` to determine if there are extensions, without having to create the cache directory.

smallstepman: very nice

src/dfx/src/lib/deps/mod.rs
```diff
@@ -260,7 +260,11 @@ pub fn get_pulled_service_candid_path(canister_id: &Principal) -> DfxResult<Path
 
 /// The path of the dir contains wasm and service.did.
 pub fn get_pulled_canister_dir(canister_id: &Principal) -> DfxResult<PathBuf> {
-    Ok(get_cache_root()?.join("pulled").join(canister_id.to_text()))
+    let p = get_cache_root()?;
+    if !p.exists() {
+        dfx_core::fs::create_dir_all(&p)?;
+    }
+    Ok(p.join("pulled").join(canister_id.to_text()))
```
ericswanson-dfinity, line 267: Given this method returns `some/dir/pulled/<canister id>/`, I'm not sure there is value in creating `some/dir/`.  The reason is that something else must be creating those subdirectories, and will have to create the parent directories if they don't exist.

```suggestion
    let p = get_cache_root()?;
    Ok(p.join("pulled").join(canister_id.to_text()))
```

That said:
- IF this method does create directories, it should create all of them.
- `if !p.exists() { create dir.. }` is known as `ensure_dir_exists(p)`

src/dfx/src/main.rs
ericswanson-dfinity, line 180: Getting close!  This shouldn't swallow errors.  You could have `ExtensionManager::new()` return `Result<Option<..>>` or use some other means.

ericswanson-dfinity (changes requested):

src/dfx/src/main.rs
ericswanson-dfinity, line 180: We shouldn't hide (all) errors here. In this case, it's not an error if the extensions directory doesn't exist.

src/dfx/src/lib/deps/mod.rs
smallstepman: wasn't sure here - I just wanted to mimic old behavior

src/dfx/src/lib/error/extension.rs
```diff
@@ -62,13 +68,16 @@ pub enum ExtensionError {
     #[error("Cannot list extensions: {0}")]
     ExtensionsDirectoryIsNotReadable(dfx_core::error::fs::FsError),
 
-    #[error("Malformed extension manifest: {0}")]
-    ExtensionManifestIsNotValidJson(dfx_core::error::structured_file::StructuredFileError),
+    #[error("Cannot load extension manifest: {0}")]
+    ExtensionManifestIsNotValid(dfx_core::error::structured_file::StructuredFileError),
```
ericswanson-dfinity, line 72:
```suggestion
    LoadExtensionManifestFailed(dfx_core::error::structured_file::StructuredFileError),
```

ericswanson-dfinity (approved):
Just one minor thing and gtg!

PR #3115, opened 2023-05-03
chore: Update to clap v4

adamspofford-dfinity:
This updates clap to v4. Notable changes:

- Argument IDs are no longer case-converted in code; `foo_bar` means `conflicts_with("foo_bar")`, not `"foo-bar"`.
- 'Validators' have been replaced with parsers, to eliminate our habit of parsing twice.
- ArgAction replaces a number of special-cased functions like from_occurrences. 
- `#[clap]` has been replaced with `#[arg]` and `#[command]`.

The full high-level changelog can be found [here](https://epage.github.io/blog/2022/09/clap4/). 

This PR would have been delayed longer until `unstable-styles` is no longer unstable, but v3 blocks work by @smallstepman, so we should instead just be careful about updating clap in the lockfile. Also, these are more or less the default colors from v3, but we might want to look into using brand colors in certain places in a future PR.


src/dfx/src/commands/beta/project/import.rs
```diff
@@ -16,7 +16,7 @@ pub struct ImportOpts {
     canister_name: Option<String>,
 
     /// Imports all canisters found in the other project.
-    #[clap(long, required_unless_present("canister-name"))]
+    #[clap(long, required_unless_present("canister_name"))]
```
smallstepman, line 19:
```suggestion
    #[arg(long, required_unless_present("canister_name"))]
```

smallstepman (commented):
shouldn't we use `arg` instead of `clap`? https://docs.rs/clap/latest/clap/index.html

smallstepman:
since we're using `clap` as a workspace dependency, I think we should also update `icx-asset` (`#[clap` -> `#[arg/command`)

adamspofford-dfinity:
It will be breaking in v5, but they're not emitting deprecation warnings for it.

ericswanson-dfinity (approved):

src/canisters/frontend/icx-asset/Cargo.toml
```diff
@@ -12,9 +12,10 @@ categories = ["command-line-interface"]
 keywords = ["internet-computer", "agent", "icp", "dfinity", "asset"]
 
 [dependencies]
+anstyle.workspace = true
 anyhow.workspace = true
 candid = { workspace = true }
-clap = { workspace = true, features = ["derive", "cargo"] }
+clap = { workspace = true, features = ["derive", "cargo", "unstable-styles"] }
```
smallstepman, line 18:
```suggestion
clap = { workspace = true, features = ["derive", "cargo", "unstable-styles", "wrap_help" ] }
```

smallstepman (changes requested):
looks like text wrapping for help command is behind feature flag now, you probably didn't notice on your widescreen ;) 
Screenshot 2023-05-04 at 17 30 03

smallstepman (approved):

PR #3118, opened 2023-05-09
feat: add option to provide a custom build command to asset canister

nathanosdev:
# Description

This PR adds the ability to provide a custom command when building assets canisters. If no command is provided, it will fallback to the current `npm run build` command.

# How Has This Been Tested?

I have tested this in my own project over the last week. I can also add e2e tests, but would like to gather feedback from the SDK team on this first since it's a new feature.

# Checklist:

- [ ] I have added e2e tests.

### DIFF
--- a/src/dfx/src/lib/builders/custom.rs
+++ b/src/dfx/src/lib/builders/custom.rs
@@ -4,7 +4,7 @@ use crate::lib::builders::{
 use crate::lib::canister_info::custom::CustomCanisterInfo;
 use crate::lib::canister_info::CanisterInfo;
 use crate::lib::environment::Environment;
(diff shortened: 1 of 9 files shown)


src/dfx/src/lib/builders/custom.rs
```diff
@@ -207,36 +206,6 @@ impl CanisterBuilder for CustomBuilder {
     }
 }
 
-fn run_command(args: Vec<String>, vars: &[super::Env<'_>], cwd: &Path) -> DfxResult<()> {
```
nathanosdev, line 210: Moved to `mod.rs` so it can be reused

PR #3119, opened 2023-05-09
fix: prevent using --argument with --all in canister installation

adamspofford-dfinity:
Currently `dfx deploy --argument` provides the same argument to all canisters, and `dfx canister install --all --argument` provides an empty argument regardless of what was passed. This PR makes both commands an error. Fixes SDK-1008.


e2e/tests-replica/deploy.bash
```diff
@@ -69,7 +69,7 @@ teardown() {
     dfx_start
     assert_command dfx canister create --all
 
-    assert_command dfx deploy --argument '("World")'
+    assert_command dfx deploy hello_backend --argument '("World")'
```
ericswanson-dfinity, line 72: Ideally we'd have some test that demonstrates commands that would have been accepted but now are rejected, both for `dfx deploy` and `dfx canister install`.  Put another way, if a future PR reverted the changes to *.rs files in this PR, no tests would fail.
```suggestion
    assert_command_fail dfx deploy --argument '("World")'
    assert_command dfx deploy hello_backend --argument '("World")'
```

ericswanson-dfinity (changes requested):

e2e/tests-dfx/deploy.bash
```diff
@@ -102,3 +103,11 @@ teardown() {
     assert_command dfx identity get-wallet
     assert_contains "Creating a wallet canister"
 }
+
+@test "deploying multiple canisters with arguments fails" {
+    dfx_start
```
ericswanson-dfinity, line 108: No need to start dfx
```suggestion
```

e2e/tests-dfx/install.bash
```diff
@@ -181,3 +181,11 @@ teardown() {
     assert_command dfx canister info e2e_project_frontend
     assert_contains db07e7e24f6f8ddf53c33a610713259a7c1eb71c270b819ebd311e2d223267f0
 }
+
+@test "installing multiple canisters with arguments fails" {
+    dfx_start
+    dfx canister create --all
+    dfx build
```
ericswanson-dfinity, line 188: No need to start or build
```suggestion
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #3122, opened 2023-05-10
feat: .env file includes all created canister IDs

sesi200:
# Description

Previously the `.env` file only included canister IDs for canisters that were listed as explicit dependencies during the build process.
Now all canisters that have a canister ID for the specified network are included in `.env`.

Fixes [SDK-1050](https://dfinity.atlassian.net/browse/SDK-1050), issue raised in [this forum thread](https://forum.dfinity.org/t/dfx-env-incorrect-canister-ids-generated/19057)

# How Has This Been Tested?

Added e2e

# Checklist:

[SDK-1050]: https://dfinity.atlassian.net/browse/SDK-1050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e/tests-dfx/network.bash
```diff
@@ -81,6 +81,6 @@ teardown() {
 
     setup_actuallylocal_shared_network
 
-    assert_command_fail dfx build --network actuallylocal
+    assert_command_fail dfx build e2e_project_frontend --network actuallylocal
```
sesi200, line 84: Not necessarily required, but I ran into problems while debugging with this not being explicit. The error message below refers explicitly to `e2e_project_frontend`, so if `[...]_backend` is built first, then the error message doesn't match even though nothing really went wrong

src/dfx-core/src/fs/mod.rs
```diff
@@ -18,6 +18,7 @@ pub fn canonicalize(path: &Path) -> Result<PathBuf, FsError> {
 }
 
 pub fn copy(from: &Path, to: &Path) -> Result<u64, FsError> {
+    create_dir_all(&parent(to)?)?;
```
ericswanson-dfinity, line 21: The methods here are meant as simple wrappers for methods in std::fs.  Methods that do more would go in the `composite` module and have a different name.

Also, if there are only certain places where the target directory might not exist, I'd rather call that out with an explicit call to `dfx_core::fs::composite::ensure_parent_dir_exists`.

src/dfx/src/lib/models/canister.rs
```diff
@@ -250,15 +250,15 @@ impl CanisterPool {
...
+    pub fn load<'a, I>(
         env: &dyn Environment,
         generate_cid: bool,
-        canister_names: &[String],
-    ) -> DfxResult<Self> {
+        canister_names: I,
+    ) -> DfxResult<Self>
+    where
+        I: Iterator<Item = &'a String>,
+    {
```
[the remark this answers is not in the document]
ericswanson-dfinity: I would prefer to go back to passing `&[String]` if we can, since all the `.iter()` parameters at call points make me wonder why that is the case

ericswanson-dfinity (changes requested):

e2e/tests-dfx/network.bash
```diff
@@ -82,5 +83,6 @@ teardown() {
     setup_actuallylocal_shared_network
 
     assert_command_fail dfx build --network actuallylocal
-    assert_match "Cannot find canister id. Please issue 'dfx canister create e2e_project_frontend --network actuallylocal"
+    assert_match "Cannot find canister id."
+    assert_match "--network actuallylocal"
```
sesi200, line 87: The order of canister building is non-deterministic since it is sourced from a `HashSet`. This caused this test to fail on CI even though it passes locally.

The changes preserve what the test is supposed to test, but stops checking for the order in which canisters are checked

e2e/tests-dfx/network.bash
```diff
@@ -73,7 +73,8 @@ teardown() {
 @test "failure message does not include network if for local network" {
     dfx_start
     assert_command_fail dfx build --network local
-    assert_match "Cannot find canister id. Please issue 'dfx canister create e2e_project_frontend"
+    assert_match "Cannot find canister id."
+    assert_not_match "local"
```
ericswanson-dfinity, line 77:
```suggestion
    assert_not_contains "--network local"
```

ericswanson-dfinity (approved):

PR #3123, opened 2023-05-10
feat: gzip option

lwshang:
# Description

Fix SDK-1102

```json
{
  "canisters" : {
    "app" : {
      "gzip" : true
    }
  }
}
```
Then `dfx` will gzip wasm module as the final step in building canisters. 

Please note that, the `wasm` field of a custom type canister must point to a non-gzip wasm file. 

And the `optimize` and `shrink` behavior is unified for all canister types. Now `shrink` is on by default.

# How Has This Been Tested?

# Checklist:


src/dfx-core/src/config/model/dfinity.rs
```diff
@@ -234,13 +234,14 @@ pub struct ConfigCanistersCanister {
 
     /// # Shrink Canister WASM
     /// Whether run `ic-wasm shrink` after building the Canister.
-    /// Enabled by default for Rust/Motoko canisters.
-    /// Disabled by default for custom canisters.
+    /// If not specify `optimize`, this option will be true by default.
+    /// To turn off all optimization/shrink, please explicitly set this option to false.
```
ericswanson-dfinity, line 238: We still shouldn't alter the wasm by default for custom canisters (including shrink).  We've gotten plenty of feedback about this.  Did something change?

ericswanson-dfinity (changes requested):

src/dfx-core/src/config/model/dfinity.rs
lwshang: My main purpose is to make the behavior consistent for all types of canisters.
When I introduced `shrink`, I make it disable by default for custom canisters because the wasm may be gzipped.
In this PR, I force the custom canister to be non-gzipped. So the `shrink` is always viable.

I'm aware that developer may want to install the same canister specified in `dfx.json` without any modification. So here is a tradeoff, consistency v.s. no modification on custom canister wasm.

Another option is to make `shrink` disable by default for all canister types.

I unified the `optimize/shrink` code from builders into `step_postbuild`. I want to avoid adding another `if ` statement there.

Actually, I want to deprecate `shrink` and use `optimize` only. Then it will be no `optimize/shrink` by default. And devs can opt in `optimize`.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/models/canister.rs
```diff
@@ -197,10 +210,31 @@ impl Canister {
             remove_metadata(&mut m, name);
 
             add_metadata(&mut m, visibility, name, data);
+            modified = true;
+        }
+
+        // If not modified and not set "gzip" explicitly, copy the wasm file directly so that hash match.
+        if !modified && !info.get_gzip() {
+            dfx_core::fs::copy(build_output_wasm_path, &wasm_path)?;
```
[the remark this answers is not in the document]
lwshang: The deleted logic does following things:
1. copy the wasm file from `build_output.wasm` (file1) to `canister.info.get_build_wasm_path()` (file2) if these two paths are not the same;
2. `apply_metadata` to file2

The new implementation does following instead:
1. read file1 as wasm module
2. do the post-process on the wasm module
3. gzip if need
4. write the wasm (maybe gzipped) to file2

The e2e tests in `build.bash`, `install.bash` and `deploy.bash` make sure that the wasm file are generated at the correct path (file2) so that the canister can be installed.

src/dfx/src/lib/models/canister.rs
```diff
@@ -102,13 +103,50 @@ impl Canister {
...
+        };
+        let wasm_path = self.info.get_build_wasm_path();
+        dfx_core::fs::composite::ensure_parent_dir_exists(&wasm_path)?;
+        let info = &self.info;
+        if info.is_pull() || info.is_remote() {
+            // We won't install such canisters with `dfx canister install` or `dfx deploy`.
+            // So no need to copy the wasm.
+            // Pull type canisters have empty `build_output_wasm_path`. Impossible to copy it.
+            return Ok(());
+        }
```
ericswanson-dfinity, line 126: Do you want to consolidate this check with the one above?  IIUC, `info.is_pull()` can never be true here, because build_output.wasm will always be WasmBuildOutput::None.  And `Pull type canisters have empty build_output_wasm_path` is no longer accurate.

src/dfx/src/lib/models/canister.rs
```diff
@@ -169,8 +187,8 @@ impl Canister {
 
             let data = match (section.path.as_ref(), section.content.as_ref()) {
                 (None, None) if section.name == CANDID_SERVICE =>
-                    std::fs::read(&idl_path)
-                .with_context(|| format!("Failed to read {}", idl_path.to_string_lossy()))?
+                    std::fs::read(build_idl_path)
+                .with_context(|| format!("Failed to read {}", build_idl_path.to_string_lossy()))?
```
ericswanson-dfinity, line 191:
```suggestion
                    dfx_core::fs::read(build_idl_path)?
```

ericswanson-dfinity (approved):

PR #3126, opened 2023-05-12
feat!: Ask for user consent when removing themselves as principal

sesi200:
# Description

Removing oneself (or the wallet one uses) can result in the loss of control over a canister.
Therefore `dfx canister update-settings` now asks for extra confirmation when removing the currently used principal/wallet from the list of controllers.
To skip this check in CI, use either the `--yes`/`-y` argument or use `echo "yes" | dfx canister update-settings <...>`.

Fixes [SDK-1049](https://dfinity.atlassian.net/browse/SDK-1049)

# How Has This Been Tested?

Added in e2e.

There was no test at all for `--remove-controller` so this test is added from scratch.

# Checklist:

[SDK-1049]: https://dfinity.atlassian.net/browse/SDK-1049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/dfx/src/commands/canister/update_settings.rs
```diff
@@ -222,6 +232,31 @@ pub async fn exec(
...
+        .remove_controller
+        .as_ref()
+        .map(|remove| remove.contains(&caller_principal))
+        .unwrap_or_default();
+    let not_in_new_controllers = opts
+        .set_controller
+        .as_ref()
+        .map(|set| !set.contains(&caller_principal))
+        .unwrap_or_default();
+    Ok(removes_themselves || not_in_new_controllers)
```
ericswanson-dfinity, line 257: I had a little trouble following what was going on here, I think due to the combination of `Option<Vec<_>>`, (sometimes )inverted membership in `Vec<_>`, and `unwrap_or_default`.  But both checks only have a single case in which they are `true`.

What would you think about this?
```suggestion
    let removes_themselves =
        matches!(&opts.remove_controller, Some(remove) if remove.contains(&caller_principal));
    let sets_without_themselves =
        matches!(&opts.set_controller, Some(set) if !set.contains(&caller_principal));
    Ok(removes_themselves || sets_without_themselves)
```

ericswanson-dfinity (changes requested):

e2e/tests-dfx/update_settings.bash
```diff
@@ -107,23 +112,23 @@ teardown() {
...
     assert_match "Set controller of \"${ID}\" to: alice"
 
     # Set controller using invalid principal/identity fails
-    assert_command_fail dfx canister update-settings hello_backend --set-controller charlie --identity alice --wallet "${ALICE_WALLET}"
+    assert_command_fail dfx canister update-settings hello_backend --set-controller charlie --identity alice --wallet "${ALICE_WALLET}" --yes
     assert_match "Identity charlie does not exist"
 
     # Set controller using invalid canister name/id fails
-    assert_command_fail dfx canister update-settings hello_assets --set-controller bob --identity alice --wallet "${ALICE_WALLET}"
+    assert_command_fail dfx canister update-settings hello_assets --set-controller bob --identity alice --wallet "${ALICE_WALLET}" --yes
```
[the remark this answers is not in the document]
sesi200: Yes, makes sense. Also added it to the other ones for consistency

src/dfx/src/commands/canister/update_settings.rs
sesi200: I wasn't aware of `matches!` with `if`. Nice idea!

ericswanson-dfinity (changes requested):
Just a typo left

PR #3135, opened 2023-05-19
feat: auto-stop canisters on delete

adamspofford-dfinity:
Currently, `dfx canister delete` will exit with an error if the canister has not been stopped yet. This PR makes it so that it will confirm deletion with the user instead. (It is intentionally not left friction-free because deletion is destructive; `dfx canister stop` can be viewed as pre-confirming for a particular canister.)

Fixes SDK-893.


e2e/tests-dfx/delete.bash
```diff
@@ -0,0 +1,34 @@
...
+    standard_teardown
+}
+
+@test "delete can be used to delete a canister" {
+    dfx_start
+    dfx deploy e2e_project_backend
+    dfx canister stop e2e_project_backend
+    assert_command dfx canister delete e2e_project_backend
+    assert_command_fail dfx canister info e2e_project_backend
+    assert_contains "Cannot find canister id. Please issue 'dfx canister create e2e_project_backend'."
```
sesi200, line 23: This only tests if dfx locally saved the ID. Can you please also test if the canister does not exist anymore on the replica?

sesi200 (approved):

PR #3136, opened 2023-05-23
fix: dfx start respects replica port config

ericswanson-dfinity:
# Description

If you put something like this in networks.json, `dfx start` will use the configured replica port rather than assigning one randomly.

```
$ cat ~/.config/dfx/networks.json 
{
  "local": {
    "bind": "127.0.0.1:8080",
    "type": "ephemeral",
    "replica": {
      "port": 8123,
      "log_level": "info",
      "subnet_type": "system"
    }
  }
}
```

Fixes https://dfinity.atlassian.net/browse/SDK-1122

# How Has This Been Tested?

Added an e2e test

# Checklist:


lwshang:
When the specified port is occupied, what will the error message look like?
We'd better direct the user to the config file(s) to change the port.

PR #3161, opened 2023-06-07
refactor(frontend-canister): Asset is_aliased -> enable_aliasing

ericswanson-dfinity:
# Description

Changes the name of the Asset field `is_aliased` to `enable_aliasing`.  Note that this does not affect the external interface.

This reflects the actual meaning of the field, which may be `true` even for assets that are not actually aliased.

The motivation is that I am writing an interface document for the asset canister, and the misnaming of this field makes parts of that more difficult.  Also, while in the `CreateAssetArguments` the field is correctly called `enable_aliasing`, the `is_aliased` naming has already propagated to the `set_asset_properties` and `get_asset_properties` methods. Therefore a further motivation is to discourage further propagation and to consider updating those methods in a backwards-compatible way.

# How Has This Been Tested?

Covered by existing tests.

# Checklist:

- [x] There was no need for corresponding changes to the documentation.


ericswanson-dfinity:
Unfortunately this will/would have to deal with upgrades

PR #3169, opened 2023-06-12
fix(extensions): ensure dfx extensions locate dfx cache accurately 

smallstepman:
# Description

Some `dfx` extensions depend on binaries from `dfx`'s cache. Previously, extensions depended on the first `dfx` in `$PATH` to locate its cache, which is error-prone for projects that declare a custom `dfx` version in their `dfx.json`. 

Fixes: https://dfinity.atlassian.net/browse/SDK-1107
Depends on: https://github.com/dfinity/dfx-extensions/pull/21

# How Has This Been Tested?

manually, by installing sns and nns extensions and executing commands which depend on binaries from cache  

# Checklist:

### DIFF
--- a/src/dfx/src/commands/extension/run.rs
+++ b/src/dfx/src/commands/extension/run.rs
@@ -27,7 +27,7 @@ impl From<Vec<OsString>> for RunOpts {
 pub fn exec(env: &dyn Environment, opts: RunOpts) -> DfxResult<()> {
     let mgr = env.new_extension_manager()?;
     let dfx_version = &env.get_version().to_string();
-    let path_to_dfx_cache = get_bin_cache(&dfx_version)?;
+    let path_to_dfx_cache = get_bin_cache(dfx_version)?;
     mgr.run_extension(&path_to_dfx_cache, opts.name, opts.params)?;
     Ok(())
 }
--- a/src/dfx/src/commands/extension/run.rs
+++ b/src/dfx/src/commands/extension/run.rs
@@ -4,16 +4,18 @@ use crate::lib::error::DfxResult;
 use clap::Parser;
 use dfx_core::config::cache::get_bin_cache;
 
+use std::ffi::OsString;
+
 #[derive(Parser, Debug)]
 pub struct RunOpts {
     /// Specifies the name of the extension to run.
-    name: String,
+    name: OsString,
     /// Specifies the parameters to pass to the extension.
-    params: Vec<String>,
+    params: Vec<OsString>,
 }
 
-impl From<Vec<String>> for RunOpts {
-    fn from(value: Vec<String>) -> Self {
+impl From<Vec<OsString>> for RunOpts {
+    fn from(value: Vec<OsString>) -> Self {
         let (extension_name, params) = value.split_first().unwrap();
         RunOpts {
             name: extension_name.clone(),
@@ -26,6 +28,6 @@ pub fn exec(env: &dyn Environment, opts: RunOpts) -> DfxResult<()> {
     let mgr = env.new_extension_manager()?;
     let dfx_version = &env.get_version().to_string();
     let path_to_dfx_cache = get_bin_cache(dfx_version)?;
-    mgr.run_extension(&path_to_dfx_cache, opts.name, opts.params)?;
+    mgr.run_extension(path_to_dfx_cache, opts.name, opts.params)?;
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


ericswanson-dfinity (changes requested):

src/dfx/src/commands/extension/run.rs
```diff
@@ -25,6 +26,12 @@ impl From<Vec<OsString>> for RunOpts {
 
 pub fn exec(env: &dyn Environment, opts: RunOpts) -> DfxResult<()> {
     let mgr = env.new_extension_manager()?;
-    mgr.run_extension(env.get_version(), opts.name, opts.params)?;
+    let dfx_version = &env.get_version().to_string();
+    let path_to_dfx_cache = get_bin_cache(&dfx_version).unwrap();
+    mgr.run_extension(
+        &*path_to_dfx_cache.to_string_lossy(),
```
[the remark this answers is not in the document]
smallstepman: > This is a sign of something wrong.

why do you think so? `to_string_lossy` returns `Cow<str>`. To me `*&` seems to be the way to get `Cow<str> -> &str`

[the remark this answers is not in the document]
ericswanson-dfinity: Using `.to_string_lossy()` to turn a `PathBuf` into a string that will be used to access that path.  "something wrong": the "lossy" part, but also passing a PathBuf as a &str, not a &Path.  I see you already corrected that here, nice!

ericswanson-dfinity (changes requested):

src/dfx/src/lib/extension/manager/execute.rs
```diff
@@ -1,19 +1,21 @@
...
     ) -> Result<(), ExtensionError> {
-        let extension_name = extension_name
-            .into_string()
-            .map_err(ExtensionError::InvalidExtensionName)?;
-
         let mut extension_binary = self.get_extension_binary(&extension_name)?;
 
+        params.extend([
+            "--dfx-cache-path".into(),
+            dfx_cache.to_string_lossy().into(),
```
ericswanson-dfinity, line 16: Killin' me with `to_string_lossy` on a path.  Is it really necessary?

smallstepman: to me, it feels like a way to go because `Path` is backed by `OsString`. The only methods available to turn it into a string are:
- `display` https://doc.rust-lang.org/std/path/struct.Display.html
  > This may perform lossy conversion, depending on the platform
- `as_os_str` & `as_mut_os_str` - pointless here, just kicking the can down the road
- `to_str` which yields None in cases where `display` and `to_string_lossy` would perform lossy conversion
- [to_string_lossy](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy)
- `Debug` (`{:?}`), which seems to be the only way to perform a lossless conversion

If you insist on staying away from lossy conversion, I can revert https://github.com/dfinity/sdk/pull/3169/commits/1d5197f816c3ab7bcfeb7c7d900a9122cd648590 and we'll use OsStrings everywhere, which I think is also fine

ericswanson-dfinity: As mentioned above, `PathBuf` (not &Path) can convert directly to OsString with `PathBuf.into_os_string()`.  I leave it to you to decide if you'd rather pass PathBuf or OsString.

smallstepman: but we're not using OsString anymore

params is now `Vec<String>`, therefore we need to convert from PathBuf to String

I honestly think sticking with `params: Vec<OsString>` will be better here

ericswanson-dfinity: There was no request to change params from Vec<OsString>, only the extension name, because it was then being converted to String.

smallstepman: how about we turn all instances of extension_name to OsStr/OsString instead? https://sourcegraph.com/search?patternType=regexp&case=yes&q=context%3Aglobal+repo%3A%5Egithub%5C.com%2Fdfinity.%2B%24+extension_name&groupBy=repo

the name is never generated by the dfx, it always comes from the outside world (either passed as a string in CLI, or as a result of `listdir`)

ericswanson-dfinity: `String` seems like a better fit for `extension_name`.  Conceptually it's a command name, like `start` or `canister` or `nns`, even though we happen to be reading it from a list of directories.

PR #3206, opened 2023-06-29
fix: Ctrl-C right after dfx start panics in 2 minutes

lwshang:
Fixes SDK-972

# Description

With `dfx` v0.14.1, sending `Ctrl-C` right after `dfx start` will result in:
```
Running dfx start for version 0.14.1
Using shared network 'local' defined in /Users/lwshang/.config/dfx/networks.json
Stopping canister http adapter...
Stopped.
Stopping the replica...
```
Then it hangs and after 2 minutes:
```
thread 'replica-actor' panicked at 'called `Result::unwrap()` on an `Err` value: Cannot start the replica: timed out', src/dfx/src/actors/replica.rs:373:78
```

If `dfx start --background` and `Ctrl-C` immediately, the process hangs in the background. And if run `dfx start` during this 2 minutes window, it will error with "Address in use".

## Cause

The reason of this bug was in the long-running replica actor start procedure.
It wait for a port file to be ready. But the `ic-starter` process has been terminated.
So it hangs until all retry effort (retry every 0.1s for 1200 times) in 2 minutes.

## Fix

This PR fix the issue by breaking out of the retry loop when receive the stop signal.

Now with the same commands, it will get following output immediately.

```
Stopping the replica...
Stopped
```

# How Has This Been Tested?

e2e test with an `exp` script.

# Checklist:


ericswanson-dfinity (commented):
This is great!  I ran this locally to test, and while it didn't get stopped by the replica, it did wait and panic due to similar issues with the canister http adapter.  What would you think about giving similar treatment to the canister http adapter and btc adapter?

```
$ dfx-wip start
Running dfx start for version 0.14.2-beta.1+rev9.2329da59
Using shared network 'local' defined in /Users/ericswanson/.config/dfx/networks.json
^CStopping canister http adapter...
thread 'canister-http-adapter-actor' panicked at 'canister http adapter socket was not created: Cannot start canister-http-adapter: timed out', src/dfx/src/actors/canister_http_adapter.rs:205:22
Stopped.
Stopping the replica...
Stopped.
```

lwshang:
Sure. I will apply the fix to both actors.

ericswanson-dfinity:
Also, it seems like this test doesn't fail if the condition occurs.

I tried adding "ps" at the end of the test, and running the following with a `dfx` built from master.

```bash
$ repeat-bats -f panic e2e/tests-dfx/start.bash --show-output-of-passing-tests
```

where `repeat-bats` is

```bash
$  cat ~/bin/repeat-bats
#!/usr/bin/env bash

set -e

for a in $(seq 1 10000); do
    set-title "--------- $a ----------"
    echo "*** Iteration $a"

    bats "$@"

done
```

Sometimes one of the tests pauses for about 2 minutes, and the output for the test includes something like this
```
Error: Failed to kill all processes.  Remaining: 65079
```

where `ps` shows
```
 65079 ttys006    0:00.09 dfx start --clean
```

Can you do something to the test so that it can detect the panic condition when it happens?

In other words, what I'd like to see is:
- if I run the test from this PR against a `dfx` binary build from `master`, I can see a test failure after some number of iterations
- if I run the same test from this PR against a dfx binary built from the PR branch, I don't see any test failures after some number of iterations

e2e/assets/expect_scripts/ctrl_c_right_after_dfx_stop.exp
```diff
@@ -3,14 +3,12 @@
...
-spawn dfx start --clean --enable-bitcoin
+spawn dfx start --clean
 sleep 2
 send \x03
 sleep 1
 
-expect "Stopping btc-adapter...
-Stopped.
-Stopping canister http adapter...
+expect "Stopping canister http adapter...
```
ericswanson-dfinity, line 11: These can happen in either order

lwshang:
There is a weird bug in this PR. It can't catch my latest commit. Will file a new PR.

PR #3207, opened 2023-06-29
feat(github-ci): add GitHub Action to install dfx

mmicu:
# Description

Add a GitHub Action that installs `dfx` in GitHub Action Workflow.

# How Has This Been Tested?

`act` can be used to test this new action, together with `.github/workflows/tests/dfx-action.yml`:

```
$ act -W .github/workflows/tests/dfx-action.yml
```

# Checklist:


action.yml
```diff
@@ -0,0 +1,16 @@
...
+      the DFX_VERSION variable that will be handled by the installation
+      script "install.sh".
+    default: '0.14.1'
+
+runs:
+  using: composite
+  steps:
+    - name: Install dfx
+      shell: sh
+      run: DFX_VERSION=${{ inputs.dfx-version }} sh -ci "$(curl -fsSL https://internetcomputer.org/install.sh)"
```
smallstepman, line 16: Could you please tweak it a little bit? I just realized that in the current form, we would have to bump the `default: ...` value whenever we create a new dfx release. Instead, let's set `default: ` to `'latest'` and do this instead
```suggestion
      run: |
        if [[ "${{ inputs.dfx-version }}" == "latest" ]]; then
          sh -c "$(curl -fsSL https://internetcomputer.org/install.sh)"
        else
          DFX_VERSION="${{ inputs.dfx-version }}" sh -c "$(curl -fsSL https://internetcomputer.org/install.sh)"
        fi
```

action.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Install dfx'
+description: 'Install the dfx utility at a particular version'
+inputs:
+  dfx-version:
+    description: >
+      The dfx version to install. This value will be used to populate
+      the DFX_VERSION variable that will be handled by the installation
+      script "install.sh".
+    default: '0.14.1'
```
smallstepman, line 9:
```suggestion
    default: 'latest'
```

smallstepman (changes requested):

smallstepman:
Let's also add changelog entry. Right under dfx changelog > UNRELEASED > DFX:
````
feat: Integration of GitHub Action for `dfx` installation

A new functionality has been implemented to enable the installation of `dfx` through GitHub Actions. This provides a more streamlined approach to installation:

```yml 
jobs:
  update-replica:
    runs-on: ubuntu-latest
    steps:
    - name: Install dfx
      uses: dfinity/sdk@master
    - name: Confirm successful installation
      run: dfx --version
```
Moreover, it accommodates the installation of specific dfx versions:

```yml
    - name: Install dfx
      uses: dfinity/sdk@master
      with: 
        dfx-version: "0.14.2-beta.2"
```
````

File: action.yml (hunk 1 above)
Comment (mmicu) [reply action.yml:16]: I kept `-ci` instead of `-c` as specified in the README.

File: action.yml [hunk 3]
```diff
@@ -6,11 +6,17 @@ inputs:
...
 runs:
   using: composite
   steps:
     - name: Install dfx
       shell: sh
-      run: DFX_VERSION=${{ inputs.dfx-version }} sh -ci "$(curl -fsSL https://internetcomputer.org/install.sh)"
+      run: |
+        echo "Version ${{ inputs.dfx-version }} will be installed"
+        if [ ${{ inputs.dfx-version }} = latest ]; then
+          sh -c "$(curl -fsSL https://internetcomputer.org/install.sh)"
```
smallstepman, line 19:
```suggestion
          sh -ci "$(curl -fsSL https://internetcomputer.org/install.sh)"
```

smallstepman (changes requested):

smallstepman (approved):
LGTM

@ericswanson-dfinity wdyt?

action.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Install dfx'
+description: 'Install the dfx utility at a particular version'
+inputs:
+  dfx-version:
+    description: >
+      The dfx version to install. This value will be used to populate
+      the DFX_VERSION variable that will be handled by the installation
+      script "install.sh".
+    default: 'latest'
```
ericswanson-dfinity, line 9: Right now `latest` refers to the last _released_ dfx version, not the last _promoted_ dfx version.  I understand that we plan to change these semantics in the future, but for the time being, this will have to be a fixed version.

ericswanson-dfinity (changes requested):

.github/workflows/tests/dfx-action.yml
```diff
@@ -0,0 +1,11 @@
+on: [push]
```
ericswanson-dfinity, line 1: Please set up this test to run on push to master, and on pull requests, but only if action.yml or this test changes.

CHANGELOG.md
```diff
@@ -4,6 +4,30 @@
...
+      run: dfx --version
+```
+
+Moreover, it accommodates the installation of specific dfx versions:
+
+```yml
+    - name: Install dfx
+      uses: dfinity/sdk@master
+      with:
+        dfx-version: "0.14.2-beta.2"
```
ericswanson-dfinity, line 28: I see where you're coming from here:

- We're using the github action defined on the master branch
```
      uses: dfinity/sdk@master
```

- to install dfx version 0.14.2-beta.2
```
      with:
        dfx-version: "0.14.2-beta.2"
```

However, it seems like it would be natural to read the following as "install dfx as built from the sdk master branch"
```
    - name: Install dfx
      uses: dfinity/sdk@master
```

Do we have an option to do the following?

```
    - name: Install dfx at currently promoted version
      uses: dfinity/sdk
```


```
    - name: Install dfx at version 0.14.2-beta.2
      uses: dfinity/sdk@0.14.2-beta.2
```

action.yml
ericswanson-dfinity, line 9: There's a potential disconnect here between:
- the git's notion of `latest`
- the latest promoted dfx version

For example, at the time of this writing, we have:
- git's notion of `latest`: 0.14.2
- latest promoted version: 0.14.1

We intend to make it so these are the same thing, but for the time being, that is not the case.

ericswanson-dfinity (changes requested):

CHANGELOG.md
smallstepman: Sadly `uses: dfinity/sdk` won’t work, I was hoping for it too - GHActions require to specify a git ref. We could do `uses: dfinity/sdk@install-dfx` for example (install-dfx would need to be a branch or a tag). Or `@latest`

jwndlng: Wouldn't that be possible if we officially [publish the action on the marketplace](https://docs.github.com/en/actions/creating-actions/publishing-actions-in-github-marketplace)? Similar to `actions/checkout@v3.5.3` we might be able to pin the version to `dfinity/sdk@v0.14.1`.

mmicu: It seems that `@ref` is mandatory, so we are always going to have one version for the `action` and one version for the actual `dfx` to install. This might be confusing.

We could have a kind of fixed ref for the action (e.g. `sdk@install-dfx` as suggested by Marcin) that we will update/overwrite if the action needs to be modified. Then, we can specify the version to install by using the `with` statement.

btw, `actions/checkout@v3` is very similar to this action. It has a certain version (`v3`) and it can refer other versions with the `ref` parameter.

jwndlng: If we're already discussing the naming convention, what if we create a separate repository specifically for hosting the action, like `dfinity/setup-dfx`? This naming convention is used by other actions that set up environments, such as [`actions/setup-node`](https://github.com/marketplace/actions/setup-node-js-environment), [`actions/setup-python`](https://github.com/marketplace/actions/setup-python) and [`hashicorp/setup-terraform`](https://github.com/marketplace/actions/hashicorp-setup-terraform). Using a dedicated repository aligns with the official approach for publishing actions on the marketplace and doesn't require additional review from GitHub. While it won't solve the versioning problem, it shouldn't be confusing since this syntax is commonly used by other actions.

ericswanson-dfinity: Great insight.  Let's do that.

smallstepman: I've requested to change the visibility of https://github.com/dfinity/setup-dfx to `public` and I'll notify once it's public

smallstepman: @mmicu https://github.com/dfinity/setup-dfx is now public 🎉

ericswanson-dfinity:
https://github.com/dfinity/setup-dfx awaits

PR #3213, opened 2023-06-30
fix: Ctrl-C right after dfx start panics in 2 minutes

lwshang:
Fixes SDK-972

Refile of #3206

# Description

With `dfx` v0.14.1, sending `Ctrl-C` right after `dfx start` will result in:
```
Running dfx start for version 0.14.1
Using shared network 'local' defined in /Users/lwshang/.config/dfx/networks.json
Stopping canister http adapter...
Stopped.
Stopping the replica...
```
Then it hangs and after 2 minutes:
```
thread 'replica-actor' panicked at 'called `Result::unwrap()` on an `Err` value: Cannot start the replica: timed out', src/dfx/src/actors/replica.rs:373:78
```

If `dfx start --background` and `Ctrl-C` immediately, the process hangs in the background. And if run `dfx start` during this 2 minutes window, it will error with "Address in use".

## Cause

The reason of this bug was in the long-running replica actor start procedure.
It wait for a port file to be ready. But the `ic-starter` process has been terminated.
So it hangs until all retry effort (retry every 0.1s for 1200 times) in 2 minutes.

## Fix

This PR fix the issue by breaking out of the retry loop when receive the stop signal.
The treatment is also applied to bitcoin and canister-http actors.

Now with the same commands, it will get following output immediately.

```
Stopping the replica...
Stopped
```

# How Has This Been Tested?

e2e test with an `exp` script.

# Checklist:


e2e/assets/expect_scripts/ctrl_c_right_after_dfx_start.exp
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/expect -df
+
+match_max 100000
+set timeout 5
```
ericswanson-dfinity, line 4: The canister http adapter times out after 5 minutes
```suggestion
set timeout 330
```

e2e/assets/expect_scripts/ctrl_c_right_after_dfx_start.exp
```diff
@@ -0,0 +1,22 @@
...
+    sleep 1
+    send \x03
+}
+
+expect {
+    timeout {
+        puts "Didn't complete before timeout."
+        exec kill -9 $pid
+        exit 2
+    }
```
ericswanson-dfinity, line 21:
```suggestion
    -re "panic" {
        puts "Panic detected."
        exit 2
    }
    -ex "expect eof" {
        wait $spawn_id
        set exit_status [lindex [wait] 3]
        if {$exit_status == 0} {
            puts "Success: Program exited with 0 error code"
            exit 0
        } else {
            puts "Error: Program exited with non-zero error code"
            exit 2
        }
    }
    timeout {
        puts "Didn't complete before timeout."
        exec kill -9 $pid
        exit 2
    }
```

ericswanson-dfinity (commented):
I was hoping to see the panic message in the output.  The test as written does detect failure if it takes more than 5 seconds for dfx to exit, so from that perspective, it seems sound.

Here is a version that detects normal exit right away and checks the error code, but if run without the fix, displays the panic along with the failure. 

What do you think?

e2e/assets/expect_scripts/ctrl_c_right_after_dfx_start.exp
lwshang: When running with fixed `dfx`, it didn't go to the `-ex "expect eof"` branch.
Could you verify if it works?

ericswanson-dfinity (approved):

PR #3219, opened 2023-07-05
fix: Wait for new module hash when installing wallet

adamspofford-dfinity:
#2407 neglected the `install_wallet` function, which does not use `install_canister` and thus did not benefit from the module hash waiting. This extracts it to shared logic between the two functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## DFX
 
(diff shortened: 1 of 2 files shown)


ericswanson-dfinity (changes requested):
Please describe in changelog

ericswanson-dfinity (approved):

PR #3221, opened 2023-07-05
chore: move extensions to dfx-core

smallstepman:
# Description

In preparation for support for custom canister types https://dfinity.atlassian.net/browse/SDK-1073

# How Has This Been Tested?

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] I have edited the CHANGELOG accordingly.
- [ ] I have made corresponding changes to the documentation.

PR #2968, opened 2023-02-13
feat(extensions): install from external registries 

smallstepman:
# Description

Modifies the `extension install` command to take a flag `--registry` that allows the registry to install from to be specified at runtime

`dfx extension install foo --registry <other-compatibility.json>`

Implements https://dfinity.atlassian.net/browse/SDK-963

# How Has This Been Tested?
added e2e test

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [x] I have edited the CHANGELOG accordingly.
- [ ] I have made corresponding changes to the documentation.

PR #3222, opened 2023-07-05
feat(extension): add support for custom canister types

smallstepman:
# Description

The PR extends functionalities to adapt to the new feature of custom canister types. The changes accommodate:
- parsing of `canister_types` in the extension manifest file 
- transformation of `.canisters.<CAN>` in dfx.json based on extension's `canister_types` declaration

Note:
- depends on https://github.com/dfinity/sdk/pull/3221
- unusable outside of https://github.com/dfinity/dfx-extensions until this lands: #2968 

Closes https://dfinity.atlassian.net/browse/SDK-1073

# How Has This Been Tested?

- [x] unit
- [x] e2e
- [x] manually

# Checklist:


lastmjs:
Question on this, if someone is using a custom type like `azle` or `kybra`, will they still be able to provide regular dfx.json properties? For example, I'd like to allow Azle and Kybra developers to override the simple `azle` and `kybra` canister configurations with the underlying dfx.json properties if they want to. Let's say they want to turn off gzip, they should be able to. Or if they want to turn on optimization or change it, they should be able to.

smallstepman:
Thanks for bringing that up; it definitely should work just as you've described it, I'll make a test case for it

src/dfx-core/src/config/model/dfinity.rs
```diff
@@ -950,42 +951,65 @@ impl Config {
...
+        content: &[u8],
+        dfx_version: &semver::Version,
+    ) -> Result<Config, StructuredFileError> {
+        let mut json: serde_json::Value = serde_json::from_slice(content)
             .map_err(|e| DeserializeJsonFileFailed(Box::new(path.clone()), e))?;
-        let json = serde_json::from_slice(content)
+        let extension_manager =
+            crate::extension::manager::ExtensionManager::new(dfx_version).unwrap();
+        custom_canister_type::transform_dfx_json_via_extension(&mut json, extension_manager)
+            .unwrap(); // TODO: error handling
```
smallstepman, line 991: I am not sure what kind of error to use over here

e2e/tests-dfx/build.bash
```diff
@@ -215,7 +215,7 @@ teardown() {
   jq '.canisters.e2e_project_backend.type="unknown_canister_type"' dfx.json | sponge dfx.json
   assert_command_fail dfx build
   # shellcheck disable=SC2016
-  assert_match 'unknown variant `unknown_canister_type`'
+  assert_match 'ExtensionNotInstalled("unknown_canister_type")' # TODO: concrete error type
```
sesi200, line 218: Open TODO

smallstepman: relates to https://github.com/dfinity/sdk/pull/3222#discussion_r1376255304

src/dfx-core/src/extension/manifest/custom_canister_type.rs
```diff
@@ -0,0 +1,364 @@
...
+        // which is not possible, because serde_json::Number does not implement it.
+        for field_name in remove_fields {
+            final_fields.remove(&field_name);
+        }
+        // Override custom canister declaration values by the real canister_declaration
+        for (key, value) in values.iter() {
+            if key != "type" && key != "canister_name" {
+                final_fields.insert(key.clone(), value.clone());
+            }
+        }
```
smallstepman, line 198: @ericswanson-dfinity I just realized this is buggy... I'm working on a fix

src/dfx-core/src/config/model/dfinity.rs
ericswanson-dfinity: Since from_slice will now have new failure modes, we'll have to change the return type.

We can reduce the scope of those failure modes, though.

How about this?

1. Instantiate the ExtensionManager in the caller
 - make EnvironmentImpl::new() create the ExtensionManager.
 - rename EnvironmentImpl::new_extension_manager()` and 

2. Pass a trait rather than the whole ExtensionManager 
- make `trait TransformConfiguration`
   - single method: `transform(&mut json) -> Result<(), TransformConfigurationError>`
   - either pass `Option<TransformConfiguration>` to these methods, or for tests pass an object that is noop for this method

3. Change these methods' error type from StructuredFileError to ReadConfigurationError (possibilities: StructuredFileError, TransformConfigurationError)

This way the error return type of `from_slice` and friends will reflect the errors that can occur, but won't also have to include errors related to instantiating the extension manager.

ericswanson-dfinity (changes requested):
Let's start with documentation in [docs/concepts](https://github.com/dfinity/sdk/tree/master/docs/concepts) describing how to define an custom canister type and how to use one in your project.  This will help with the review for questions like "what is `"type": "<x>:<y>"`

src/dfx-core/src/extension/manifest/custom_canister_type.rs
smallstepman: After all, I'm not sure what should be the desired behavior...
I extended unit tests with:
```rust
    #[test]
    fn test_op_replace_basic() {
        test_op!(
            custom_canister_template = r#"
        {
            "main": { "replace": { "input": "{{canister_name}}", "search": "frontend_(.*)", "output": "oowee/$1/main.ts" } }
        }
        "#,
            dfx_json_canister_values = r#"
        {
            "canister_name": "frontend_xyz"
        }
        "#,
            expected = r#"
        {
            "main": "oowee/xyz/main.ts"
        }
        "#
        );
    }

    #[test]
    fn test_op_replace_nested() {
        test_op!(
            custom_canister_template = r#"
        {
            "main": { "replace": { "input": "{{main}}", "search": "(.*)", "output": "oowee/$1" } }
        }
        "#,
            dfx_json_canister_values = r#"
        {
            "main": "src/main.ts"
        }
        "#,
            expected = r#"
        {
            "main": "oowee/src/main.ts"
        }
        "#
        );
    }
```

the `_basic` will pass, but what should be the expected outcome for `_nested`? This is related to https://github.com/dfinity/sdk/pull/3222#issuecomment-1624073606

ericswanson-dfinity: I don't understand how this is related to the comment.

From reading the comment I think the expectation is that
``` json
{
    "type": "kybra",
    "gzip": "true"
}
```
should transform according to the rules of the kybra extension, but also keep the "gzip" field.

Also I think these tests would benefit from some prose descriptions of what's being tested, what's the expected behavior, and what are the nuances.  As a reader I don't want to have to parse a regex in my head in order to figure out what the test is checking.

smallstepman: apologies, I mixed too much info in that example; let me untangle
### case 1 - append user-defined field
is when someone defines `gzip: true` in their dfx.json, and `gzip` is not defined in custom canister type template. This works as expected. 
```rust
    #[test]
    fn test_op_replacement_1() {
        test_op!(
            custom_canister_template = r#"
        {
            "main": "something.py"
        }
        "#,
            dfx_json_canister_values = r#"
        {
            "gzip": true
        }
        "#,
            expected = r#"
        {
            "gzip": true,
            "main": "something.py"
        }
        "#
        );
    }
```

### case 2 - override the field
is when someone defines `gzip: true` in their dfx.json, and `gzip: false` is defined in custom canister type template. This also works as expected. 

```rust
    #[test]
    fn test_op_replacement_2() {
        test_op!(
            custom_canister_template = r#"
        {
            "main": "something.py",
            "gzip": false
        }
        "#,
            dfx_json_canister_values = r#"
        {
            "gzip": true
        }
        "#,
            expected = r#"
        {
            "gzip": true,
            "main": "something.py"
        }
        "#
        );
    }
```

### case 3 - nested field override 
is when someone defines `gzip: true` in their dfx.json, and `gzip: false` is defined in custom canister type template, and also `{{gzip}}` is used in the template
```rust 
    #[test]
    fn test_op_replacement_3() {
        test_op!(
            custom_canister_template = r#"
        {
            "main": "{{gzip}}.py",
            "gzip": false
        }
        "#,
            dfx_json_canister_values = r#"
        {
            "gzip": true
        }
        "#,
            expected = r#"
        {
            "gzip": true,
            "main": "true.py"
        }
        "#
        );
    }
```
Currently, the above test will pass, which I think makes the most sense?

ericswanson-dfinity:
Replaced by https://github.com/dfinity/sdk/pull/3641

PR #3235, opened 2023-07-12
fix(extension-run): allow params with multiple values

smallstepman:
-------

# Description

Adds a configuration flag for `extension.json`: `.subcommands.<CMD>.args.<ARG>.multiple`, which allows passing multiple values for one param, e.g.:
```console
dfx extension install nns
dfx nns install --ledger-accounts fefe efef eeff
```

# How Has This Been Tested?

e2e + unit

# Checklist:

### DIFF
--- a/src/dfx/src/lib/extension/manifest/extension.rs
+++ b/src/dfx/src/lib/extension/manifest/extension.rs
@@ -60,7 +60,8 @@ pub struct ExtensionSubcommandArgOpts {
     pub about: Option<String>,
     pub long: Option<String>,
     pub short: Option<char>,
-    pub multiple: Option<bool>,
+    #[serde(default)]
+    pub multiple: bool,
 }
 
 impl ExtensionSubcommandArgOpts {
@@ -79,8 +80,8 @@ impl ExtensionSubcommandArgOpts {
         if let Some(s) = self.short {
             arg = arg.short(s);
         }
-        if matches!(self.multiple, Some(true)) {
-            arg = arg.num_args(0..)
+        if self.multiple {
+            arg = arg.num_args(0..);
         }
         Ok(arg
             // let's not enforce any restrictions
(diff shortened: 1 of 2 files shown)


src/dfx/src/lib/extension/manifest/extension.rs
```diff
@@ -78,9 +79,13 @@ impl ExtensionSubcommandArgOpts {
...
             arg = arg.short(s);
         }
+        match self.multiple {
+            Some(m) if m => arg = arg.num_args(0..),
+            _ => {}
+        }
         Ok(arg
             // let's not enforce any restrictions
-            .allow_hyphen_values(false)
+            .allow_hyphen_values(true)
```
smallstepman, line 88: this isn't what this PR truly targets, but I thought its good to fix it anyway

ericswanson-dfinity (approved):

src/dfx/src/lib/extension/manifest/extension.rs
```diff
@@ -60,6 +60,7 @@ pub struct ExtensionSubcommandArgOpts {
     pub about: Option<String>,
     pub long: Option<String>,
     pub short: Option<char>,
+    pub multiple: Option<bool>,
```
ericswanson-dfinity, line 63: beyond the scope of this PR, but this should have doc comments, and we should generate a JSON schema.

Also, especially with bool arguments that default to false, we can simplify things a little bit by using `bool` (default false) rather than `Option<bool>` (default None, assumed to be false):
```suggestion
    #[serde(default)]
    pub multiple: bool,
```

smallstepman: thanks! 
https://dfinity.atlassian.net/browse/SDK-1180

I'll change it now from `Option<bool>` to `bool`

PR #3258, opened 2023-07-20
docs: add asset canister interface doc

ericswanson-dfinity:
Adds documentation for the asset canister interface.  Maybe we will have to move it in order to make it accessible from the portal.

The document doesn't describe any details at all about the `http_request` method.  I'd like to leave that as follow-up work and will open a JIRA ticket to do so if there is agreement.


docs/design/asset-canister-interface.md
```diff
@@ -0,0 +1,508 @@
+# Asset Canister Interface
```
dfx-json, line 1: First off, I think it's nice that you are separating these concepts: Asset Canister Interface and Asset Canister implementation. For me, this is helpful in understanding the things I can and can't do as a developer.

Maybe we also ought to think about renaming our asset canister implementation so it's clear its _one_ possible implementation of many of the interface.

dfx-json: It may also be useful (predominantly to veteran IC developers) to note that this is not to be confused with the _implementation_ of the asset canister that they are likely accustomed to using

docs/design/asset-canister-interface.md
```diff
@@ -0,0 +1,508 @@
...
+| [Configuration Reference](#configuration-reference)      |
+| [Permission Reference](#permission-reference)            |
+
+## Retrieving Assets
+
+Asset retrieval begins with a call to either [get()](#method-get) or [http_request()](#method-http_request).
+
+### Asset Lookup
+
+The asset canister looks up assets by [key](#key) in a case-sensitive manner.
```
dfx-json, line 31: It may be helpful to understand what a `key` is in this context. Is it a filename? Is it something generated by dfx? As a developer, why do I care about the key?

docs/design/asset-canister-interface.md
ericswanson-dfinity: Before today's commit https://github.com/dfinity/sdk/pull/3258/commits/43660ece47261bab5a136f577bdaec0fe262d66f I had this:

> This document is meant to describe the interface, not the idiosyncrasies of the implementation.

I wonder if I should re-add something like this

docs/design/asset-canister-interface.md
```diff
@@ -0,0 +1,508 @@
...
+1. [create_batch()](#method-create_batch) once.
+2. [create_chunk()](#method-create_chunk) one or more times, which can occur concurrently.
+3. [commit_batch()](#method-commit_batch) zero or more times with `batch_id: 0`.
+4. [commit_batch()](#method-commit_batch) once with the batch ID from step 1, which indicates the batch is complete.
+
+The reason for multiple rather than single calls to [commit_batch][#method-commit_batch] is that certificate computation for an entire batch may exceed per-message computation limits.
+
+### Batch Updates By Proposal
+
+If an SNS controls an asset canister, it can update the assets by proposal. In this scenario, there are two principals:
```
dfx-json, line 66: Link to SNS documentation would be useful here for those unfamiliar with SNS

docs/design/asset-canister-interface.md
```diff
@@ -0,0 +1,508 @@
...
+4. [compute_evidence()](#method-compute_evidence) until the method returns `Some(evidence)`.
+
+The Preparer then furnishes the Committer with the following information:
+- the batch ID
+- the computed evidence
+
+The Committer then calls the following method upon approval of the proposal:
+1. [commit_proposed_batch()](#method-commit_proposed_batch)
+
+If the proposal is not approved, the Preparer must call [delete_batch()](#method-delete_batch).
```
dfx-json, line 83: should we talk at all about what happens if the Preparer fails to call `delete_batch()`?

docs/design/asset-canister-interface.md
```diff
@@ -0,0 +1,508 @@
...
+
+It is also possible to upload a single content encoding of a single asset by calling the [store()](#method-store) method.  The size of the content encoding must not exceed the message ingress limit.
+
+## Type Reference
+
+### Asset
+
+#### Key
+
+The `key` is a case-sensitive string that identifies the asset. By convention, all asset keys begin with a forward slash. For example, `/index.html`.
```
dfx-json, line 95: my previous comment alluded to this definition. perhaps a link to this section from up above would suffice.

docs/design/asset-canister-interface.md
```diff
@@ -0,0 +1,508 @@
...
+
+The `max_age` field is the maximum number of seconds that the asset can be cached by a browser or CDN. It is used to set the `max-age` value of the `Cache-Control` header when serving the asset over HTTP.
+
+#### Headers
+
+The `headers` field is a list of additional headers to set when serving the asset over HTTP.
+
+#### Enable Aliasing
+
+This field is sometimes called `is_aliased` in the interface, but it should be called `enable_aliasing` everywhere.  It enables a behavior.  It does not indicate that the behavior is currently active for the asset.
```
dfx-json, line 127: this paragraph seems to have a bit of authorial intrusion. can I use `enable_aliasing` as a header field?

docs/design/asset-canister-interface.md
```diff
@@ -0,0 +1,508 @@
...
+
+### Method: `certified_tree`
+
+This method returns the certified tree.
+
+### Convenience Methods
+
+Each of these methods is the equivalent of its respective batch operation.
+
+While they are provided for "convenience," some don't actually make sense.  For example, [set_asset_content()](#operation-setassetcontent) requires chunk ids, but [create_chunk()](#method-createchunk) requires a batch.
```
dfx-json, line 369: putting the next two paragraphs within an admonition may be a nice way to add the authorial perspective

dfx-json (approved):

docs/design/asset-canister-interface.md
ericswanson-dfinity: Yeah, I struggled a bit with how much to define, and at what point, and how much to duplicate.  That's why `key` is a link here, to the field's definition in the "Type Reference."  I'll try adding some more prose here.

> Key
The key is a case-sensitive string that identifies the asset. By convention, all asset keys begin with a forward slash. For example, /index.html.

docs/design/asset-canister-interface.md
ericswanson-dfinity: Reworded.  I learned a new term today.  Thanks!

docs/design/asset-canister-interface.md
ericswanson-dfinity: But I do still think a little more prose above will help, even if it duplicates some of what is here.

PR #3267, opened 2023-07-28
ci: Added workflow to update list of whitelisted canisters in dfinity/motoko-playground repo

smallstepman:
# Description

Adds workflow which:
1. takes semver of released dfx version (checks out git tag, and computes sha of assetcanister.wasm.gz)
1. modifies `service/wasm-utils/whitelisted-wasms.txt` in `dfinity/motoko-playground`  by appending `HASH` before the last line 
1. opens a PR in `dfinity/motoko-playground`  

Closes https://dfinity.atlassian.net/browse/SDK-1186
cc @sesi200 

# How Has This Been Tested?

Tested manually on fork repo:
- **<updated-again!>** workflow run: https://github.com/smallstepman/sdk/actions/runs/5825183618/job/15796092095
- **<updated-again!>** created PR: https://github.com/smallstepman/motoko-playground/pull/11/files

# Checklist:


.github/workflows/broadcast-frontend-hash_modify-file.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+FILE="$1"
+DFX_VERSION="$2"
+NEW_HASH="$3"
+
+awk -v dfx_version="$DFX_VERSION" -v new_hash="$NEW_HASH" '
+BEGIN { OFS="" }
+
+/const WHITELISTED_WASMS: \[&str; [0-9]+\] = \[/ {
```
chenyan-dfinity, line 10: This assumes a lot of things about the Rust code. How about appending the hash to a txt file, and the Rust code uses `include_str!` to load it?

.github/workflows/broadcast-frontend-hash_modify-file.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+FILE="$1"
+DFX_VERSION="$2"
+NEW_HASH="$3"
+
```
ericswanson-dfinity, line 6: While `awk` is a good choice here, it's easier to write than to read.

It might help to describe at a high level what is going on here.  The pathname tells us roughly "as part of the broadcast-frontend-hash workflow, modify a file."

Example:
```
# In a declaration like the below from https://github.com/dfinity/motoko-playground/blob/main/service/wasm-utils/lib.rs,
# increment the array count and prepend a new entry.
#   const WHITELISTED_WASMS: [&str; 1] = [ 
#      "88d1e5795d29debc1ff56fa0696dcb3adfa67f82fe2739d1aa644263838174b9", // dfx 0.15.0 frontend canister
#.  ];

File: .github/workflows/broadcast-frontend-hash_modify-file.sh [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+awk -v dfx_version="$DFX_VERSION" -v new_hash="$NEW_HASH" '
+BEGIN { OFS="" }
+
+/const WHITELISTED_WASMS: \[&str; [0-9]+\] = \[/ {
+    split($0, arr, ";")
+    split(arr[2], subarr, "]")
+    num = subarr[1] + 1
+    print arr[1] "; " num "]" subarr[2]
```
ericswanson-dfinity, line 14: When I ran this script locally, the resulting diff had a trailing whitespace after the final `[`.  This might cause the PR to fail CI, if it's making a `cargo fmt` check.

.github/workflows/broadcast-frontend-hash.yml
```diff
@@ -0,0 +1,70 @@
...
+          token: ${{ secrets.NIV_UPDATER_TOKEN }}
+          repository: ${{ env.PLAYGROUD_REPO }}
+          path: motoko-playground
+
+      - name: Get new hash
+        working-directory: ./sdk
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          export NEW_HASH=$(.github/workflows/broadcast-frontend-hash_extract-hash-from-release-notes.sh ${{ inputs.dfx_version }})
```
ericswanson-dfinity, line 42: shellcheck warns about getting the results of a subshell and exporting it on the same line, because it ignores errors.

But also, in the context of a github workflow where we are appending this to `$GITHUB_ENV` anyway, we don't need to export at all.

```suggestion
          NEW_HASH=$(.github/workflows/broadcast-frontend-hash_extract-hash-from-release-notes.sh ${{ inputs.dfx_version }})
```

.github/workflows/broadcast-frontend-hash_extract-hash-from-release-notes.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+DFX_VERSION="$1" # e.g. 0.14.1
+
+if [[ "$DFX_VERSION" == "latest" ]]
```
ericswanson-dfinity, line 5: Another option:
1. check out the sdk repo at the release tag
2. run `shasum -a 256 src/distributed/assetstorage.wasm.gz`

This will work even before the release notes are added to the release tag.

.github/workflows/broadcast-frontend-hash_extract-hash-from-release-notes.sh
```diff
@@ -0,0 +1,24 @@
...
+end=$(awk -v start="$start" 'NR > start && /^### /{ print NR; exit }' release.md)
+
+if [[ -z "$end" ]]
+then
+  end=$(wc -l < release.md)
+
+fi
+
+# subtract 1 from end, so it won't include next header line
+end=$((end-1))
```
ericswanson-dfinity, line 22: Does it make sense to do this if the `wc -l` path was taken above?

ericswanson-dfinity (changes requested):
IIRC, we discussed how it would be good to be able to open these PRs in advance of making a release.  This is due to the natural lag time in opening a PR and having it approved.  Otherwise, when we post that a new dfx release is available, we have to note that it won't work with the playground until some PR is approved (and possibly even deployed).

So I would suggest rather than scraping release notes for a hash, have this workflow get the hash from `src/distributed/assetstorage.wasm.gz` on a ref. This could either be `master` (in advance of a release), but it could also be a release tag after the fact.

.github/workflows/broadcast-frontend-hash_modify-file.sh
smallstepman: https://github.com/dfinity/motoko-playground/pull/185

.github/workflows/broadcast-frontend-hash_extract-hash-from-release-notes.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
smallstepman: This is just personal opinion, but the placement of that particular directory always seemed unintuitive to me. What do you think about moving it from `scripts/workflows` to `.github/workflows/scripts`?

.github/workflows/broadcast-frontend-hash.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Modify Rust file
+        working-directory: ./motoko-playground
+        run: |
+          tempfile=$(mktemp)
+          # `$i` specifies that the insertion should happen at the line before the last line. `$` matches the last line, and `i` inserts before the matched line.
+          # `\/\/` is `//` escaped for sed, same for `\"` -> `"`, and `\ ` -> ` `
+          sed '$i\ \ \ \ \"${{ env.NEW_HASH }}\", \/\/ dfx ${{ inputs.dfx_version }} frontend canister' "${{ env.FILE }}" > "$tempfile"
+          # not worth it to install sponge
+          mv "$tempfile" "${{ env.FILE }}"
```
ericswanson-dfinity, line 54: For reference: https://github.com/dfinity/motoko-playground/pull/185

First, let's look at https://github.com/dfinity/motoko-playground/blob/main/service/wasm-utils/whitelisted_wasms.txt
```
[
    "88d1e5795d29debc1ff56fa0696dcb3adfa67f82fe2739d1aa644263838174b9", // dfx 0.15.0 frontend canister
]
```

The extension of `whitelisted_wasms.txt` is `.txt`, which communicates that it's a text file.  But it contains what looks like JSON array brackets.  However, it also contains a comment, so maybe its content is json5.

But continuing, `fn is_whitelisted(wasm)` just looks for the hash anywhere in the file:
```
    let white_list = include!("whitelisted_wasms.txt");
    if white_list.contains(&wasm_hash.as_str()) {
```

My recommendation is to revise that whitelist file to be an actual text file, containing no JSON brackets.  Its format is, per line: a hash, followed optionally by a description.

```
88d1e5795d29debc1ff56fa0696dcb3adfa67f82fe2739d1aa644263838174b9 dfx 0.15.0 frontend canister
```

That works fine with the existing method of looking for the hash anywhere in the file, and adding a new entry will be trivial:

```suggestion
          echo "${{ env.NEW_HASH }} dfx ${{ inputs.dfx_version }} frontend canister" >> "${{ env.FILE }}"
```

ericswanson-dfinity (changes requested):

.github/workflows/broadcast-frontend-hash.yml
smallstepman: The whitelisted_wasms.txt was accepted by the repo owners. I'd like to avoid going round and round; could you @chenyan-dfinity please take a look and tell us what you think?

chenyan-dfinity: That txt file is a Rust code snippet, not json. What's the motivation for this change? We don't have to use `sed` here, `head -n -1` would do the work.

ericswanson-dfinity: Got it.  I missed the distinction between `include!("whitelisted_wasms.txt")` and `include_str!("whitelisted_wasms.txt")`.  The motivation is to have some reasonable way to add to the contents.  If we can get there with `head` rather than `sed`, then all good.

smallstepman: The behavior of the `head` command differs between the GNU and BSD versions. The command `head -n -1` will work on the GNU version of `head` but will not work as expected on the BSD version (which is the one included on macOS). GH workflow is based on `ubuntu-latest` so it will work there, however, the proposed `sed` solution works on both MacOS and Linux, making it debuggable and easier to work with in my opinion. Leaving the format of the `.txt` file aside, I don't understand what's so bad about `sed` solution, in my opinion, the only unclear thing there was `$i` which behavior of I documented in the comment above it. 

Here is `head`-based solution that works on both OSes:
```bash
head -n $(($(wc -l < whitelisted_wasms.txt) - 1)) whitelisted_wasms.txt > tmp
echo "    \"${{ env.NEW_HASH }}\", \/\/ dfx ${{ inputs.dfx_version }} frontend canister\n]" >> tmp
mv -f tmp whitelisted_wasms.txt
```

ericswanson-dfinity: Recommendation: go with `head -n -1`:
- It's simpler than either the `sed` or `head(math(wc))` solutions
- It works on `ubuntu-latest` where the workflow runs
- A github workflow isn't so "debuggable" with `gh workflow run` if you need to be concerned about differences between the local environment and where it `runs-on`, because what does `gh workflow run` prove in that case?

> don't understand what's so bad about sed solution

A variant on an old joke: "A software engineer, confronted with a problem, thinks "I know, I'll use `sed`." Now he has two problems."

Also maintenance.  Sometime in the future, someone else will have to understand how this works.

With more experience will come a more productive response to `sed '$i\ \ \ \ \"${{ env.NEW_HASH }}\", \/\/ `: "I'll find a different way to do this."

.github/workflows/broadcast-frontend-hash_extract-hash-from-release-notes.sh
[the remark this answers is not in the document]
ericswanson-dfinity: Both locations have their merits; neither option is right or wrong.  An opportunity for bikeshedding.

The thinking behind having them in `scripts/workflows/` is that GitHub directly reads files under .github/ to determine its configuration for our repo.  Put another way, the `.github/` directory has a specific structure and each file has a specific meaning (to GitHub):
- .github/ISSUE_TEMPLATE/*.md issue templates
- .github/PULL_REQUEST_TEMPLATE.md ...
- .github/CODEOWNERS
- .github/workflows/*.yml workflows
- and so on..

One could certainly argue that these scripts define part of that configuration.  While that is true, GitHub doesn't load them directly when determining its configuration; our job steps run them.  They don't have the same meaning to GitHub in the same way that files like `.github/CODEOWNERS` and `.github/workflows/e2e.yml` do.

ericswanson-dfinity (approved):

PR #3281, opened 2023-08-04
chore: update Rust toolchain

sesi200:
# Description

ProdSec recommends bumping the Rust version to 1.71.1

# How Has This Been Tested?

CI

# Checklist:


ericswanson-dfinity (changes requested):
Please also update the github workflows

src/dfx/src/lib/builders/mod.rs
```diff
@@ -351,21 +353,25 @@ pub fn run_command(args: Vec<String>, vars: &[Env<'_>], cwd: &Path) -> DfxResult
 
 /// Set the permission of the given file to be writeable.
 pub fn set_perms_readwrite(file_path: &PathBuf) -> DfxResult<()> {
-    let mut perms = std::fs::metadata(file_path)
-        .with_context(|| {
+    #[cfg(not(target_os = "windows"))]
```
sesi200, line 356: I don't understand the documentation. Other places in the repo only care about permissions on unix, so this could be the way to do it?

ericswanson-dfinity, line 356: Since this seems to be controlling usage of `use std::os::unix::fs::PermissionsExt;`, I think we ought to use `cfg(unix)`
The other places in our codebase that have `#[cfg(not(target_os = "windows"))]` related to permissions are similarly saying "on not-windows, do this unix filesystem thing".  I think they would be improved by instead doing unix things on cfg(unix).

```suggestion
    #[cfg(unix)]
```

ericswanson-dfinity: The intent of this method is to make a file owner-writable.  See https://github.com/dfinity/sdk/pull/3017 for where this was necessary in order to avoid `Permission denied` errors in post-build.

One day we may have a functioning windows build, but until then, it's fine if this method only does anything on cfg(unix).

ericswanson-dfinity (approved):

Issue #2771, opened 2022-11-17
`--no-frontend` flag


letmejustputthishere:
I tried the following:

1. `dfx new --no-frontend`

I expected to see this happen: No asset canister present, no entries in `dfx.json`

Instead, this happened: Asset canister and entrie present.

### Meta

`dfx --version`:
```
0.12.0
```

<details><summary>Further log output</summary>
<p>

```
<log output>
```

</p>
</details>

letmejustputthishere:
if this behaviour is changed, we need to update this [tutorial](https://internetcomputer.org/docs/current/developer-docs/build/cdks/cdk-rs-dfinity/multiply-dependency#update-interface-description-file) and remove the step where we delete the frontend canister from `dfx.json`

sesi200:
The tutorial is still correct. We hope that this situation will be improved once dfx extensions land

domwoe:
`dfx new --no-frontend` has no associated explanation in `dfx new` and doesn't work as expected (There should be no asset canister in `dfx.json` and no `src/xxx_frontend`.)  as @letmejustputthishere noted.

PR #3283, opened 2023-08-05
chore: Switch `dfx new` templates to patch-based system

adamspofford-dfinity:
This changes the mechanism of `dfx new` template application. Previously, the template corresponding to `--type` contained all project structure and interesting tidbits, and then the frontend template touched only the actual files. Now, `.patch` and `.json-patch` files are automatically interpreted to overlay templates, allowing a base README and dfx.json to be specified, the project type template to fill in only the backend canister and README links, and the frontend template to fill in only the frontend canister.

This PR aims to change zero existing functionality - the observable effect should be zero, with the new tests succeeding both before and after. New templates, as well as axing sample-asset.txt and changing default dotenv, will follow in another PR.

This required implementing a patch file applicator crate, located at `src/lib/apply-patch`. This crate will be in dfinity/sdk only temporarily; a later PR will migrate it out to its own repo and package, with more features implemented, and the parts that are missing have been commented as such.


src/dfx/assets/new_project_base_files/dfx.json
```diff
@@ -0,0 +1,11 @@
+{
+    "version": 1,
+    "canisters": {},
+    "defaults": {
+        "build": {
+            "packtool": "",
+            "args": ""
+        }
+    },
```
sesi200, line 9:
```suggestion
```
These lines serve no purpose AFAIK. Is it worth keeping them?

src/dfx/src/commands/new.rs
```diff
@@ -196,7 +197,32 @@ fn write_files_from_entries<R: Sized + Read>(
...
+                .with_context(|| format!("Failed to parse {}", p.display()))?;
+            let patch = patch::Patch::from_single(v)
+                .map_err(|e| anyhow!("Failed to parse {}: {e}", p.display()))?;
+            let to_patch = p.with_extension("");
+            let existing_content = dfx_core::fs::read_to_string(&to_patch)?;
+            let patched_content = apply_patch::apply_to(&patch, &existing_content)
+                .with_context(|| format!("Failed to patch {}", to_patch.display()))?;
+            dfx_core::fs::write(&to_patch, &patched_content)?;
+        } else {
+            create_file(log, p.as_path(), &v, dry_run)?;
```
sesi200, line 224: This line adheres to `dry_run`, but the other branches write the file unconditionally. Should the other branches also take `dry_run` into account?

smallstepman:
> This PR aims to change zero existing functionality - the observable effect should be zero

I'm curious if this PR does close these items https://github.com/dfinity/sdk/pull/2020 https://github.com/dfinity/sdk/issues/2771 https://github.com/dfinity/sdk/pull/2386 https://github.com/dfinity/sdk/issues/1648, or if this will come in future PR?

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+        // finally, copy everything between the final patch and the end of the file
+        new_content.push_str(&content[prev_idx..]);
+        Ok(new_content)
+    }
+}
+
+fn compare_range(content: &str, lines: &[&str], ignore_whitespace: bool) -> bool {
+    for (left, &right) in content.lines().zip(lines) {
+        if ignore_whitespace {
```
ericswanson-dfinity, line 142: is the ignore/exact whitespace logic reversed here?

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+                }
+            }
+        }
+        // finally, copy everything between the final patch and the end of the file
+        new_content.push_str(&content[prev_idx..]);
+        Ok(new_content)
+    }
+}
+
+fn compare_range(content: &str, lines: &[&str], ignore_whitespace: bool) -> bool {
```
ericswanson-dfinity, line 140:
```suggestion
fn equal_range(content: &str, lines: &[&str], ignore_whitespace: bool) -> bool {
```

src/dfx/src/commands/new.rs
```diff
@@ -196,7 +197,32 @@ fn write_files_from_entries<R: Sized + Read>(
...
+            let to_patch = p.with_extension("json");
+            ensure!(
+                to_patch.exists(),
+                "Failed to patch {}: not found",
+                to_patch.display()
+            );
+            let mut value = load_json_file(&to_patch)?;
+            json_patch::patch(&mut value, &patch)
+                .with_context(|| format!("Failed to patch {}", to_patch.display()))?;
+            save_json_file(&to_patch, &value)?;
```
ericswanson-dfinity, line 212: suggestion: extract method

src/dfx/src/commands/new.rs
```diff
@@ -196,7 +197,32 @@ fn write_files_from_entries<R: Sized + Read>(
...
+        } else if p.extension() == Some("patch".as_ref()) {
+            let v = std::str::from_utf8(&v)
+                .with_context(|| format!("Failed to parse {}", p.display()))?;
+            let patch = patch::Patch::from_single(v)
+                .map_err(|e| anyhow!("Failed to parse {}: {e}", p.display()))?;
+            let to_patch = p.with_extension("");
+            let existing_content = dfx_core::fs::read_to_string(&to_patch)?;
+            let patched_content = apply_patch::apply_to(&patch, &existing_content)
+                .with_context(|| format!("Failed to patch {}", to_patch.display()))?;
+            dfx_core::fs::write(&to_patch, &patched_content)?;
```
ericswanson-dfinity, line 222: suggestion: extract method (and this will make it more obvious, as @sesi200 noted, that `dry_run` isn't taken into account)

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+                    + 1;
+                if compare_range(
+                    &content[start..end],
+                    &expected_lines,
+                    self.ignore_whitespace,
+                ) {
+                    start..end
+                } else {
+                    return Err(MismatchError::LineMismatch);
+                }
```
ericswanson-dfinity, line 109: suggestion: extract method

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+        Self {
+            _reject_relative_path_segments: false,
+            ..self
+        }
+    }
+    /// Applies a single-file patch to `content`. File paths in the patch file are ignored.
+    // todo: use an iterator instead of returning String. mvp: our files are small.
+    pub fn apply_to(&self, patch: &Patch, content: &str) -> Result<String, MismatchError> {
+        let mut expected_lines = vec![];
+        let mut prev_idx = 0;
```
ericswanson-dfinity, line 65: IIUC, `prev_idx` is an offset within `content`.  The data before `prev_idx` has been dealt with (either copied to the output, or deleted), and after is yet to be dealt with.

Suggestions to improve readability:
- rename `content` -> `original`
- rename `new_content` -> `revised`
- One of the following for `prev_idx`:
  - rename `prev_idx` to something like `read_offset`.  Might even be able to get away with `i`.
  - change it to a slice called something like `original_remaining` or even `remaining` that essentially advances through `original`: it's always what `&original[prev_idx..]` would have been.

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[error("File did not contain enough lines")]
+    NotEnoughLines,
+    #[error("Mismatch between context/removal line and file")]
+    LineMismatch,
+}
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
```
ericswanson-dfinity, line 174: `std::io::Error` generally doesn't have enough contextual information to be useful by itself.  I would only add std::io::Error as a part of another type that does have context.
```suggestion
```

ericswanson-dfinity (commented):
For this apply-patch library, I think one of the big challenges is going to be making it readable and keeping it that way.  I left some notes that I hope/think will help.  I understand you are going for an MVP and may not want to really think about that right now.

src/canisters/frontend/ic-certified-assets/src/asset_certification/types/http.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(deprecated)] // candid arc_type
```
ericswanson-dfinity, line 1: We do need to figure out what to do about these warnings.  Is this it?

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+            }
+        } else if !left
+            .chars()
+            .filter(|ch| !ch.is_whitespace())
+            .eq(right.chars().filter(|ch| !ch.is_whitespace()))
+        {
+            return false;
+        }
+    }
+    true
```
ericswanson-dfinity, line 154:
```suggestion
    content.lines().zip(lines).all(|(left, &right)| {
        if ignore_whitespace {
            left.chars()
                .filter(|ch| !ch.is_whitespace())
                .eq(right.chars().filter(|ch| !ch.is_whitespace()))
        } else {
            left == right
        }
    })
```

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                    Line::Add(s) => {
+                        new_content.push_str(s);
+                        new_content.push('\n');
+                    }
+                    Line::Remove(_) => {
+                        orig_lines.next().unwrap();
+                    }
+                }
+            }
```
ericswanson-dfinity, line 132: suggestion: extract method

src/dfx/src/util/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(deprecated)] // candid arc_type
```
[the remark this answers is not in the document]
adamspofford-dfinity: It is. The warning appears everywhere Type is used.

src/lib/apply-patch/src/lib.rs
adamspofford-dfinity: I'll just delete this then, it's a placeholder for nonexistent-as-of-yet functionality anyway.

src/dfx/assets/new_project_base_files/dfx.json
adamspofford-dfinity: If it should be removed, I'll remove it when I update the templates themselves.

adamspofford-dfinity:
> > This PR aims to change zero existing functionality - the observable effect should be zero
> 
> I'm curious if this PR does close these items #2020 #2771 #2386 #1648, or if this will come in future PR?

Future PR.

src/dfx/assets/new_project_base_files/dfx.json
```diff
@@ -0,0 +1,11 @@
...
+    "output_env_file": ".env"
+}
```
ericswanson-dfinity, line 11: let's have a newline at the end of this file

src/dfx/assets/new_project_motoko_files/README.md.patch
```diff
@@ -0,0 +1,7 @@
+--- a/README.md
++++ b/README.md
+@@ -9,2 +9,4 @@
+ - [Quick Start](https://internetcomputer.org/docs/quickstart/quickstart-intro)
+ - [SDK Developer Tools](https://internetcomputer.org/docs/developers-guide/sdk-guide)
++- [Motoko Programming Language Guide](https://internetcomputer.org/docs/current/developer-docs/build/cdks/motoko-dfinity/motoko/)
++- [Motoko Language Quick Reference](https://internetcomputer.org/docs/current/references/motoko-ref/)
```
ericswanson-dfinity, line 7: This link (https://internetcomputer.org/docs/current/references/motoko-ref/) currently redirects to https://internetcomputer.org/docs/current/motoko/main/base/ ("Motoko base library reference").  This is a list of base library types (a useful link to have in the readme, directly).

I imagine this is copied from somewhere.

The Motoko language quick reference url is: https://internetcomputer.org/docs/current/motoko/main/language-manual.

adamspofford-dfinity: It's what was in the original, I didn't check the file. They're also all old redirects, I'll update those too just in case they break someday.

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    fn find_fixed_range(
+        &self,
+        original_content: &str,
+        hunk: &Hunk,
+        expected_lines: &[&str],
+    ) -> Result<Range<usize>, MismatchError> {
+        let mut lines = original_content.match_indices('\n');
```
ericswanson-dfinity, line 125: Would this be better called `line_endings` or `newlines`?

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+            0
+        } else {
+            lines
+                .nth(start_line - 1)
+                .ok_or_else(|| MismatchError::NotEnoughLines {
+                    expected: start_line - 1,
+                    found: original_content.lines().count(),
+                })?
+                .0
+                + 1
```
ericswanson-dfinity, line 139: Do you think there's enough duplication between here and the `let end` expression to apply DRY here?

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+        if ignore_whitespace {
+            if !from_content
+                .chars()
+                .filter(|ch| !ch.is_whitespace())
+                .eq(from_patch.chars().filter(|ch| !ch.is_whitespace()))
+            {
+                return Err(MismatchError::LineMismatch {
+                    from_content: from_content.to_string(),
+                    from_patch: from_patch.to_string(),
+                    whitespace_insensitive: ignore_whitespace,
```
ericswanson-dfinity, line 184: We now have three ways in the code of expressing a similar idea:
- ignore_whitespace
- whitespace_insensitive
- exact_whitespace

Could we pick one?

adamspofford-dfinity: I would prefer to keep `ignore_whitespace` and `whitespace_insensitive` in the API - one is an instruction to the system, one is the state of a match that produced an error. The name of the internal variable can be changed to match one of the two, though.

src/lib/apply-patch/src/lib.rs
adamspofford-dfinity: I see *an* opportunity, but I assume you're talking about the error-raising part and I'd prefer that stay local.

ericswanson-dfinity: I was thinking all 8 lines of the construct that's duplicated.  I see what you did there (`.map(|(newline_index, _)| newline_index + 1);`), which is nice.  It does look like the almost-duplication in the error-raising part has introduced a bug, though (details to follow).

src/lib/apply-patch/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+        } else {
+            line_indices
+                .nth(start_line - 1)
+                .ok_or_else(|| MismatchError::NotEnoughLines {
+                    expected: start_line - 1,
+                    found: original_content.lines().count(),
+                })?
+        };
+        // start..end should be the byte range in `content` to be patched
+        let end = line_indices.nth(expected_lines.len() - 1).ok_or_else(|| {
```
ericswanson-dfinity, line 142: Should this be the following? This is in part why I was hoping for more DRY, though I don't feel too strongly about these 4(8) LOC.

```suggestion
        let end = line_indices.nth(start_line + expected_lines.len() - 1).ok_or_else(|| {
```

adamspofford-dfinity: No. `nth` is stateful.

ericswanson-dfinity: Okay, got it now.  The earlier version had this, which made me wonder:
```
.nth(start_line -1)
...
expected: start_line - 1
```
followed by
```
.nth(expected_lines.len() - 1)
...
expected: expected_lines.len() - 1,
```

adamspofford-dfinity: Yes, it was a bug I caught while updating it. Only affected the error message.

PR #3295, opened 2023-08-09
chore: --emulator parameter is deprecated

mraszyk:
This PR adds a warning that the `--emulator` is deprecated and will be discontinued soon.


ericswanson-dfinity (changes requested):
Another place to mention this: https://github.com/dfinity/sdk/blob/master/docs/cli-reference/dfx-start.md?plain=1#L21

mraszyk:
> Another place to mention this: https://github.com/dfinity/sdk/blob/master/docs/cli-reference/dfx-start.md?plain=1#L21

Thank you for pointing this out! Done.

ericswanson-dfinity (approved):

PR #3297, opened 2023-08-11
feat(extension): allow specifying number of values for subcommand's argument

smallstepman:
# Description

I want to suggest adding `values` (_naming?_) parameter (`.subcommands.<SUBCMD>.args.<ARG>.values`) in the `extension.json` manifest schema, which specifies how many values an arg can take. The parameter can be either positive integer or string, and can represent:
- "unlimited" amount of values with a format like `"values": "unlimited"`
- a specific amount of values with a format like `"values": 4`, meaning that the argument must take exactly 4 values (developers can specify `0`, meaning "arg doesn't take values", useful for "pure flags", e.g. `cargo build --workspace`)
- a range of allowable values with a format like `"values": "2..5"`, meaning that the argument can take anywhere from 2 to 5 values

Replaces `"multiple": <bool>` parameter.

This came up in https://github.com/dfinity/dfx-extensions/pull/38

# How Has This Been Tested?

- [x] unit
- [ ] e2e (TODO - will add them in a separate PR to avoid blocking the beta release on Friday evening)

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] I have edited the CHANGELOG accordingly.
- [ ] ~I have made corresponding changes to the documentation.~

PR #3299, opened 2023-08-11
chore: Release 0.15.0-beta.4

ericswanson-dfinity:
Incorporates https://github.com/dfinity/sdk/pull/3298

fix: return asset certificate for cert v1 even if encoding does not match


adamspofford-dfinity (approved):

smallstepman:
is there a chance we could also fit this in? https://github.com/dfinity/sdk/pull/3297

ericswanson-dfinity:
> is there a chance we could also fit this in? #3297

- it is too late for beta.4
- we should only merge after actually removing the `multiple` field

Recommendation:
1. create and push a tag on master `0.15.0-ext.0`.  This will create release binaries.
2. use that tag to update the dfx-extensions repo so use the new `values` field
3. merge a PR to master here that removes the `multiple` field
4. merge #3297 and the PR created in 3 to the 0.15.0 release branch for beta.5
5. update the dfx-extensions repo to use beta.5

smallstepman:
Sounds good!

PR #3338, opened 2023-08-29
fix: dfx deploy urls printed for asset canisters

krpeacock:
# Description

Currently, we're not printing frontend urls for most asset canisters, and it's inconvenient. Given a dfx config with no "frontend" block such as 

```json
"hello_frontend": {
      "dependencies": [
        "hello_backend"
      ],
      "source": [
        "src/hello_frontend/assets",
        "dist/hello_frontend/"
      ],
      "type": "assets"
    }
```

we get

image

After this change, we get 

Screenshot 2023-08-29 at 10 39 31 AM

# How Has This Been Tested?

Tested manually on a dfx new starter

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # UNRELEASED
 
+### fix: dfx deploy urls printed for asset canisters
+
 ### chore: --emulator parameter is deprecated and will be discontinued soon
 
 Added warning that the `--emulator` is deprecated and will be discontinued soon.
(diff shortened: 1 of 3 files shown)


ericswanson-dfinity (changes requested):
Could we have an e2e test that demonstrates the problem and shows that it's fixed?

I ran `dfx new` locally and `dfx deploy` did show the frontend URL, but after deleting the `frontend` key it did not, as described in this PR.

Here is a test that shows that given a project created with `dfx new`, `dfx deploy` does show the frontend url:
https://github.com/dfinity/sdk/blob/master/e2e/tests-dfx/frontend.bash#L26

This would mean adding something like this to remove the frontend key
```
jq 'del(.canisters.e2e_project_frontend.frontend)' dfx.json | sponge dfx.json
```
and re-running dfx deploy and the assertion

ericswanson-dfinity (approved):

PR #3351, opened 2023-09-07
feat: killall command

adamspofford-dfinity:
This adds `dfx killall`, which kills any processes that dfx starts, in case something is misbehaving. Implements SDK-1077.


ericswanson-dfinity (changes requested):
Per the contributing document and default PR description template for this repo, please add a document in https://github.com/dfinity/sdk/tree/master/docs/cli-reference for this command.

Please also update the documentation for `dfx stop`.

The overall goal should be to help someone looking at documentation (or `--help` output) understand in what circumstances would they use `dfx stop` vs `dfx killall`.

ericswanson-dfinity (changes requested):
`dfx start` will restart replica, icx-proxy, and the adapter processes if they are killed.  sandbox_launcher may do the same with canister_sandbox.   Could we kill these processes in some order, and/or loop here until no more processes are found, or something else?

src/dfx/src/commands/killall.rs
```diff
@@ -0,0 +1,34 @@
...
+    let self_pid = sysinfo::get_current_pid().map_err(Error::msg)?;
+    for (pid, proc) in info.processes() {
+        if proc.exe().file_name().is_some_and(|name| {
+            *pid != self_pid
+                && (name == "dfx"
+                    || name == "icx-proxy"
+                    || name == "replica"
+                    || name == "ic-ref"
+                    || name == "ic-btc-adapter"
+                    || name == "ic-https-outcalls-adapter"
```
ericswanson-dfinity, line 25: The ic-https-outcalls-adapter used to be called the ic-canister-http-adapter.  All of these processes that dfx launches are found under `get_bin_cache_root()`.  Might want to consider killing any process found within that.

CHANGELOG.md
```diff
@@ -952,6 +956,7 @@ Install a specific version of an extension, bypassing version checks.
 ### feat: Updated handling of missing values in state tree certificates
 
 The `Unknown` lookup of a path in a certificate results in an `AgentError` (the IC returns `Absent` for non-existing paths).
+>>>>>>> master
```
ericswanson-dfinity, line 959: leftover merge artifact

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #3390, opened 2023-09-26
refactor: standardize indentation for e2e test scripts

ericswanson-dfinity:
# Description

We have some e2e test scripts that use 4-space indentation, and others that have 2-space indentation.  Within some e2e test script files, we have some tests that use 4-space indentation and others that use 2-space indentation.

This PR changes all e2e test scripts to use 2-space indentation, for consistency and not having to wonder what to use when adding a new test.

# How Has This Been Tested?

CI will cover it

### DIFF
--- a/e2e/tests-dfx/assetscanister.bash
+++ b/e2e/tests-dfx/assetscanister.bash
@@ -3,42 +3,42 @@
 load ../utils/_
 
 setup() {
-    standard_setup
+  standard_setup
 
-    dfx_new
+  dfx_new
 }
 
 teardown() {
-    dfx_stop
+  dfx_stop
 
-    standard_teardown
+  standard_teardown
 }
 
 create_batch() {
-    reg="batch_id = ([0-9]*) : nat"
-    assert_command      dfx canister call e2e_project_frontend create_batch '(record { })'
-    # shellcheck disable=SC2154
-    [[ "$stdout" =~ $reg ]]
-    BATCH_ID="${BASH_REMATCH[1]}"
-    echo "$BATCH_ID"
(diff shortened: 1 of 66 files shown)


krpeacock:
I'm down with the consistency, but is there any way we could enforce this rule going forward with linting?

lwshang:
How do you format these bats scripts?
I just tried `shfmt` which doesn't work.

ericswanson-dfinity:
> I'm down with the consistency, but is there any way we could enforce this rule going forward with linting?

We could with [super-linter](https://github.com/super-linter/super-linter) if shfmt worked here, but it doesn't.  (And see draft PR https://github.com/dfinity/sdk/pull/3293 to add super-linter)

```
$ shfmt -i 2 e2e/tests-dfx/assetscanister.bash 
e2e/tests-dfx/assetscanister.bash:57:1: "}" can only be used to close a block
```

It would be nice, but it's beyond the scope of this PR to add CI enforcement.

lwshang:
Found a possible solution:
https://github.com/bats-core/bats-core/blob/f7defb94362f2053a3e73d13086a167448ea9133/docs/source/writing-tests.md#comment-syntax

`bats` has valid bash alternative syntax. Should we consider using it?

ericswanson-dfinity:
> Found a possible solution: https://github.com/bats-core/bats-core/blob/f7defb94362f2053a3e73d13086a167448ea9133/docs/source/writing-tests.md#comment-syntax
> 
> `bats` has valid bash alternative syntax. Should we consider using it?

Sure, but not as part of this PR.

lwshang (approved):

PR #3431, opened 2023-11-03
feat: Read install arg from file

bitdivine:
# Description
This PR adds support for providing install arguments from a file.

This is motivated by these issues:

* It is impossible to pass a large canister init arg as the length of shell arguments is limited by the terminal (which in turn may be limited by the operating system).
* For shorter arguments, it is more convenient to write `--argument-file somefile.hex` than `--argument "$(cat somefile.hex)"`.
* The `dfx canister call` method supports `--argument-file` for similar reasons.  It would be nice to be consistent.

# How Has This Been Tested?

e2e tests are included in the PR.

# Checklist:

### DIFF
--- a/docs/cli-reference/dfx-canister.md
+++ b/docs/cli-reference/dfx-canister.md
@@ -438,6 +438,7 @@ You can use the following optional flags with the `dfx canister install` command
 
 | Flag                  | Description                                                                                                                                                             |
 |-----------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `--argument-file`     | Specifies the file from which to read the argument to pass to the init method.  Stdin may be referred to as `-`.                                                        |
 | `--async-call`        | Enables you to continue without waiting for the result of the installation to be returned by polling the Internet Computer or the local canister execution environment. |
 | `--upgrade-unchanged` | Upgrade the canister even if the .wasm did not change.                                                                                                                  |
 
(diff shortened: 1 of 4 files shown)


e2e/tests-dfx/install.bash
```diff
@@ -182,7 +182,17 @@ teardown() {
...
+
+@test "installing with an argument in a file succeeds" {
+  TMPFILE="$(mktemp)"
+  echo '()' >"$TMPFILE"
+  assert_command dfx canister install e2e_project_backend --argument-file "$TMPFILE"
+}
+
 @test "installing multiple canisters with arguments fails" {
-  assert_command_fail dfx canister install --all --argument hello
+  assert_command_fail dfx canister install --all --argument '()'
```
bitdivine, line 196: This is so that the commands that succeed and fail are as similar as possible.

docs/cli-reference/dfx-canister.md
```diff
@@ -438,6 +438,7 @@ You can use the following optional flags with the `dfx canister install` command
 
 | Flag                  | Description                                                                                                                                                             |
 |-----------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `--argument-file`     | Specifies the file from which to read the argument to pass to the init method.  Stdin may be referred to as `-`.                                                        |
```
sesi200, line 441: > Stdin may be referred to as `-`.

This is not tested. Can you maybe add a test for that too?

sesi200 (approved):

docs/cli-reference/dfx-canister.md
bitdivine: Sure thing.  Done.

PR #3434, opened 2023-11-03
fix(ic-asset): .ic-assets.json `allow_raw_access` overrides 

smallstepman:
# Description

.ic-assets.json: "allow_raw_access" default overrides previous config entries even if not present in a match entry. This fixes the issue.

<details> <summary> Jira ticket description </summary> 

> Consider the following .ic-assets.json file.  
> ```
> [
> 	{
> 		"match": "**/favicons/**/*",
> 		"allow_raw_access": true
> 	},
> 	{
> 		"match": ".well-known",
> 		"ignore": false
> 	},
> 	{
> 		"match": "**/*",
> 		"headers": {
> 			"X-Frame-Options": "DENY",
> 			"X-Content-Type-Options": "nosniff",
> 			"Strict-Transport-Security": "max-age=31536000; includeSubDomains",
> 			"Referrer-Policy": "same-origin",
> 			"X-XSS-Protection": "1; mode=block"
> 		}
> 	}
> ]
> ```
> If there is a file in the favicons/ directory, we would expect its allow_raw_access setting to be true.
> 
> However, the "match": "**/*" section also matches for files in the favicons/ directory.  Even though the configuration does not specify a value for allow_raw_access, a default of false is assumed.
> 
> So given the above configuration, the actual property seen for allow_raw_access is false.
> 
> Note: the default value of allow_raw_access has changed from false to true in dfx 0.14.4 and later, so the problem won’t present exactly the same.
> 
> A fix is to reorder the configuration, to put the favicons directory match at the end:
> ```
> [
> 	{
> 		"match": ".well-known",
> 		"ignore": false
> 	},
> 	{
> 		"match": "**/*",
> 		"headers": {
> 			"X-Frame-Options": "DENY",
> 			"X-Content-Type-Options": "nosniff",
> 			"Strict-Transport-Security": "max-age=31536000; includeSubDomains",
> 			"Referrer-Policy": "same-origin",
> 			"X-XSS-Protection": "1; mode=block"
> 		}
> 	},
> 	{
> 		"match": "**/favicons/**/*",
> 		"allow_raw_access": true
> 	},
> ]
> ```
> However, note that there could be similar problems with other default settings.
> 
> 
> 
> Expected behavior: No matter the order of the above configuration, files in the favicons/ directory (in this example) should have allow_raw_access true.  Since the "match": "**/*" configuration section does not contain allow_raw_access, it should not affect the configuration of that value.
> 

</details>

Fixes https://dfinity.atlassian.net/browse/SDK-1238

# How Has This Been Tested?

unit test

# Checklist:


src/canisters/frontend/ic-asset/src/asset/config.rs
```diff
@@ -248,16 +248,13 @@ impl AssetConfig {
...
         if other.ignore.is_some() {
             self.ignore = other.ignore;
         }
-
         if other.enable_aliasing.is_some() {
             self.enable_aliasing = other.enable_aliasing;
         }
-
-        if other.allow_raw_access.is_some() {
+        if self.allow_raw_access.is_none() && other.allow_raw_access.is_some() {
```
ericswanson-dfinity, line 257: Since `ignore`, `enable_aliasing`, and `cache` use the same logic, will they have the same problem?

src/canisters/frontend/ic-asset/src/asset/config.rs
```diff
@@ -936,4 +933,82 @@ mod with_tempdir {
...
+            assets_config2
+                .get_asset_config(
+                    assets_dir2
+                        .join("nested/deep/the-next-thing.toml")
+                        .as_path(),
+                )
+                .unwrap()
+        };
+
+        assert_eq!(x.allow_raw_access, y.allow_raw_access);
```
ericswanson-dfinity, line 1011: it's true that they should be the same, but we also expect them both to be `false`.  
```suggestion
        assert_eq!(x.allow_raw_access, false);
        assert_eq!(y.allow_raw_access, false);
```

e2e/tests-dfx/assetscanister.bash
```diff
@@ -1484,38 +1484,33 @@ CHERRIES" "$stdout"
   "match": "nevermatchme",
   "cache": {
     "max_age": 2000
-  },
-  "allow_raw_access": true
```
smallstepman, line 1488: incorrectly introduced in #3369

src/canisters/frontend/ic-asset/src/asset/config.rs
```diff
@@ -66,19 +62,14 @@ pub struct AssetConfigRule {
     allow_raw_access: Option<bool>,
 }
 
-#[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq, Default)]
 enum Maybe<T> {
     Null,
+    #[default]
```
smallstepman, line 68: lint

src/canisters/frontend/ic-asset/src/asset/config.rs
```diff
@@ -427,6 +416,16 @@ mod rule_utils {
...
+            if let Some(allow_raw_access) = self.allow_raw_access {
+                s.push_str(&format!(
+                    "  - enable raw access: {}\n",
+                    if allow_raw_access {
+                        "enabled"
+                    } else {
+                        "disabled"
+                    }
+                ));
+            }
```
smallstepman, line 428: makes life easier during debugging

ericswanson-dfinity (changes requested):
Would you please write something in the PR description about _how_ this PR fixes the issue?  I see changes to use the `derivative` crate, but that looks like a refactor, not something that changes functionality.

Also please describe the problem that this solves in a little more detail in the PR description, since access to JIRA is not public.

smallstepman:
I pasted the content of jira ticket into pr description

ericswanson-dfinity (changes requested):
Not to be merged until we understand how/why this PR fixes the issue.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ Updated Motoko to [0.10.2](https://github.com/dfinity/motoko/releases/tag/0.10.2
 
 ### Frontend canister
 
+Fixed an issue where the `allow_raw_access` configuration value was being affected by the order of  declarations in the configuration file in `.ic-assets.json`.
```
ericswanson-dfinity, line 50: In the changelog, please describe the issue in enough detail that someone with a project, the behavior of which will change after this fix, will be able to recognize that fact.

ericswanson-dfinity:
> I pasted the content of jira ticket into pr description

I appreciate the effort, but unfortunately it's not sufficient in this case.

The JIRA ticket describes a problem. Its purpose is to describe that problem in enough detail that the assignee can understand what's being asked for, and hopefully reproduce the problem.  It doesn't describe the cause of the problem because it can't: the cause isn't known at the time of writing of the JIRA ticket.

The PR description should describe the change and the reasoning behind it.   The PR author should understand their change well enough to describe what they did and why.  The PR description should give reviewers an idea of what they are looking at.  A succinct yet complete PR description will also demonstrate to reviewers that the PR author understands their own change.

src/canisters/frontend/ic-asset/src/asset/config.rs
```diff
@@ -348,7 +337,7 @@ mod rule_utils {
         headers: Maybe<HeadersConfig>,
         ignore: Option<bool>,
         enable_aliasing: Option<bool>,
-        #[serde(default = "super::default_raw_access")]
+        #[derivative(Default(value = "Some(true)"))]
```
ericswanson-dfinity, line 340: An idea as to the underlying cause: this _removes_ the serde default, replacing it with an "impl Default" default.

What would you think about removing the `#[derivative(Default(value = "Some(true)"))]` as well?  At this level, we're interested in knowing what's in the config file rule, and None seems more accurate.

sa-github-api:
Dear @smallstepman,

In order to potentially merge your code in this open-source repository and therefore proceed with your contribution, we need to have your approval on DFINITY's [CLA](https://github.com/dfinity/cla/blob/main/CLA.md).

If you decide to agree with it, please visit this [issue](https://github.com/dfinity/cla/issues/442) and read the instructions there. Once you have signed it, re-trigger the workflow on this PR to see if your code can be merged.

— The DFINITY Foundation

ericswanson-dfinity:
Replaced by https://github.com/dfinity/sdk/pull/3635

PR #3437, opened 2023-11-08
chore: update to candid 0.9.10

ericswanson-dfinity:


chenyan-dfinity (commented):
Just to double check if this also updates the asset canister? There seems to be no cargo.lock changes, so we are already at 0.9.10?

ericswanson-dfinity:
Turns out we have already updated to 0.9.11.  This does affect the asset canister.  I built it and there were no changes.

PR #3443, opened 2023-11-15
feat: add upgrade arg to asset canister to overwrite permissions

sesi200:
# Description

Adding the option to (re)set all permissions using upgrade arguments. This is especially useful for SNSes that cannot make calls as the canister's controller because any calls triggered by vote come from the governance canister while the root canister is the only controller

Fixes [SDK-1290](https://dfinity.atlassian.net/browse/SDK-1290)

# How Has This Been Tested?

Added e2e

# Checklist:

[SDK-1290]: https://dfinity.atlassian.net/browse/SDK-1290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ericswanson-dfinity (approved):

docs/design/asset-canister-interface.md
```diff
@@ -159,6 +159,30 @@ The size of any chunk cannot exceed the message ingress limit.
...
+})
+```
+
+The methods `init` and `post_upgrade` are called automatically by the system after new code is installed in the canister.
+
+Both methods take the same argument type by definition. Therefore, to be able to have different arguments for the two cases, an enum is used to make the distinction.
+If `init` is called with the `Upgrade` variant or if `post_upgrade` is called with the `Init` variant the asset canister traps and thereby reverts the code changes.
+
+In `Upgrade`, the field `set_permissions` can be used to (re)set the list of principals with the listed permissions.
+Any of the permissions in `set_permissions` that are not `null` are set to the newly provided list of principals.
```
ericswanson-dfinity, line 183: Please update the service definition above and the text here to reflect the removal of the "double opt"

src/canisters/frontend/ic-certified-assets/assets.did
```diff
@@ -139,7 +139,25 @@ type ListPermitted = record { permission: Permission };
...
+  Upgrade: UpgradeArgs;
+};
+
+type InitArgs = record {};
+
+type UpgradeArgs = record {
+  set_permissions: opt SetPermissions;
+};
+
+/// Sets the list of principals with a certain permission for every permission that is set.
```
ericswanson-dfinity, line 153:
```suggestion
/// Sets the list of principals granted each permission.
```

PR #3473, opened 2023-12-07
feat: use cycles ledger to create canisters

sesi200:
# Description

This PR makes dfx use the cycles ledger to create canisters when no wallet is configured.

Fixes [SDK-1173](https://dfinity.atlassian.net/browse/SDK-1173)

# How Has This Been Tested?

Added e2e

# Checklist:

- [x] I have NOT edited the CHANGELOG accordingly because this will be gated behind a feature flag

[SDK-1173]: https://dfinity.atlassian.net/browse/SDK-1173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/dfx/src/lib/operations/cycles_ledger.rs
```diff
@@ -1,21 +1,35 @@
...
+use ic_utils::interfaces::management_canister::builders::CanisterSettings;
 use ic_utils::Canister;
 use icrc_ledger_types::icrc1;
+use icrc_ledger_types::icrc1::account::Subaccount;
 use icrc_ledger_types::icrc1::transfer::{BlockIndex, TransferError};
+use serde::Deserialize;
 use slog::{info, Logger};
+use thiserror::Error;
+
+/// Cycles ledger feature flag to turn off behaviour that would be confusing while cycles ledger is not enabled yet.
```
THLO, line 25:
```suggestion
/// Cycles ledger feature flag to turn off behavior that would be confusing while cycles ledger is not enabled yet.
```

THLO (approved):

src/dfx/src/lib/operations/cycles_ledger.rs
```diff
@@ -170,3 +184,89 @@ pub async fn send(
...
+#[context("Failed to create canister via cycles ledger.")]
+pub async fn create_with_cycles_ledger(
+    agent: &Agent,
+    with_cycles: Option<u128>,
+    from_subaccount: Option<Subaccount>,
+    settings: DfxCanisterSettings,
+    cycles_ledger_canister_id: Principal,
+) -> DfxResult<Principal> {
+    #[derive(CandidType, Clone, Debug)]
+    // TODO(FI-1022): Import types from cycles ledger crate once available
```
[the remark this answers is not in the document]
sesi200: No, it's just not done yet

e2e/tests-dfx/error_diagnosis.bash
```diff
@@ -44,15 +44,3 @@ teardown() {
   assert_match "not part of the controllers" # this is part of the error explanation
   assert_match "'dfx canister update-settings --add-controller <controller principal to add> <canister id/name or --all> \(--network ic\)'" # this is part of the solution
 }
-
-@test "Instruct user to set a wallet" {
```
ericswanson-dfinity, line 48: Until the feature is enabled, should this test continue to pass?

src/dfx/src/lib/operations/canister/create_canister.rs
```diff
@@ -76,7 +80,23 @@ pub async fn create_canister(
     let agent = env.get_agent();
     let cid = match call_sender {
         CallSender::SelectedId => {
-            create_with_management_canister(env, agent, with_cycles, specified_id, settings).await
+            if !CYCLES_LEDGER_ENABLED
+                || (!env.get_network_descriptor().is_ic
+                    && std::env::var("DFX_DISABLE_AUTO_WALLET").is_err())
```
ericswanson-dfinity, line 85: Suggestion for readability.  What do you think?
```suggestion
            let cycles_ledger_disabled = !CYCLES_LEDGER_ENABLED;
            let ic_network = env.get_network_descriptor().is_ic;
            let enable_auto_wallet = std::env::var("DFX_DISABLE_AUTO_WALLET").is_err();

            if cycles_ledger_disabled || !ic_network || enable_auto_wallet
```

src/dfx/src/lib/operations/cycles_ledger.rs
```diff
@@ -170,3 +184,89 @@ pub async fn send(
...
+            .unwrap(),
+        )
+        .call_and_wait()
+        .await
+        .map_err(|err| anyhow!(err))?;
+    let create_result = Decode!(
+        &result,
+        Result<CreateCanisterSuccess, CreateCanisterError>
+    )
+    .map_err(|_| anyhow!("Failed to decode response."))?;
```
ericswanson-dfinity, line 270: Should this be `.context` rather than discarding the decoding error?

src/dfx/src/lib/operations/canister/deploy_canisters.rs
```diff
@@ -113,14 +116,19 @@ pub async fn deploy_canisters(
...
-            .await?;
-            proxy_sender = CallSender::Wallet(*wallet.canister_id_());
-            &proxy_sender
+            .await
+            {
+                Ok(wallet) => {
+                    proxy_sender = CallSender::Wallet(*wallet.canister_id_());
+                    &proxy_sender
+                }
+                Err(_) => call_sender,
```
ericswanson-dfinity, line 130: Is it appropriate to use `call_sender` in all error cases, or just some of them?  For example, if `create_wallet` created the wallet, installed the code, but failed in `wallet_store_wallet_wasm`, should this swallow the error and use call_sender?  Or is there some smaller set of error conditions under which this should happen.

src/dfx/src/lib/operations/cycles_ledger.rs
```diff
@@ -170,3 +184,89 @@ pub async fn send(
...
+        GenericError { error_code: Nat, message: String },
+    }
+    #[derive(Deserialize, CandidType, Clone, Debug, PartialEq, Eq)]
+    struct CreateCanisterSuccess {
+        pub block_id: Nat,
+        pub canister_id: Principal,
+    }
+
+    let cycles = with_cycles.unwrap_or(CANISTER_CREATE_FEE + CANISTER_INITIAL_CYCLE_BALANCE);
+    let result = agent
```
ericswanson-dfinity, line 249: Please set `created_at_time`, so that it's an idempotent operation, and wrap this call in retry logic.  This will reduce the likelihood of burning cycles creating a canister but not storing the canister id.

ericswanson-dfinity (changes requested):

src/dfx/src/lib/operations/canister/create_canister.rs
sesi200: This is now 
```
let auto_wallet_disabled = std::env::var("DFX_DISABLE_AUTO_WALLET").is_ok();
let ic_network = env.get_network_descriptor().is_ic;
if CYCLES_LEDGER_ENABLED && (ic_network || auto_wallet_disabled)
```
WDYT?

src/dfx/src/lib/operations/cycles_ledger.rs
sesi200: I'm setting and logging the `created_at_time` now so the result will be recoverable. I don't think it makes sense to add a loop because there is no error for which a retry really makes sense (except for whatever the agent can retry on its own).

src/dfx/src/lib/operations/canister/create_canister.rs
ericswanson-dfinity: Are they equivalent?

Regardless, see if you can reduce it to three terms and a single logic type.

"A && B && C" is going to be easier to read and understand than "A && (B || D)"

src/dfx/src/lib/operations/cycles_ledger.rs
ericswanson-dfinity: The case where retry makes sense is the case where the replica executed the operation, but the client didn't find out about it.

src/dfx/src/lib/identity/wallet.rs
```diff
@@ -23,29 +22,46 @@ use ic_utils::interfaces::{ManagementCanister, WalletCanister};
 use slog::info;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum GetOrCreateWalletCanisterError {
+    #[error(
+        "No wallet cofigured for combination of identity '{identity}' and network '{network}'"
```
ericswanson-dfinity, line 30:
```suggestion
        "No wallet configured for combination of identity '{identity}' and network '{network}'"
```

src/dfx/src/lib/identity/wallet.rs
```diff
@@ -150,19 +165,19 @@ pub async fn create_wallet(
 /// While developing locally, this always creates a new wallet, even if `create` is false.
 /// This can be inhibited by setting the DFX_DISABLE_AUTO_WALLET env var.
 #[allow(clippy::needless_lifetimes)]
-#[context("Failed to get wallet canister caller for identity '{}' on network '{}'.", name, network.name)]
+// #[context("Failed to get wallet canister caller for identity '{}' on network '{}'.", name, network.name)]
```
ericswanson-dfinity, line 168:
```suggestion
```

src/dfx/src/commands/canister/create.rs
```diff
@@ -80,6 +82,23 @@ pub struct CanisterCreateOpts {
     /// Bypasses the Wallet canister.
     #[arg(long)]
     no_wallet: bool,
+
+    /// Transaction timestamp, in nanoseconds, for use in controlling transaction deduplication, default is system time.
+    /// https://internetcomputer.org/docs/current/developer-docs/integrations/icrc-1/#transaction-deduplication-
+    //TODO(SDK-1331): unhide
+    #[arg(long, hide = true)]
+    created_at_time: Option<u64>,
```
ericswanson-dfinity, line 90: The `--created-at-time` and `--all` options should conflict with each other, because `dfx canister create --all --created-at-time <t>` could try to assign the same canister id to more than one canister.

src/dfx/src/commands/deploy.rs
```diff
@@ -99,6 +101,23 @@ pub struct DeployOpts {
     /// Compute evidence and compare it against expected evidence
     #[arg(long, conflicts_with("by_proposal"))]
     compute_evidence: bool,
+
+    /// Transaction timestamp, in nanoseconds, for use in controlling transaction deduplication, default is system time.
+    /// https://internetcomputer.org/docs/current/developer-docs/integrations/icrc-1/#transaction-deduplication-
+    //TODO(SDK-1331): unhide
+    #[arg(long, hide = true)]
+    created_at_time: Option<u64>,
```
ericswanson-dfinity, line 109: The `--created-at-time` parameter should require that the `canister_name` parameter is also present.  It could also be allowed if the project has only one canister, but I'm not sure it's worth the complexity.

src/dfx/src/commands/deploy.rs
```diff
@@ -156,6 +175,14 @@ pub fn exec(env: &dyn Environment, opts: DeployOpts) -> DfxResult {
 
     let call_sender = CallSender::from(&opts.wallet)
         .map_err(|e| anyhow!("Failed to determine call sender: {}", e))?;
+    let created_at_time = opts.created_at_time.or_else(|| {
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_nanos() as u64;
+        info!(env.get_logger(), "created-at-time is {now}.");
```
ericswanson-dfinity, line 183: in the case where the user retries by running the command again (with the `--created-at-time option), they're going to need to know which canister to retry for, in case the canister settings differ.  (We might be able to get away with retrying only the first canister to be created, or we might not.)

src/dfx/src/lib/operations/cycles_ledger.rs
ericswanson-dfinity: Though it is possible to handle all of this by making the user retry their command (this time specifying `--created-at-time` and altering the command to create or deploy a single canister), in a case where we have an idempotent operation, in my opinion it's better to handle that retry in code.

As for determine when to retry, you are correct: it is difficult.  A guess is `AgentError::TimeoutWaitingForResponse() | AgentError::TransportError(_)`, here: https://github.com/dfinity/sdk/blob/master/src/dfx/src/lib/retryable.rs#L6.

Regarding "whatever the agent can retry on its own", the agent does not retry `TransportError` because it can't. 
 (See https://github.com/dfinity/agent-rs/blob/main/ic-agent/src/agent/http_transport/reqwest_transport.rs#L88 for an example, and up the call stack).  It can't because after sending the initial request, there exists a class of errors for which it can't know if the server received the request and is executing it or not, and the agent doesn't know if the request is idempotent or not.

Relying on the agent's retry mechanisms also means relying on those mechanisms to be correct.  They may or may not be.  For example, Agent's [fn wait](https://github.com/dfinity/agent-rs/blob/main/ic-agent/src/agent/mod.rs#L673) doesn't retry if [fn poll](https://github.com/dfinity/agent-rs/blob/main/ic-agent/src/agent/mod.rs#L632) returns an Err, and fn poll will return an error right away in the case of connection timeout when making the read_state call.

Given the above, what do you think?

sesi200: The idempotency argument makes sense. I'll use `retryable` for now and if we find a better heuristic we can always impove it

src/dfx/src/commands/canister/create.rs
```diff
@@ -102,14 +115,25 @@ pub async fn exec(
...
-        call_sender = &proxy_sender;
+        .await
+        {
+            Ok(wallet) => {
+                proxy_sender = CallSender::Wallet(*wallet.canister_id_());
+                call_sender = &proxy_sender;
+            }
+            Err(err) => {
+                if CYCLES_LEDGER_ENABLED {
+                    debug!(env.get_logger(), "No wallet configured.");
```
ericswanson-dfinity, line 131: should this look for `GetOrCreateWalletCanisterError::NoWalletConfigured`, rather than hiding any other error?

ericswanson-dfinity (approved):

PR #3476, opened 2023-12-11
fix: add localhost to asset canister's CSP

mraszyk:
This MR adds `http://localhost:*` as `connect-src` into the asset canister's CSP. This will enable browsing the asset canister at `http://<canister-id>.localhost:<port>` in most browsers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # UNRELEASED
 
+### fix: allow `http://localhost:*` as `connect-src` in the asset canister's CSP
+
+This will enable browsing the asset canister at `http://<canister-id>.localhost:<port>` in most browsers.
+
 ### fix: frontend code crashing when there is no canister ID
 
 ### feat: `dfx ledger top-up` also accepts canister names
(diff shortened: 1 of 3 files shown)


smallstepman (approved):

smallstepman:
I wonder if we should also add `https://localhost` - remember the safari issue that has been recently discussed @adamspofford-dfinity @krpeacock ?

krpeacock:
It wouldn't hurt

smallstepman:
extending e2e test with multiple schemas (`localhost:`, `127...`, `<can-id>.localhost:`) would also strengthen the reliability of FE canister for local deployments (future work necessary for testing mainnet deployments)
- https://github.com/dfinity/sdk/actions/runs/7166344405/job/19521096556#step:7:146
- https://github.com/dfinity/sdk/blob/master/.github/workflows/e2e.yml#L186-L190

@mraszyk lemme know if you want to contribute on that front as well

mraszyk:
> extending e2e test with multiple schemas (`localhost:`, `127...`, `<can-id>.localhost:`) would also strengthen the reliability of FE canister for local deployments (future work necessary for testing mainnet deployments)
> 
> * https://github.com/dfinity/sdk/actions/runs/7166344405/job/19521096556#step:7:146
> * https://github.com/dfinity/sdk/blob/master/.github/workflows/e2e.yml#L186-L190
> 
> @mraszyk lemme know if you want to contribute on that front as well

I wouldn't mind if you could please update the e2e tests.

mraszyk:
@smallstepman Do you want to extend e2e tests as part of this MR or do you have a separate ticket for that?

PR #3481, opened 2023-12-18
feat!: update `dfx cycles` commands with maninet `cycles-ledger` canister ID

smallstepman:
# Description

[The canister ID for `cycles-ledger` became known](https://forum.dfinity.org/t/improving-the-cycle-management-experience/19895/10), therefore, it is no longer required to pass it as an CLI argument.  

# How Has This Been Tested?

covered by CI

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # UNRELEASED
 
-### chore: update `dfx cycles` commands with maninet `cycles-ledger` canister ID
+### chore: update `dfx cycles` commands with mainnet `cycles-ledger` canister ID
 
 ### fix: frontend code crashing when there is no canister ID
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # UNRELEASED
 
+### chore: update `dfx cycles` commands with maninet `cycles-ledger` canister ID
+
 ### fix: frontend code crashing when there is no canister ID
 
 ### feat: `dfx ledger top-up` also accepts canister names
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 # UNRELEASED
 
-### chore: update `dfx cycles` commands with mainnet `cycles-ledger` canister ID
+### feat!: update `dfx cycles` commands with mainnet `cycles-ledger` canister ID
 
+The `dfx cycles` command no longer needs nor accepts the `--cycles-ledger-canister-id <canister id>` parameter.
 ### fix: allow `http://localhost:*` as `connect-src` in the asset canister's CSP
 
 This will enable browsing the asset canister at `http://<canister-id>.localhost:<port>` in most browsers.
--- a/src/dfx/src/lib/operations/cycles_ledger.rs
+++ b/src/dfx/src/lib/operations/cycles_ledger.rs
@@ -16,7 +16,15 @@ use slog::{info, Logger};
 const ICRC1_BALANCE_OF_METHOD: &str = "icrc1_balance_of";
 const ICRC1_TRANSFER_METHOD: &str = "icrc1_transfer";
 const SEND_METHOD: &str = "send";
-const CYCLES_LEDGER_CANISTER_ID: &str = "um5iw-rqaaa-aaaaq-qaaba-cai";
+const CYCLES_LEDGER_CANISTER_ID: Principal =
+    Principal::from_slice(&[0x00, 0x00, 0x00, 0x00, 0x02, 0x10, 0x00, 0x02, 0x01, 0x01]);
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 # UNRELEASED
 
+### chore: update `dfx cycles` commands with maninet `cycles-ledger` canister ID
```
ericswanson-dfinity, line 5:
```suggestion
### chore: update `dfx cycles` commands with mainnet `cycles-ledger` canister ID
```

src/dfx/src/lib/operations/cycles_ledger.rs
```diff
@@ -16,16 +16,16 @@ use slog::{info, Logger};
 const ICRC1_BALANCE_OF_METHOD: &str = "icrc1_balance_of";
 const ICRC1_TRANSFER_METHOD: &str = "icrc1_transfer";
 const SEND_METHOD: &str = "send";
+const CYCLES_LEDGER_CANISTER_ID: &str = "um5iw-rqaaa-aaaaq-qaaba-cai";
```
ericswanson-dfinity, line 19: Would you please handle this the same way we do MAINNET_CYCLE_MINTER_CANISTER_ID?  That way it's a `const Principal`, and doesn't have to be converted from a string.

ericswanson-dfinity (changes requested):
This will probably also require a change to assets/cycles-ledger/dfx.json and/or utils/cycles-ledger.bash in order to create the canister locally with the expected principal.

src/dfx/src/lib/operations/cycles_ledger.rs
smallstepman: I thought about it, but I was concerned about the fact that the canister ID becomes effectively almost non-searchable:
```rust
pub const MAINNET_CYCLE_MINTER_CANISTER_ID: Principal =
    Principal::from_slice(&[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x01, 0x01]);
```
, meaning, if I only knew the ID of the minter canister `mqygn-kiaaa-aaaar-qaadq-cai` (which can be found on forum, for example) and would like to search through DFINITY GitHub org to find all references in the code, I wouldn't find it. I also concluded the time to compute Principal from String should be negligible from the user's perspective. 

But there may be other good reasons to store it as a hex which I don't know about, and if so, then I'll annotate with a comment with the canister ID above.

smallstepman:
@ericswanson-dfinity would the following approach suffice:
```
  assert_command dfx deploy cycles-ledger --specified-id "um5iw-rqaaa-aaaaq-qaaba-cai"
  assert_command dfx deploy cycles-depositor --argument "(record {ledger_id = principal \"$(dfx canister id cycles-ledger)\"})" --with-cycles 10000000000000 --specified-id "ul4oc-4iaaa-aaaaq-qaabq-cai"
```

or should I deploy them separately from test's `dfx.json`, and reference them as remote canisters?

src/dfx/src/lib/operations/cycles_ledger.rs
ericswanson-dfinity: Part of the approach involves a test that verifies that the const Principal matches the expected value. See https://github.com/dfinity/sdk/blob/master/src/dfx/src/lib/ledger_types/mod.rs#L181 for an example with MAINNET_CYCLE_MINTER_CANISTER_ID.

It's not about execution time at all (which should be obvious).  It's about having a `const Principal` that doesn't need to go through .into() and can be passed to any method that takes &Principal, and also about consistency.  We shouldn't do it one way for MAINNET_LEDGER_CANISTER_ID, MAINNET_CYCLE_MINTER_CANISTER_ID, MAINNET_MOTOKO_PLAYGROUND_CANISTER_ID, DANK_PRINCIPAL, MAINNET_BITCOIN_CANISTER_ID and another way for CYCLES_LEDGER_CANISTER_ID.  (That said, there are two others that snuck by with const &str).

ericswanson-dfinity:
> @ericswanson-dfinity would the following approach suffice:

Yes, `dfx deploy --specified-id` is fine.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 # UNRELEASED
 
+### chore: update `dfx cycles` commands with mainnet `cycles-ledger` canister ID
+
```
ericswanson-dfinity, line 6:
```suggestion
### feat!: update `dfx cycles` commands with mainnet `cycles-ledger` canister ID

The `dfx cycles` command no longer needs nor accepts the `--cycles-ledger-canister-id <canister id>` parameter.
```

ericswanson-dfinity (approved):

PR #3510, opened 2024-01-17
feat: dfx deps: wasm_hash_url and loose the hash check

lwshang:
SDK-1345

# Description

## Loose the constraint of wasm_hash :

- If the hash of downloaded wasm doesn’t match the provided hash, dfx deps won’t abort. Instead, it will print a warning message.
- The hash can still benefit wasm cache invalidation. If a downloaded wasm has the matching hash, dfx deps can skip the downloading.

## Add another optional field wasm_hash_url in the pullable metadata.

- Providers like II can provide the hash through this URL instead of hard coding the hash directly.
- The hash of an alternative wasm module doesn’t have to be known beforehand. It can be modified later.
- For II, all flavors can be built in parallel. We will only need:
  - The PROD wasm has the wasm_url and wasm_hash_url in its pullable metadata.
  - The DEV build job pushes the wasm module and its hash through the URLs above.
  - The PROD and DEV build won’t intertwine with each other

# How Has This Been Tested?

e2e tests / deps.bash

# Checklist:


src/dfx-core/src/config/model/dfinity.rs
```diff
@@ -170,7 +170,15 @@ pub struct Pullable {
...
     /// Only define this if the on-chain canister wasm is expected not to match the wasm at wasm_url.
+    /// The hash can also be specified via a URL using the `wasm_hash_url` field.
+    /// If both are defined, the `wasm_hash_url` field will be ignored.
     pub wasm_hash: Option<String>,
+    /// # wasm_hash_url
+    /// Specify the SHA256 hash of the wasm module via this URL.
+    /// Only define this if the on-chain canister wasm is expected not to match the wasm at wasm_url.
+    /// The hash can also be specified directly using the `wasm_hash` field.
+    /// If both are defined, the `wasm_hash_url` field will be ignored.
+    pub wasm_hash_url: Option<String>,
```
ericswanson-dfinity, line 181: Should see a change to https://github.com/dfinity/sdk/blob/master/docs/dfx-json-schema.json

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

PR #3511, opened 2024-01-17
feat: Add large wasm support to `canister install`

adamspofford-dfinity:
This PR implements dfx support for dfinity/agent-rs#490, so `dfx deploy`/`dfx canister call` will silently start working on larger wasm modules. It also includes the replica upgrade for 072b2a6586c409efa88f2244d658307ff3a645d8 which enabled the chunked wasm feature.


ericswanson-dfinity (changes requested):

e2e/tests-dfx/install.bash
```diff
@@ -152,6 +152,16 @@ teardown() {
...
 }
 
+@test "can install >2MiB wasm" {
+  install_asset large_canister
+  dfx_start
+  dfx canister create --all
+  assert_command dfx build
+  assert_command dfx canister install --all
+  assert_command dfx canister info large
+  assert_match "Module hash: 0x$(sha2sum .dfx/local/canisters/large/large.wasm.gz | head -c 64)"
```
ericswanson-dfinity, line 162: As written, the line below expands to `assert_match "Module hash: 0x"` which passes no matter what:
- `sha2sum` doesn't exist (see `sha256sum` in test above)
- `large.wasm.gz` doesn't exist because the canister wasm isn't gzipped

bash doesn't notice these errors for two reasons:
- `set -o pipefail` is not in effect, and `head` returns 0
- the command expansion is inline.  The shell would notice the error in a line like `VAR="$(expand this)"`.  No idea why shellcheck doesn't flag this.

The form below still wouldn't error out with the original `sha2sum` command, since `set -o pipefail` is not set.  I have started a separate branch to explore setting this.

```suggestion
  HASH="$(sha256sum .dfx/local/canisters/large/large.wasm | head -c 64)"
  assert_match "Module hash: 0x$HASH"
```

ericswanson-dfinity (changes requested):

e2e/tests-dfx/install.bash
ericswanson-dfinity: The form in the above suggestion would now detect the error, now that https://github.com/dfinity/sdk/pull/3513 is merged.

ericswanson-dfinity (approved):

PR #3514, opened 2024-01-18
chore: deprecate `dfx wallet redeem-faucet-coupon` 

smallstepman:
# Description

Upcoming PR will introduce functionality allowing for redeeming the coupon straight into cycles-ledger (no wallet involved), making `wallet` part of the command a bit confusing. 

Additionally, we wish to steer people away from `dfx wallet` unless they really want to use one.

# How Has This Been Tested?

covered by CI

# Checklist:


docs/cli-reference/dfx-cycles.md
```diff
@@ -148,3 +149,56 @@ Send cycles to a canister by principal:
 ``` bash
 dfx cycles top-up bkyz2-fmaaa-aaaaa-qaaaq-cai 1000000000 --network ic
 ```
+
+## dfx wallet redeem-faucet-coupon
```
sesi200, line 153:
```suggestion
## dfx cycles redeem-faucet-coupon
```
same in the rest of the section

ericswanson-dfinity (commented):
It's great to notice that a command's location isn't going to make sense after some upcoming change, and to move it accordingly.

In the past, when we've removed a command, we've deprecated it first, and then removed it in a later version.  This makes it so developers can upgrade, and will have time to make changes after noticing a change like this.

What would you think about doing this here?

smallstepman:
Absolutely, that makes complete sense. In the meantime, the same has been discussed on Slack with @dfx-json & @sesi200; the current plan is to:
- put deprecation warning on `dfx wallet redeem-...` (by reworking this PR)
- add new functionality under `dfx cycles redeem` (by reworking the #3515)
- remove `dfx wallet` completely and turn it into an extension (future work)

src/dfx/src/commands/wallet/mod.rs
```diff
@@ -48,6 +48,9 @@ enum SubCommand {
     Custodians(custodians::CustodiansOpts),
     Deauthorize(deauthorize::DeauthorizeOpts),
     Name(name::NameOpts),
+    #[deprecated(
+        note = "This command is deprecated and will be removed in a future version. Eventually, the wallet canister will become an dfx extension."
```
ericswanson-dfinity, line 52: This should say what command they should use instead.  The deprecation warning doesn't need to mention the intention to move the functionality to an extension.

Also, we should display the deprecation warning at runtime.  As far as I can tell, this is only displayed at compile time.

```suggestion
        note = "This command is deprecated and will be removed in a future version. Use XXXXXX instead."
```


```bash
$ dfx-wip wallet redeem-faucet-coupon abc
WARN: Trying to redeem a wallet coupon on a local replica. Did you forget to use '--network ic'?
Redeeming coupon. This may take up to 30 seconds...
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (changes requested):
The changelog should also mention the deprecation

src/dfx/src/commands/wallet/mod.rs
smallstepman: right... 🤦

ericswanson-dfinity (commented):
Overall I would recommend putting this deprecation warning message in the PR that introduces the command that replaces it.

The goal is that someone who sees the deprecation warning can act on it, by replacing their call to the old command with a call to the new command.

sesi200:
Closing in favor of #3569

Issue #3003, opened 2023-03-04
Triggered "FIX ME! variant { ok; err : text } <: opt principal via special opt rule."


ByronBecker:
I was upgrading a canister and ran into this issue.

```
% dfx deploy cycleops                              
Deploying: cycleops
All canisters have already been created.
Building canisters...
Shrink WASM module size.
Installing canisters...
FIX ME! variant { ok; err : text } <: opt principal via special opt rule.
This means the sender and receiver type has diverged, and can cause data loss.
Upgrading code for canister cycleops, with canister ID sp3hj-caaaa-aaaaa-aaajq-cai
Deployed canisters.
```

Prior to the upgrade, this is the code that was changed.
```
  public shared ({ caller }) func stopTopupTimerCanister() : async ?Principal {
    if (not Admins.isAdmin(admins, caller)) return null

    switch (topupTimerActor) {
      case (?timer) {
        let principal = Principal.fromActor(timer);
        await ManagementCanisterActions.stopCanister(principal);
        ?principal
      };
      case null { null };
    };
  };
```

Became
```
  public shared ({ caller }) func stopTopupTimerCanister() : async Result.Result<(), Text> {
    if (not Admins.isAdmin(admins, caller)) return #err("Not authorized");

    switch (topupTimerActor) {
      case (?timer) {
        let principal = Principal.fromActor(timer);
        await ManagementCanisterActions.stopCanister(principal);
        #ok
      };
      case null { #err("No TopupTimer canister exists") };
    };
  };
```

sesi200:
Tracked internally under [SDK-1024](https://dfinity.atlassian.net/browse/SDK-1024)

Issue #2874, opened 2023-01-13
Candid file check failed: Cannot import


paulyoung:
I recently started breaking up my Candid file into smaller files to make it more manageable.

For example, I moved everything related to Canistergeek to `candid/canistergeek.did` and now do `import "canistergeek.did";` in `candid/codebase.did`.

My `candid_interface_compatibility` test (similar to [this one](https://github.com/dfinity/sdk/blob/89b4bd9ae0961089790c4982dcbb8461026c6bdb/src/canisters/frontend/ic-certified-assets/src/lib.rs#L288-L306)) supports this, and now also my `ic-repl` tests (thanks to @chenyan2002 and https://github.com/chenyan2002/ic-repl/pull/50)

However, `dfx generate` now fails.

I tried the following:

1. `dfx generate --verbose codebase_backend`

I expected to see this happen:

```
Generating type declarations for canister codebase_backend:
  frontend/declarations/codebase_backend/codebase_backend.did.js
```

Instead, this happened:

```
Error: (BFailed while trying to generate type declarations for 'codebase_backend'.
Caused by: (BFailed while trying to generate type declarations for 'codebase_backend'.
  Candid file check failed for frontend/declarations/codebase_backend/codebase_backend.did.
    Cannot import "canistergeek.did"
```

https://github.com/dfinity/sdk/blob/fc6d58955d02ba38e123159096c1df9e972757ce/src/dfx/src/util/mod.rs#L144-L152

### Meta

`dfx --version`:
```
dfx 0.12.0-beta.3
```

sesi200:
@chenyan-dfinity Is this something that we can fix by upgrading Motoko?

PR #3518, opened 2024-01-19
fix: fix some remaining candid issues

### LINKED ISSUE #3519: FIX ME special opt rule hit!
Just hit this message when deploying to a canister. I was purposely changing the type signature of my API (breaking change), and hit this message instead of signaling a dfx deploy warning.

```
FIX ME! vec record { text; AttributeValue } <: opt UserMetadata/1 via special opt rule.
This means the sender and receiver type has diverged, and can cause data loss.
```

chenyan-dfinity:
# Description

* Support `import` for local did file. Fix SDK-746 and #2874 
* Subtyping check reports special opt rule as error, so that developers can confirm the type in the upgrade. Fix #3519 #3003 

# How Has This Been Tested?

# Checklist:


src/dfx/src/lib/models/canister.rs
```diff
@@ -338,30 +338,40 @@ fn wasm_opt_level_convert(opt_level: WasmOptLevel) -> OptLevel {
...
 
 fn separate_candid(path: &Path) -> DfxResult<(String, String)> {
+    use candid::pretty::candid::{compile, pp_args};
+    use candid::types::internal::TypeInner;
+    use candid_parser::{
+        pretty_parse,
+        types::{Dec, IDLProg},
+    };
+    let did = dfx_core::fs::read_to_string(path)?;
+    let prog = pretty_parse::<IDLProg>(path.to_str().unwrap(), &did)?;
```
ericswanson-dfinity, line 348: Let's avoid unwrap if we can.  Something like this, or return an error.
```suggestion
    let prog = pretty_parse::<IDLProg>(&format!("{}", path.display()), &did)?;
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (changes requested):
Would you please add e2e tests that demonstrate what's being fixed here, and would detect regressions?

e2e/tests-dfx/metadata.bash
```diff
@@ -166,9 +166,9 @@ teardown() {
   dfx canister metadata prebuilt_custom_blank_build candid:service >from_canister.txt
   diff main.did from_canister.txt
 
-  # this canister has a build step, but it is an empty array, so dfx leaves the candid:service metadata as-is
-  dfx canister metadata prebuilt_custom_empty_build candid:service >from_canister.txt
-  diff main.did from_canister.txt
```
ericswanson-dfinity, line 171: Is there a reason to delete this part of this test?  This test is verifying existing behavior.

chenyan-dfinity: Fixed.

ericswanson-dfinity (approved):

PR #3546, opened 2024-01-29
feat: Candid assist

chenyan-dfinity:
# Description

Ask for user input when candid argument is not provided. A nice side effect is that `dfx deploy` now works with multiple canisters with init args (we cannot use `--argument` flag when multiple canisters require init args) 

# How Has This Been Tested?

e2e test

# Checklist:


adamspofford-dfinity:
> Not easy to write e2e test, because it's only activated when used in terminal.

DFX has a number of interactive tests in Expect format, in `e2e/assets/expect_scripts`. See [here](https://github.com/dfinity/sdk/blob/master/e2e/tests-dfx/identity.bash#L192) for an example, may require updating [this line](https://github.com/dfinity/sdk/blob/master/scripts/workflows/provision-linux.sh#L35).

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 # 0.17.0
 
+### feat: candid assist feature
```
ericswanson-dfinity, line 7: This should go in the "unreleased" section above

src/dfx/src/util/mod.rs
```diff
@@ -215,6 +240,30 @@ pub fn blob_from_arguments(
...
+pub fn gather_principals_from_env(env: &dyn Environment) -> BTreeMap<String, String> {
+    use ic_agent::Identity;
+    let mut res: BTreeMap<String, String> = BTreeMap::new();
+    if let Ok(mgr) = env.new_identity_manager() {
+        let logger = env.get_logger();
+        if let Ok(names) = mgr.get_identity_names(logger) {
+            let mut map = names
+                .iter()
+                .filter_map(|name| {
+                    let id = mgr.load_identity(name, logger).ok()?;
```
ericswanson-dfinity, line 252: This should probably skip any encrypted identities, as well as HSM identities, and explicitly add `env.get_selected_identity_principal()`.

This was result when I ran `dfx new candid_assist && cd candid_assist && dfx deploy`:

```
The UI canister on the "local" network is "b77ix-eeaaa-aaaaa-qaada-cai"
Installing code for canister candid_assist_backend, with canister ID br5f7-7uaaa-aaaaa-qaaca-cai
Please enter the passphrase for your identity: 
```

My default identity doesn't have a passphrase.

After I entered a passphrase, on a later identity, it popped up the keychain dialog.

This will also potentially be an issue for HSM identities.  

Maybe we should store the identity's principal in the identity.json file for each identity.

ericswanson-dfinity (changes requested):

src/dfx/src/util/mod.rs
chenyan-dfinity: Good point. Is there a method to know which identities are encrypted? This is a best-effort function, we can start by including only the non-encrypted identities, or just the canister id from dfx.json, and gradually add more principals if people find it useful.

ericswanson-dfinity: Not exactly.  `IdentityManager.get_identity_config_or_default` is probably the closest.  I don't think it will do the right thing for the anonymous identity, which will be included in the results of get_identity_names().

ericswanson-dfinity: You'd want to look for identities with either:
- no identity.json
- load identity.json as IdentityConfiguration and all members are None

ericswanson-dfinity (changes requested):
This happens when I run `dfx new hello && cd hello && dfx deploy`:

```
Installing code for canister hello_backend, with canister ID by6od-j4aaa-aaaaa-qaadq-cai
✔ Do you want to enter an optional value? · yes
  Enter a value for AssetCanisterArgs : variant { Upgrade : UpgradeArgs; Init : InitArgs }
  ✔ Select a variant · Init
    Enter a value for InitArgs : record {}
Sending the following argument:
(opt variant { Init = record {} })

Do you want to send this message? [y/N]
```

Expected behavior: deploying the default project doesn't block asking for any parameters.

(Also: it's prompting for the frontend canister, not the backend canister, but there's no indication of that at the "Do you want to enter an optional value?" prompt.)

chenyan-dfinity:
Should be good now. I move the install prompts at the top. It also helps with other warning messages. Restored the behavior of not asking for arguments when all types are optional.

e2e/tests-dfx/call.bash
```diff
@@ -39,6 +39,13 @@ teardown() {
   assert_eq '(record { c = "A"; d = "B" })'
 }
 
+@test "call without argument, using candid assistant" {
+  install_asset echo
+  dfx_start
+  dfx deploy
+  assert_command "${BATS_TEST_DIRNAME}/../assets/expect_scripts/candid_assist.exp"
+}
+
```
ericswanson-dfinity, line 48: Can we have one of these for `deploy` and `install` as well?

Here is the output I see when deploying to a canister declared as `actor class c(p: { x: Nat; y: Int;}) `:
```
Installing canisters...
Installing code for canister dd_backend, with canister ID bkyz2-fmaaa-aaaaa-qaaaq-cai
Enter field x : nat
  ✔ Enter a nat · 6
Enter field y : int
  ✔ Enter an int · 8
Sending the following argument:
(record { x = 6 : nat; y = 8 : int })

Do you want to send this message? [y/N]
```


Suggested wording:

```
Installing canisters...
Installing code for canister dd_backend, with canister ID bkyz2-fmaaa-aaaaa-qaaaq-cai

Canister dd_backend requires an initialization argument.
Enter field x : nat
  ✔ Enter a nat · 6
Enter field y : int
  ✔ Enter an int · 8

Initialization argument for dd_backend will be:
(record { x = 6 : nat; y = 8 : int })

Do you want to initialize canister dd_backend with this argument? [y/N]
```

src/dfx/src/util/mod.rs
```diff
@@ -224,6 +229,26 @@ pub fn blob_from_arguments(
...
                         args.to_bytes_with_types(env, &func.args)
+                    } else if is_terminal {
+                        use candid_parser::assist::{input_args, Context};
+                        let mut ctx = Context::new(env.clone());
+                        if let Some(env) = dfx_env {
+                            let principals = gather_principals_from_env(env);
+                            if !principals.is_empty() {
+                                let mut map = BTreeMap::new();
+                                map.insert("principal".to_string(), principals);
+                                ctx.set_completion(map);
```
ericswanson-dfinity, line 240: I tried deploying a canister with this declaration:
```
actor class c(x: Principal) {
```

It displayed the autocompletion options, but I can't actually use them:
```
Installing code for canister dd_backend, with canister ID bkyz2-fmaaa-aaaaa-qaaaq-cai
Auto-completions: __Candid_UI, alice, bob, dd_backend, dd_frontend, default,  mainnet, mainnet2
✘ Text must be in valid Base32 encoding.
? Enter a principal › 
```
The result I got when I tried typing any of the "autocompletions" was `Text must be in valid Base32 encoding`.  The up/down arrows did nothing, if they were supposed to.

Would you please start by adding an e2e test that demonstrates selection of a principal through autocompletion?

ericswanson-dfinity (changes requested):

src/dfx/src/util/mod.rs
chenyan-dfinity: You need to press `Tab` to replace the text with principal. The existing test already covers the auto-completion with principal type.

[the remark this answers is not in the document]
chenyan-dfinity: It's intended. We probably need to add anonymous manually in the identity export.

e2e/tests-dfx/call.bash
chenyan-dfinity: Done

src/dfx/src/util/mod.rs
```diff
@@ -224,6 +230,39 @@ pub fn blob_from_arguments(
...
+                                "This canister requires the following initialization argument."
+                            );
+                        } else {
+                            eprintln!("This method requires the following arguments.");
+                        }
+                        let args = input_args(&ctx, &func.args)?;
+                        eprintln!("Sending the following argument:\n{}\n", args);
+                        if is_init_arg {
+                            eprintln!(
+                                "Do you want to initialize canister with this argument? [y/N]"
```
ericswanson-dfinity, line 255: either "initialize the canister" or "initialize canister {name}" for correct grammar

e2e/tests-dfx/call.bash
ericswanson-dfinity: Here's what I see now, when deploying a canister with declaration `actor class c(x: Principal) {`

```
Installing code for canister dd_backend, with canister ID bkyz2-fmaaa-aaaaa-qaaaq-cai
This canister requires the following initialization argument.
Auto-completions: alice, anonymous, bob, dd_backend, dd_frontend, default, mainnet
? Enter a principal › 
```

Note "This canister requires the following initialization argument." but no following description of the argument.  The term "requires the following initialization argument" would have to refer to a "following" description of that argument.  It can't be used to refer only to the upcoming prompt for that argument.

chenyan-dfinity: Changed back to "This canister requires an initialization argument."

ericswanson-dfinity (approved):

PR #3563, opened 2024-02-01
fix(deps): content of wasm_hash_url can have extra fields than the hash

lwshang:
# Description

`wasm_hash_url` used to only accept the hash.

However, it is natural to point `wasm_hash_url` to the `<FILE>.sha256` file generated by `shasum` or `sha256sum` which consists of the hash and the file name.

Now when `dfx deps pull`, such content will be accept properly.

# How Has This Been Tested?

e2e deps.bash

# Checklist:


src/dfx/src/commands/deps/pull.rs
```diff
@@ -305,9 +305,15 @@ async fn get_hash_on_chain(
         let wasm_hash_content = download_file(&wasm_hash_url)
             .await
             .with_context(|| format!("Failed to download wasm_hash from {wasm_hash_url}."))?;
-        let wasm_hash_encoded = String::from_utf8(wasm_hash_content)
+        let wasm_hash_str = String::from_utf8(wasm_hash_content)
             .with_context(|| format!("Content from {wasm_hash_url} is not valid text."))?;
-        Ok(hex::decode(&wasm_hash_encoded)
+        // The content might contain the file name (usually from tools like shasum or sha256sum).
+        // We only need the hash part.
+        let wasm_hash_encoded = wasm_hash_str
```
ericswanson-dfinity, line 312: Is the intent to accept either of the following in the downloaded file?
<hash>
<filename> <hash>

It looks like the new code will only accept the second form.

If the intent is to accept both (which it seems so from the doc update0, then we should make the e2e test verify that both forms work.

ericswanson-dfinity (changes requested):

src/dfx/src/commands/deps/pull.rs
lwshang: The intent is to accept both. The e2e test was updated.

ericswanson-dfinity (approved):

PR #3581, opened 2024-02-08
chore: Update agent-rs to 0.33.0

adamspofford-dfinity:
This includes the compatibility change to the chunked wasm API in dfinity/agent-rs#512


ericswanson-dfinity (commented):
Is there a reason we wouldn't cut an agent releaae 0.33.0 for this?

adamspofford-dfinity:
Good point.

ericswanson-dfinity (approved):

PR #3586, opened 2024-02-12
chore: updating changelog

krpeacock:


ericswanson-dfinity (approved):

krpeacock:
@ericswanson-dfinity what should I do about the failing macos test? Do we need to add a section to `provision-darwin` pinning the rust version?

PR #3613, opened 2024-02-22
fix: better error message for bitcoin/ecdsa `dfx canister call`

adamspofford-dfinity:


Cargo.toml
```diff
@@ -21,11 +21,11 @@ license = "Apache-2.0"
 [workspace.dependencies]
 candid = "0.10.3"
 candid_parser = "0.1.3"
-ic-agent = "0.33.0"
+ic-agent = { git = "https://github.com/dfinity/agent-rs.git", rev = "d537a790a1d4d12f3882f5443a5c03d70ef9f4d8" }
```
ericswanson-dfinity, line 24: We've been asked to make a new agent release with this functionality anyway, so please release an 0.34.0 agent, unless I'm missing something

adamspofford-dfinity: I think you have gotten the feature wrong - it is for dfinity/agent-rs#520. And I am pretty sure that the need for a release there was just so it could be imported in dfx.

adamspofford-dfinity: In general I don't want the agent-rs release cadence to become 'whenever a new not-necessarily-released-yet feature gets merged into dfx'

ericswanson-dfinity (approved):

PR #3616, opened 2024-02-23
fix: no more disappearing canister ids from .env files

ericswanson-dfinity:
# Description

Made it so `dfx deploy` and `dfx canister install` will always write canister ids for
all canisters in the project to the .env file, even if they aren't being deployed/installed or
a dependency of a canister being deployed/installed.

Fixes https://dfinity.atlassian.net/browse/SDK-1369

# How Has This Been Tested?

Updated an e2e test (dfx build) and added two more (dfx deploy, dfx canister install)

# Checklist:

### DIFF
--- a/e2e/tests-dfx/build.bash
+++ b/e2e/tests-dfx/build.bash
@@ -96,6 +96,14 @@ teardown() {
   assert_match "syntax error"
 }
 
+@test "build --check fails on invalid motoko" {
+  install_asset invalid
+  dfx_start
+  # --check does not need create --all
+  assert_command_fail dfx build --check
+  assert_match "syntax error"
+}
+
 @test "build supports relative imports" {
   install_asset import
   dfx_start
--- a/e2e/tests-dfx/build.bash
+++ b/e2e/tests-dfx/build.bash
@@ -61,8 +61,6 @@ teardown() {
 
 @test "build uses default build args" {
   install_asset default_args
-  dfx_start
-  dfx canister create --all
   assert_command_fail dfx build --check
   assert_match "unknown option"
   assert_match "compacting-gcX"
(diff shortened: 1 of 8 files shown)


e2e/tests-dfx/build.bash
```diff
@@ -96,6 +96,14 @@ teardown() {
   assert_match "syntax error"
 }
 
+@test "build --check fails on invalid motoko" {
+  install_asset invalid
+  dfx_start
+  # --check does not need create --all
```
lwshang, line 102:
```suggestion
```

`dfx build --check` needs not `dfx start`.
There are three other tests above also only run `dfx build` with `--check`.

e2e/tests-dfx/dotenv.bash
```diff
@@ -51,15 +51,48 @@ teardown() {
...
+  dfx_start
+  dfx canister create --all
+  dfx build e2e_project_frontend
+  # set a post-install script so the install will create a .env file
+  jq '.canisters.e2e_project_frontend.post_install="echo post install"' dfx.json | sponge dfx.json
+
+  writes_all_environment_variables_to_env canister install
+}
+
+writes_all_environment_variables_to_env() {
```
lwshang, line 78:
```suggestion
assert_dotenv_contains_all_variables_after_command() {
```

I was confused when I see no assertions in the new tests above.
Then I found out that this method make assertions inside.

lwshang (approved):
LGTM. Only two minor comments about the tests.

e2e/tests-dfx/build.bash
ericswanson-dfinity: I'm going to add a test that still calls `dfx start` and `dfx canister create --all` and verifies `dfx build --check`.  The reason is that there are slightly different code paths involved.

An earlier commit on this PR broke `dfx build --check` when there are no canister ids, but not when canister ids are known.

PR #3617, opened 2024-02-23
fix(generate): Rust canister source candid wrongly deleted

lwshang:
SDK-1380

# Description

Fixed a bug where `dfx generate` would delete a canister's source candid file if the `declarations.bindings` in `dfx.json` did not include "did".

Reported in https://github.com/nathanosdev/sdk_rust_candid_issue

## Explanation

`dfx generate`, in brief, gets the candid file of a canister and generate other bindings based on this .

The incorrect implementation required all impl of `CanisterBuilder` to have a `generate_idl()`  method. And the `generate_idl()` method should copy the source candid file into `declarations/` dir. After generating all bindings, `dfx generate` will remove the candid file (should be `declarations/<canister>.did`) if `declarations.bindings` doesn't contains "did".

The `RustCanisterBuilder` didn't follow the requirement of `generate_id()`. Instead, its `generate_id()` returns the source candid file path directly. And this source candid file got deleted.

This PR simplifies the whole implementation. The bindings are generated from the source candid files directly. There is no logic of deleting candid files.

# How Has This Been Tested?

e2e/generate.bash

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ Also the warning of hash mismatch is removed since it scares users and users can
 
 ### fix(generate): Rust canister source candid wrongly deleted
 
+Fixed a bug where `dfx generate` would delete a canister's source candid file if the `declarations.bindings` in `dfx.json` did not include "did".
+
 # 0.17.0
 
 ### feat: new starter templates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ The `dfx upgrade` command now prints a message directing the user to install dfx
 
 Also the warning of hash mismatch is removed since it scares users and users can't fix it locally.
 
+### fix(generate): Rust canister source candid wrongly deleted
+
 # 0.17.0
 
 ### feat: new starter templates
--- a/src/dfx/src/lib/builders/mod.rs
+++ b/src/dfx/src/lib/builders/mod.rs
@@ -134,12 +134,12 @@ pub trait CanisterBuilder {
         if bindings.is_empty() {
             eprintln!("`{}.declarations.bindings` in dfx.json was set to be an empty list, so no type declarations will be generated.", &info.get_name());
             return Ok(());
-        } else {
-            eprintln!(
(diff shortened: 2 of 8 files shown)


ericswanson-dfinity (changes requested):
Would you please describe in the PR description what the PR solves and how you solved it?

- The repo is public, but access to JIRA is limited.
- We don't have any control over the linked repo, which the owner might delete after we fix the issue.

The title is close to describing the problem, but doesn't say when the problem arises.  The description could say something about the declarations not including "did".

Sometimes the solution is clear between reading the problem description and looking at the PR, but I'm having some trouble following.  It seems like you're simplifying how this all works, so a few words might help.

src/dfx/src/lib/builders/mod.rs
```diff
@@ -202,29 +208,26 @@ pub trait CanisterBuilder {
...
-                .strip_prefix(info.get_workspace_root())
-                .unwrap_or(&generated_idl_path);
-            eprintln!("  {}", &relative_idl_path.display());
+        // Candid
+        if bindings.contains(&"did".to_string()) {
+            let output_did_path = generate_output_dir
+                .join(info.get_name())
+                .with_extension("did");
+            dfx_core::fs::copy(&did_from_build, &output_did_path)?;
+            eprintln!("  {}", &output_did_path.display());
```
ericswanson-dfinity, line 217: Any need for this here?
```
dfx_core::fs::set_permissions_readwrite(&output_did_path)?;
```

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ The `dfx upgrade` command now prints a message directing the user to install dfx
 
 Also the warning of hash mismatch is removed since it scares users and users can't fix it locally.
 
+### fix(generate): Rust canister source candid wrongly deleted
+
```
ericswanson-dfinity, line 52:
```suggestion

Fixed a bug where `dfx generate` would delete a canister's source candid file
if the declaration bindings did not include "did".

```

ericswanson-dfinity (approved):

PR #3627, opened 2024-02-29
chore: use the right Subaccount type in dfx cycles commands

sesi200:
# Description

`dfx cycles` uses the ICP ledger `Subaccount` type instead of the `icrc1` variant, which is wrong

# How Has This Been Tested?

Should not produce any functional changes. Covered by e2e

# Checklist:

### DIFF
--- a/src/dfx/src/commands/cycles/approve.rs
+++ b/src/dfx/src/commands/cycles/approve.rs
@@ -1,11 +1,11 @@
-use crate::lib::environment::Environment;
 use crate::lib::error::DfxResult;
-use crate::lib::nns_types::account_identifier::Subaccount;
 use crate::lib::operations::cycles_ledger;
 use crate::lib::root_key::fetch_root_key_if_needed;
 use crate::util::clap::parsers::cycle_amount_parser;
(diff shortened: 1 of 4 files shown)


src/dfx/src/commands/cycles/approve.rs
```diff
@@ -1,11 +1,11 @@
-use crate::lib::environment::Environment;
 use crate::lib::error::DfxResult;
-use crate::lib::nns_types::account_identifier::Subaccount;
 use crate::lib::operations::cycles_ledger;
 use crate::lib::root_key::fetch_root_key_if_needed;
 use crate::util::clap::parsers::cycle_amount_parser;
+use crate::{lib::environment::Environment, util::clap::parsers::icrc_subaccount_parser};
```
lwshang, line 5: It might be the IDE that weirdly group these items. It's better to put them into separate lines and let `cargo fmt` to order properly.
Don't forget to fix other files below as well.

lwshang (approved):
LGTM. One minor issue about the `use` items grouping.

PR #3646, opened 2024-03-06
fix: [IC-272] use query call on management canister fetch_canister_logs method

maksymar:
# Description

The management canister method `fetch_canister_logs` can only be called as a query, not an update call, so this PR fixes the issue.

# How Has This Been Tested?

Built and run dfx locally to verify a rust canister collects the logs and they can be retrieved by using `dfx canister logs <canister_id>`. 

Output example:
image

# Checklist:

- [x] NO NEED for the changes to the documentation.


maksymar:
> Can you add a test for this?

Done, PTAL.

lwshang (approved):

PR #3653, opened 2024-03-07
chore: improve release procedure

lwshang:
# Description

- Rewrote the release guide
    - Removed the "Overview" at the top because GitHub UI has built-in toc support for Markdown.
    - Grouped steps into stages
    - Added samples for all PR steps
- Avoid building from clean twice in the release script
    - The second build was only for updating Cargo.lock. So replaced with a faster `cargo update -p dfx` command.
- The "publish" workflow will create the release with "Pre-release" set (and "Latest" not set)
    - We had to change the flags of each beta releases
    - We should only make the final release "Latest" after promoting it
- The "broadcast-frontend-hash" workflow will create the PR for motoko-playground in "draft" mode
    - All PRs created before promotion should be in draft mode to avoid accidental merge

# Checklist:


ericswanson-dfinity (approved):
Thanks!

PR #3660, opened 2024-03-14
chore: downgrade Bitcoin Canister back to 2023-10-13 (revert #3593)

lwshang:
Reverts dfinity/sdk#3593

Bitcoin Canister release/2024-01-22 doesn't work with application subnet. (It drains all the cycles on purpose which is fine only on the system subnet).

The bitcoin canister will be fixed in a few weeks.
Before that, we need to downgrade it to the previous version.

To detect such break in the future, I added two bitcoin.bash e2e tests:
* check if the canister has enough cycles balance.
* check if the management canister Bitcoin API can be called.
  * When testing the two query API, I found out that their effective_canister_id were set incorrectly. Fixed in this PR.
  * A `bitcoin.did` is added for the test. It was refined from the `ic.did` in the spec and added the `regtest` network variant.

Also filled the missing changelog entry in v0.18.0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ target/
 
 # DB repo for cargo-audit
 advisory-db/
+
+# mops
+**/.mops/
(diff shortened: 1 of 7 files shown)


ericswanson-dfinity (approved):
This would be a good opportunity to also transfer over the "bitcoin methods succeed" test from https://github.com/dfinity/cdk-rs/blob/main/examples/management_canister/tests/basic.bats, since it should be able to pass here now

ericswanson-dfinity (approved):

PR #3663, opened 2024-03-15
feat: canister call accepts --output json

ericswanson-dfinity:
# Description

this is the same as dfx canister call ... | idl2json

See: https://github.com/dfinity/idl2json

Fixes: https://dfinity.atlassian.net/browse/SDK-1370

# How Has This Been Tested?

Added e2e tests

# Checklist:

### DIFF
--- a/src/dfx/src/util/mod.rs
+++ b/src/dfx/src/util/mod.rs
@@ -89,7 +89,7 @@ pub fn print_idl_blob(
             if output_type == "idl" {
                 println!("{:?}", result?);
             } else if output_type == "json" {
-                let json = convert_all(&result?, &Idl2JsonOptions::default())?;
+                let json = convert_all(&result?)?;
                 println!("{}", json);
             } else {
                 println!("{}", result?);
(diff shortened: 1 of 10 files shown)


src/dfx/src/util/mod.rs
```diff
@@ -87,6 +88,9 @@ pub fn print_idl_blob(
             }
             if output_type == "idl" {
                 println!("{:?}", result?);
+            } else if output_type == "json" {
+                let json = convert_all(&result?, &Idl2JsonOptions::default())?;
```
lwshang, line 92: It seems that we only have one usage of `convert_all` here.
So the `idl2json_options` is always set with `Idl2JsonOptions::default()`.
Then in `convert_one`:
```
    (if idl2json_options.compact {
        serde_json::to_string
    } else {
        serde_json::to_string_pretty
    })(&json_value)
```
It always go into the second branch (`idl2json_options.compact == false`).

What about removing the `idl2json_options` argument from both `convert_one` and `convert_all`? The implementation will be simplified a lot.

Or do we want to let the user be able to set `compact` in the future? If so then it makes sense to keep the current implementation.

ericswanson-dfinity: Yeah, I considered this too..  I'll make the change

lwshang (approved):

PR #3667, opened 2024-03-18
chore: Update agent to cc675b0

adamspofford-dfinity:


ericswanson-dfinity (changes requested):
Please add the following to the changelog, for a change that came along with the update to agent-rs.  I was unable to update dfx and include this changelog, because of breaking changes on agent-rs main.

```
fix: Output of dfx ping is now valid JSON. Added commas in between fields, and newlines to improve formatting.
```

ericswanson-dfinity (approved):

PR #3670, opened 2024-03-20
feat: add tech_stack to the Canister Metadata Standard

lwshang:
SDK-1406

# Description

Canister authors can opt in to display the tech stack of the canister.
Providing a standard format of such information makes it easier to build tools like a Canister Explorer.
Check https://github.com/dfinity/sdk/blob/lwshang/sdk-1406/docs/concepts/tech-stack.md for more details.

The design was inspired by [ProducersSection](https://github.com/WebAssembly/tool-conventions/blob/main/ProducersSection.md) of the wasm standard.

This [forum thread](https://forum.dfinity.org/t/rfc-tech-stack-canister-metadata-standard-extension/28692/8) has the community feedbacks that drive the evolution of this feature.

# How Has This Been Tested?

e2e test in metadata.bash

# Checklist:


src/dfx/src/lib/builders/mod.rs
```diff
@@ -324,27 +324,49 @@ fn ensure_trailing_newline(s: String) -> String {
...
+    let (first, rest) = match command.split_once(' ') {
+        Some((first, rest)) => (first, rest),
+        None => (command, ""),
+    };
+
+    let canonicalized = dfx_core::fs::canonicalize(&cwd.join(first))
+        .or_else(|_| which::which(first))
+        .map_err(|_| anyhow!("Cannot find command or file {first}"))?;
+    let canonicalized_command = [
+        canonicalized.to_string_lossy().to_string(),
```
ericswanson-dfinity, line 348: Can we avoid passing this path through to_string_lossy()?  The result of `which()` in particular doesn't necessarily have to pass through losslessly. I think `cmd.arg(canonicalized).arg(rest)` will achieve this.

src/dfx/src/lib/builders/mod.rs
```diff
@@ -324,27 +324,49 @@ fn ensure_trailing_newline(s: String) -> String {
...
+    command: &str,
+    vars: &[Env<'_>],
+    cwd: &Path,
+    io_inherit: bool,
+) -> DfxResult<Vec<u8>> {
+    // No commands, noop.
+    if command.is_empty() {
+        return Ok(vec![]);
+    }
+    let (first, rest) = match command.split_once(' ') {
```
ericswanson-dfinity, line 339: Since this no longer calls shell_words::split, it won't handle spaces in the command name. (Of course, no one would do this, but the call to shell_words::split handled this before, so this could be a breaking change)
```
"build": "command\\ with\\ spaces the parameters" 
```
or
```
"build": "./some\\ commands/something the parameters."
```

```

File: src/dfx/src/lib/builders/mod.rs [hunk 3]
```diff
@@ -324,27 +324,49 @@ fn ensure_trailing_newline(s: String) -> String {
...
-        .or_else(|_| which::which(command_name))
-        .map_err(|_| anyhow!("Cannot find command or file {command_name}"))?;
-    let mut cmd = Command::new(canonicalized);
-
-    cmd.args(arguments)
-        .current_dir(cwd)
-        .stdout(Stdio::inherit())
-        .stderr(Stdio::inherit());
-
+/// Run a command and return its output.
```
ericswanson-dfinity, line 327: Suggestion (you can probably think of better names):
```
pub fn call_command(command, vars, cwd) -> DfxResult<()> {
    run_command(command, vars, cwd, false).map(|_|())
}

pub fn command_output(command, vars, cwd) -> DfxResult<Vec<u8>> {
    run_command(command, vars, cwd, true)
}

fn run_command(command, vars, cwd, return_output) -> DfxResult<Vec<u8>> {
...
}
```

ericswanson-dfinity (changes requested):

src/dfx/src/lib/builders/mod.rs
lwshang: I refactored this part a bit.

After simplification, there are two cases for the `command`:
- It's a file name relative to `cwd`, which may contains whitespace
  - canonicalize the path and execute it directly.
- It's a shell command which may have arguments
  - simply run `sh -c <command>` which allow pipes

docs/concepts/tech-stack.md
```diff
@@ -0,0 +1,132 @@
...
+Each custom field must define a `"field"` name.
+
+The value of the field can be defined in two ways:
+
+- `"value"`: This defines the value directly.
+- `"value_command"`:
+  - This should be a CLI command.
+  - The command will be run in the workspace root (the dir contains `dfx.json`). 
+  - The stdout should be a valid UTF-8 string.
+  - The stdout will be stripped to get the version.
```
ericswanson-dfinity, line 132: This part could apply to any custom field, not just version

docs/concepts/tech-stack.md
```diff
@@ -0,0 +1,132 @@
...
+
+```json
+{
+  "type": "motoko",
+  "main": "main.mo",
+  "tech_stack": {
+    "cdk": [
+      {
+        "name": "ic-cdk",
+        "custom_fields": [
```
ericswanson-dfinity, line 74: I think we should give the team some time to review the structure of this data

ericswanson-dfinity: I get a bit lost with the fact that `version` down in the custom_fields array turns up as a sibling to `name`

lwshang: I added a custom field to the `bitcoin` item: `"address": "bcrt1qfe264m0ycx2vcqvqyhs0gpxk6tw8ug6hqeps2d"`.

`"version"` and `"address"` are both custom fields.

I hope this another example beyond `"version"` can make the intention more clear.

docs/concepts/tech-stack.md
```diff
@@ -0,0 +1,116 @@
...
+
+- The command will be executed in the workspace root directory, which contains the `dfx.json` file.
+- The stdout should be a valid UTF-8 string.
+- The field value will be obtained by trimming the stdout, removing any leading and trailing whitespace.
+
+## Q&A
+
+### Who should set `tech_stack`?
+
+In the short future, CDK will be able to set `tech_stack` without requiring extra configuration in `dfx.json`.
```
ericswanson-dfinity, line 110:
```suggestion
In the near future, CDK will be able to set `tech_stack` without requiring extra configuration in `dfx.json`.
```

PR #3692, opened 2024-04-02
feat: add "dfx completion" command

ghost:
# Description

Adds `dfx completion` command to generate shell completion scripts.

The CLI is similar to [kubectl completion](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_completion/)

Fixes https://dfinity.atlassian.net/browse/SDK-1528

# How Has This Been Tested?

Added e2e tests which spot-check the output for expected commands.

Tested locally on osx with zsh and bash.

With zsh, these both work:
```
$ source <(dfx completion zsh)
$ source <(dfx completion) # uses bash completion, which works but is not as feature-rich
```

On OSX at least, there is some issue with source from stdin, but source from a file works:
```
$ bash

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
bash-3.2$ source <(dfx-wip completion)
bash-3.2$ thread 'main' panicked at /Users/ericswanson/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_complete-4.5.1/src/shells/bash.rs:23:9:
Failed to write to generated file
bash-3.2$ dfx-wip completion >a
bash-3.2$ source a
bash-3.2$ 
```

# Checklist:

- [z] I have made corresponding changes to the documentation.

### DIFF
--- a/docs/cli-reference/dfx-completion.mdx
+++ b/docs/cli-reference/dfx-completion.mdx
@@ -25,7 +25,7 @@ You can use the following option with the `dfx completion` command.
 
 | Option                | Description                        -|
 |-----------------------|-------------------------------------|
-| `--bin-name <name>`   | Name of the binary.  Default `dfx`. |
+| `--bin-name <name>`   | Name of the command. Only needed if referring to dfx by another name, such as with an alias. Default `dfx`. |
 
 ## Examples
 
--- a/src/dfx/src/commands/completion.rs
+++ b/src/dfx/src/commands/completion.rs
@@ -9,7 +9,7 @@ use clap_complete::Shell;
 /// Generate a shell completion script.
 #[derive(Parser)]
 pub struct CompletionOpts {
-    /// The name of the binary
+    /// The name of the command. Only needed if referring to dfx by another name, such as with an alias.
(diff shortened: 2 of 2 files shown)


docs/cli-reference/dfx-completion.mdx
```diff
@@ -0,0 +1,52 @@
...
+The values supported for SHELL are `bash`, `elvish`, `fish`, `powershell`,
+and `zsh`. The default value is `bash`.
+
+## Options
+
+You can use the following option with the `dfx completion` command.
+
+| Option                | Description                        -|
+|-----------------------|-------------------------------------|
+| `--bin-name <name>`   | Name of the binary.  Default `dfx`. |
```
lwshang, line 28: This option made me confused.

"Can `dfx completion` generate completion for some other binaries other than `dfx`?"

After checking the docs of `clap_complete::generate()`, I finally realized that this option is for the case that a user may install the `dfx` with a different name.

I'm wondering if we can improve the explanation. Or what about rename the option to something like `--rename-dfx`?

lwshang (approved):

PR #3697, opened 2024-04-04
fix: dfx only loads dfx.json for commands that need it

ericswanson-dfinity:
# Description

For example, this will work now:
```bash
echo garbage >dfx.json && dfx identity get-principal
```

This is meant to support reloading the config file in the future, which is why it uses RefCell and not OnceCell.

Fixes https://dfinity.atlassian.net/browse/SDK-1559

# How Has This Been Tested?

Added an e2e test

# Checklist:

### DIFF
--- a/e2e/tests-dfx/usage.bash
+++ b/e2e/tests-dfx/usage.bash
@@ -46,5 +46,5 @@ teardown() {
 
 @test "does not unconditionally read dfx.json" {
   echo "garbage" >dfx.json
-  dfx identity get-principal
+  assert_command dfx identity get-principal
 }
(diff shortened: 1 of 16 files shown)


e2e/tests-dfx/usage.bash
```diff
@@ -43,3 +43,8 @@ teardown() {
   dfx identity get-principal
   assert_directory_not_exists .dfx
 }
+
+@test "does not unconditionally read dfx.json" {
+  echo "garbage" >dfx.json
+  dfx identity get-principal
```
lwshang, line 49:
```suggestion
  assert_command dfx identity get-principal
```

lwshang (approved):
LGTM. Just a missing `assert_command` in the e2e test.

PR #3723, opened 2024-04-23
fix: for dfx new failure if node is installed but npm is not

ericswanson-dfinity:
# Description

`dfx new` would fail with "Failed to scaffold frontend code" if node was installed, but npm was not installed.

Also, now checks the status code when calling node and npm to see if they are installed. This means if there is a script that outputs something like the below, dfx will no longer assume the program is installed.

```
$ npm
Command 'npm' not found, but can be installed with: apt install npm
Please ask your administrator.
```

# How Has This Been Tested?

Added an e2e test

# Checklist:

### DIFF
--- a/e2e/tests-dfx/new.bash
+++ b/e2e/tests-dfx/new.bash
@@ -96,12 +96,8 @@ teardown() {
   which node || skip "node not installed"
   which npm || skip "npm not installed"
 
-  mkdir forward-to-node
-  cat > forward-to-node/node <<EOF
-#!/bin/bash
-$(dirname "$(which node)")/node "\$@"
-EOF
-  chmod +x forward-to-node/node
+  mkdir node-only
+  cp "$(which node)" node-only/node
 
   mkdir node-installable
   cat > node-installable/node <<EOF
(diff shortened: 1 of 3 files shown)


e2e/tests-dfx/new.bash
```diff
@@ -91,3 +91,68 @@ teardown() {
...
+  FORWARD_TO_NODE_DIR=$PWD/forward-to-node
+  NODE_INSTALLABLE_DIR=$PWD/node-installable
+  NPM_INSTALLABLE_DIR=$PWD/npm-installable
+
+  # neither node nor npm are installed (no binaries)
+  PATH="/usr/bin:/bin:$DFX_BIN_DIR" \
+    assert_command dfx new e2e_project1 --type motoko --frontend sveltekit
+  assert_contains "Node could not be found. Skipping installing the frontend example code."
+  assert_contains "npm could not be found. Skipping installing the frontend example code."
+  assert_contains "You can bypass this check by using the --frontend flag."
```
lwshang, line 136: The executed command has the `--frontend` flag. 
IIUC, this line "You can bypass this check by using the --frontend flag." cannot be true.
The instruction might be outdated after the big `dfx new` rewrite.

ericswanson-dfinity: Tracked in https://dfinity.atlassian.net/browse/SDKTG-322

PR #3727, opened 2024-04-24
feat: Update agent to dac18df and add `wasm_memory_limit` support

adamspofford-dfinity:
Includes the new `wasm_memory_limit` canister setting.


ericswanson-dfinity (changes requested):
e2e

src/dfx-core/src/config/model/dfinity.rs
```diff
@@ -429,6 +430,18 @@ pub struct InitializationValues {
...
+    /// Specifies a soft limit (in bytes) on the Wasm memory usage of the canister.
+    ///
+    /// Update calls, timers, heartbeats, installs, and post-upgrades fail if the
+    /// WASM memory usage exceeds this limit. The main purpose of this setting is
+    /// to protect against the case when the canister reaches the hard 4GiB
+    /// limit.
+    ///
+    /// Must be a number between 0 and 2^48 (i.e. 256 TiB), inclusive.
+    #[schemars(with = "Option<u64>")]
+    pub wasm_memory_limit: Option<Byte>,
```
lwshang, line 444: The type and schemar of this filed seems to be consistent with the existing `memory_allocation` field above.

But it is not consistent with the CLI flag `--wasm-memory-limit` below.

Can we have a unified description on the input format for this config field and the CLI flag?

lwshang (changes requested):
A few suggestions:
1. Rename the PR title. It is far beyond upgrade agent-rs dependency.
2. Add tests

src/dfx-core/src/util/mod.rs
```diff
@@ -9,3 +15,31 @@ pub fn expiry_duration() -> Duration {
...
+    }
+    fn schema_id() -> Cow<'static, str> {
+        Cow::Borrowed("byte_unit::Byte")
+    }
+    fn json_schema(_: &mut SchemaGenerator) -> Schema {
+        Schema::Object(SchemaObject {
+            instance_type: Some(vec![InstanceType::Integer, InstanceType::String].into()),
+            number: None,
+            string: Some(Box::new(StringValidation {
+                pattern: Some("^[0-9]+( *([KkMmGgTtPpEeZzYy]i?)?B)?$".to_string()),
```
lwshang, line 33: The e2e test shows that `8b` is allowed. But this regex denies it. The final `B` can be lowercase.

e2e/tests-dfx/update_settings.bash
```diff
@@ -42,6 +42,20 @@ teardown() {
   assert_match "Freezing threshold: 100_000_000_000"
 }
 
+@test "set wasm memory limit" {
```
lwshang, line 45: What about expanding this test to also cover following paths?
* `dfx canister create --wasm-memory-limit`
* Setting the field in `dfx.json`

lwshang (approved):

ericswanson-dfinity (approved):

PR #3728, opened 2024-04-26
ci: fix path on Windows

lwshang:
# Description

So that we won't get the CI warning like [this](https://github.com/dfinity/sdk/actions/runs/8852106002/job/24310068530#step:8:12).

Also did some other simplifications to the workflows.


.github/workflows/shellcheck.yml
```diff
@@ -23,19 +23,11 @@ env:
 
 jobs:
   check_macos:
```
adamspofford-dfinity, line 25: Maybe this should be renamed

adamspofford-dfinity (approved):

PR #3755, opened 2024-05-14
feat: subaccount can be derived from principal in dfx ledger account-id

mraszyk:
# Description

This PR makes it possible to derive the subaccount from a principal in the `dfx ledger account-id` command.
This is useful for NNS voters to validate NNS subnet rental request proposals by deriving the ledger account (controlled by the subnet rental canister's principal and using the proposer's principal as the subaccount) to which the proposer should have transfered the deposit.

# How Has This Been Tested?

A new e2e test has been added comparing the account ID returned by DFX against the account ID computed by the SRC for the caller.

### DIFF
--- a/src/dfx/src/commands/ledger/account_id.rs
+++ b/src/dfx/src/commands/ledger/account_id.rs
@@ -19,6 +19,10 @@ pub struct AccountIdOpts {
     #[arg(long, value_name = "SUBACCOUNT")]
     /// Subaccount identifier (64 character long hex string).
     pub subaccount: Option<Subaccount>,
+
+    #[arg(long, value_name = "SUBACCOUNT_FROM_PRINCIPAL")]
+    /// Principal from which the subaccount identifier is derived.
+    pub subaccount_from_principal: Option<Principal>,
 }
 
 pub async fn exec(env: &dyn Environment, opts: AccountIdOpts) -> DfxResult {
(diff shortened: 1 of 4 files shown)


src/dfx/src/commands/ledger/account_id.rs
```diff
@@ -36,6 +40,17 @@ pub async fn exec(env: &dyn Environment, opts: AccountIdOpts) -> DfxResult {
         env.get_selected_identity_principal()
             .context("No identity is selected")?
     };
-    println!("{}", AccountIdentifier::new(principal, opts.subaccount));
+    let subaccount = if let Some(subaccount) = opts.subaccount {
+        if opts.subaccount_from_principal.is_some() {
+            return Err(anyhow!(
+                "You can specify at most one of subaccount and subaccount-from-principal arguments."
+            ));
+        }
```
sesi200, line 48: Can be handled by clap, e.g. like this: https://github.com/dfinity/sdk/blob/cb67f24b7acb9aaa4609f38a34527a9436fb6b72/src/dfx/src/commands/deploy.rs#L78

mraszyk: Thank you for the suggestion! done

ericswanson-dfinity (changes requested):
Please update https://github.com/dfinity/sdk/blob/master/docs/cli-reference/dfx-ledger.mdx

ericswanson-dfinity (approved):

PR #3756, opened 2024-05-14
fix: Capitalization of "Wasm"

ulan:
This replaces incorrect capitalization of "WASM" with "Wasm", which is the official abbreviation of WebAssembly.

See: https://webassembly.org/

The reason why not all letters are capitalized is because "Wasm" is not an acronym, but rather an abbreviation.

This change also fixes capitalization of the canister status output to be consistent wih other fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### chore: Improve help text of `dfx identity new` to include which characters are valid in identity names
 
+### fix: Capitalization of "Wasm" in docs and messages
+
 ## Dependencies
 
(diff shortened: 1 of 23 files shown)


ericswanson-dfinity (approved):

ulan:
@ericswanson-dfinity: thanks for the fast review! To double check: is it safe to change the user facing messages this way? I was wondering it that could break some scripts.

ericswanson-dfinity:
> @ericswanson-dfinity: thanks for the fast review! To double check: is it safe to change the user facing messages this way? I was wondering it that could break some scripts.

It's always possible that changing the output of a program could break some scripts.  I suppose you could call out the change to `dfx canister status` output in the changelog.

PR #3761, opened 2024-05-15
feat: PocketIC support

adamspofford-dfinity:


src/dfx/src/actors/pocketic.rs
```diff
@@ -0,0 +1,323 @@
...
+        loop {
+            let _ = std::fs::remove_file(&config.port_file);
+            let last_start = std::time::Instant::now();
+            debug!(logger, "Starting PocketIC...");
+            let mut child = cmd.spawn().expect("Could not start PocketIC.");
+
+            let port = PocketIc::wait_for_port_file(&config.port_file).unwrap();
+
+            if let Err(e) = block_on_initialize_pocketic(port, logger.clone()) {
+                error!(logger, "Failed to initialize replica: {:#}", e);
```
ericswanson-dfinity, line 228:
```suggestion
                error!(logger, "Failed to initialize PocketIC: {:#}", e);
```

e2e/tests-dfx/start.bash
```diff
@@ -92,6 +93,7 @@ teardown() {
 }
 
 @test "dfx restarts the replica" {
+  [[ "$USE_POCKETIC" ]] && skip "skipped for pocketic"
```
ericswanson-dfinity, line 96: can we have an equivalent test for dfx restarting pocketic?

e2e/tests-dfx/start.bash
```diff
@@ -155,6 +158,7 @@ teardown() {
 }
 
 @test "dfx restarts icx-proxy when the replica restarts" {
+  [[ "$USE_POCKETIC" ]] && skip "skipped for pocketic"
```
ericswanson-dfinity, line 161: can we have an equivalent test for restarting icx-proxy when pocketic restarts?

src/dfx/src/commands/start.rs
```diff
@@ -321,7 +327,11 @@ pub fn exec(
...
     };
 
-    let effective_config = CachedConfig::replica(&replica_config);
+    let effective_config = if pocketic {
+        CachedConfig::pocketic()
+    } else {
+        CachedConfig::replica(&replica_config)
+    };
 
     if !clean && !force && previous_config_path.exists() {
```
ericswanson-dfinity, line 336: Since the PocketIC doesn't persist any data, this will need to reset the network id, so that any projects will reset their local canister id store.

To see the problem, try this procedure:
1. `dfx start --pocketic --background --clean`
2. `dfx deploy` (notice this succeeds)
3. `dfx stop`
4. `dfx start --pocketic --background`
5. `dfx deploy`

notice the second deploy fails:
```
Caused by: The replica returned a rejection error: reject code DestinationInvalid, reject message Canister txyno-ch777-77776-aaaaq-cai not found, error code Some("IC0301")
```
Expected behavior: `dfx deploy` creates new canisters and deploys successfully

src/dfx/src/actors/pocketic.rs
```diff
@@ -0,0 +1,323 @@
...
+                let _ = child.kill();
+                let _ = child.wait();
+                if receiver.try_recv().is_ok() {
+                    debug!(logger, "Got signal to stop.");
+                    break;
+                } else {
+                    continue;
+                }
+            }
+            addr.do_send(signals::PocketIcRestarted { port });
```
ericswanson-dfinity, line 238: When running without `--background`, `dfx start` indicates pretty clearly when it's ready:
```
$ dfx start --clean 
Running dfx start for version 0.20.1-beta.0+rev26.e12a2404
Using shared network 'local' defined in /Users/ericswanson/.config/dfx/networks.json
Initialized replica.
Dashboard: http://localhost:59729/_/dashboard
```

However, when running with PocketIC, it can give the idea that it got stuck:
```
$ dfx start --clean --pocketic
Running dfx start for version 0.20.1-beta.0+rev26.e12a2404
Using shared network 'local' defined in /Users/ericswanson/.config/dfx/networks.json
```

Can we display some indication that it's ready to go?

src/dfx-core/src/config/model/local_server_descriptor.rs
```diff
@@ -146,6 +146,16 @@ impl LocalServerDescriptor {
...
     }
 
+    /// This file contains the listening port of the pocket-ic process
+    pub fn pocketic_port_path(&self) -> PathBuf {
+        self.data_directory.join("pocket-ic.port")
+    }
+
+    /// Returns whether the local server is PocketIC (as opposed to the replica)
+    pub fn is_pocketic(&self) -> bool {
+        self.pocketic_port_path().exists()
```
ericswanson-dfinity, line 156: This probably won't work well with the call to `empty_writable_path` in start.rs, which unconditionally creates an empty file.  

https://github.com/dfinity/sdk/pull/3761/files#diff-04d9df0d52b6d87bc92f994bdc87cfe37982f42e3db092347d0c5b688380f8adR246

src/dfx/src/lib/environment.rs
```diff
@@ -284,12 +295,20 @@ impl<'a> AgentEnvironment<'a> {
                 and use it in mainnet-facing commands with the `--identity` flag", identity.name());
         }
         let url = network_descriptor.first_provider()?;
+        let effective_canister_id = network_descriptor
+            .local_server_descriptor
+            .as_ref()
+            .and_then(|d| {
+                d.is_pocketic()
+                    .then(|| Principal::from_text("tqzl2-p7777-77776-aaaaa-cai").unwrap())
```
ericswanson-dfinity, line 303: Please follow the method used by MAINNET_LEDGER_CANISTER_ID, MAINNET_CYCLE_MINTER_CANISTER_ID, and others, for defining const Principals

ericswanson-dfinity (changes requested):

ericswanson-dfinity (changes requested):
Please update https://github.com/dfinity/sdk/blob/master/docs/cli-reference/dfx-start.mdx

e2e/tests-dfx/start.bash
adamspofford-dfinity: what do you think of a flag to write the PID when --background is used?

src/dfx/src/lib/environment.rs
```diff
@@ -358,7 +375,16 @@ impl<'a> Environment for AgentEnvironment<'a> {
     }
 
     fn get_effective_canister_id(&self) -> Principal {
-        self.backend.get_effective_canister_id()
+        self.backend
+            .get_override_effective_canister_id()
+            .unwrap_or_else(|| {
+                self.effective_canister_id
```
mraszyk, line 381: I wonder why this is attempted in the middle.

scripts/workflows/e2e-matrix.py
```diff
@@ -15,8 +15,22 @@ def test_scripts(prefix):
...
+            "test": "dfx/bitcoin"
+        },
+        {
+            "backend": "pocketic",
+            "test": "dfx/canister_http"
+        },
+        {
+            "backend": "pocketic",
+            "test": "dfx/canister_logs"
+        }
```
mraszyk, line 32: Note for a follow up PR: this should be reenabled once PocketIC is taken from the latest master commit (enabling canister logging).

e2e/tests-dfx/start.bash
```diff
@@ -449,5 +456,6 @@ teardown() {
 }
 
 @test "dfx start then ctrl-c won't hang and panic but stop actors quickly" {
+  [[ "$USE_POCKETIC" ]] && skip "skipped for pocketic"
```
mraszyk, line 459: why is this test disabled for PocketIC?

e2e/tests-dfx/ping.bash
```diff
@@ -64,6 +64,7 @@ teardown() {
 }
 
 @test "dfx ping succeeds by arbitrary network name to a nonstandard port" {
+  [[ "$USE_POCKETIC" ]] && skip "skipped for pocketic"
```
mraszyk, line 67: is dfx ping not supported for PocketIC?

mraszyk:
Since PocketIC doesn't provide the replica dashboard, we should make sure that the replica dashboard link is not printed if PocketIC is used.

src/dfx/src/lib/environment.rs
adamspofford-dfinity: It's more important than the underlying Environment's previous setting but not as important as the user override.

src/dfx/src/actors/pocketic.rs
```diff
@@ -0,0 +1,328 @@
...
+        .send()
+        .await?
+        .error_for_status()?;
+    info!(logger, "Initialized PocketIC.");
+    Ok(())
+}
+
+#[cfg(not(unix))]
+async fn initialize_pocketic(_: u16, _: Logger) -> DfxResult {
+    bail!("PocketIC client library not supported on this platform")
```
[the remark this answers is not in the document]
adamspofford-dfinity: That is why it's there, though. The library doesn't compile.

e2e/tests-dfx/start.bash
ericswanson-dfinity: I'd expect dfx to write a pid file unconditionally, like it does for the replica

adamspofford-dfinity:
@mraszyk The dashboard log is part of the replica module, not part of the start command.

e2e/utils/_.bash
```diff
@@ -130,30 +133,38 @@ dfx_start() {
             fi
         done
         if ! $has_param; then
-            args+=( "$param_name" "$param_value" )
+            args+=( "$param_name" ${param_value+"$param_value"} )
```
ericswanson-dfinity, line 136: I'm having trouble following what's going on here.  I'm sure this works, but could we do this in a more conventional way by deleting the `$#` check above and doing this?
```
if [[ $# == 1 ]]; then
    args += ( "$param_name" )
else
    local param_value=$2
    args += ( "$param_name" "$param_value" )
fi
```

e2e/utils/_.bash
```diff
@@ -268,6 +283,14 @@ get_replica_port() {
   cat "$E2E_NETWORK_DATA_DIRECTORY/replica-configuration/replica-1.port"
 }
 
+get_pocketic_pid() {
+  cat "$E2E_NETWORK_DATA_DIRECTORY/pocket-ic-pid"
+}
+
+get_pocketic_port() {
+  cat "$E2E_NETWORK_DATA_DIRECTORY/pocket-ic.port"
```
ericswanson-dfinity, line 291: This can be `pocket-ic-port`, which will be consistent with `pocket-ic-pid` as well as `webserver-port`.  The `replica-1.port` filename is the only one that's inconsistent with everything else, and that is a historical artifact.

ericswanson-dfinity (approved):

PR #3770, opened 2024-05-24
fix: fix url decoding in ic-certified-assets

raymondk:
# Description

The url_decode function in ic-certified-assets didn't correctly decode some special charaters.
Instead of fixing our implementation of the decode function, this code is now using https://crates.io/crates/percent-encoding-rfc3986.
See this PR for more context: https://github.com/dfinity/sdk/pull/3767

It was suggested to use https://crates.io/crates/percent-encoding because it is more widely used and part of servo. I found that it can have unexpected behavior when decoding an invalid string (see: https://github.com/servo/rust-url/blob/main/percent_encoding/src/lib.rs#L381). I preferred percent-encoding-rfc3986 because it handles bad encodings better by returning an error if the decoding fails. Since we're only using the library for decoding how it encodes is not really relevant and other than the error handling, the implementation of the error handling is the same.

Fixes # (SDK-1412)

# How Has This Been Tested?

Added some unit tests and fixed the broken ones.

# Checklist:


raymondk:
Going to close this PR and use https://crates.io/crates/percent-encoding instead.

PR #3836, opened 2024-07-17
feat!: Change crate selection logic to use an explicit crate field

adamspofford-dfinity:
Currently, dfx asks you to specify a package, and selects what it thinks is the default crate from that package. This misfires when a package contains multiple possible crates, as @sesi200 reported with the cycles ledger. This PR changes it to only assume that the default crate is the one with the same name as the package, and use a new `crate` field in dfx.json to explicitly specify non-default crates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 # UNRELEASED
 
+### feat!: add `crate` field to dfx.json
+
+It is now possible to specify a particular crate within a Rust package to use for a canister module, using the `crate` field.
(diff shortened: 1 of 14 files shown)


lwshang (approved):
Is this really a breaking change?

adamspofford-dfinity:
Repeat of slack discussion for posterity:
- The breaking change is if you were using a renamed library or binary crate, it won't be auto-detected without the crate field anymore
- This is necessary to avoid misconfigurations causing dfx to use the wrong crate silently.

PR #3841, opened 2024-07-25
feat: accept --wallet <identity name> in addition to --wallet <principal>

ericswanson-dfinity:
# Description

The `--wallet` parameter can now be either the principal of a wallet or the name of an identity.

If the name of an identity, dfx looks up the associated wallet's principal.

This means `--wallet <name>` is the equivalent of `--wallet $(dfx identity get-wallet --identity <name>)`.

Fixes https://dfinity.atlassian.net/browse/SDK-1483

# How Has This Been Tested?

Added e2e tests and updated existing e2e tests to use the new (shorter) form.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,14 @@ It is now possible to specify a particular crate within a Rust package to use fo
 This enables specifying crates with different names than the package. In a few cases these were previously auto-detected
 by dfx, you will need to add this field if you were using such a setup.
 
+### feat: the `--wallet` parameter now accepts an identity name
+
+The `--wallet` parameter can now be either a principal or the name of an identity.
+
+If the name of an identity, dfx looks up the associated wallet's principal.
+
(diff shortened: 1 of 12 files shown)


lwshang (approved):
LGTM.

---

I have an idea that might further improve this feature: What about also allowing users pass `--wallet` as a flag without specifying a Principal or an identity? Then the wallet for current active identity (`dfx identity whoami`) will be used. 

Of course, this doesn't need to be included in this PR. It might be something we can consider for a future iteration.

ericswanson-dfinity:
> LGTM.
> 
> I have an idea that might further improve this feature: What about also allowing users pass `--wallet` as a flag without specifying a Principal or an identity? Then the wallet for current active identity (`dfx identity whoami`) will be used.
> 
> Of course, this doesn't need to be included in this PR. It might be something we can consider for a future iteration.

Sounds reasonable to consider for a future iteration, if clap supports it

PR #3843, opened 2024-07-26
feat: replace icx-proxy with PocketIC HTTP gateway 

adamspofford-dfinity:


src/dfx/src/actors/pocketic_proxy.rs
```diff
@@ -0,0 +1,361 @@
...
+            let last_start = std::time::Instant::now();
+            debug!(logger, "Starting pocket-ic gateway...");
+            let mut child = cmd.spawn().expect("Could not start pocket-ic gateway.");
+
+            std::fs::write(&pocketic_proxy_pid_path, "")
+                .expect("Could not write to pocketic-proxy-pid file.");
+            std::fs::write(&pocketic_proxy_pid_path, child.id().to_string())
+                .expect("Could not write to pocketic-proxy-pid file.");
+            let port =
+                PocketIcProxy::wait_for_ready(&pocketic_proxy_port_path, &ready_file).unwrap();
```
mraszyk, line 249: Could we also check for the kill signals here so that we aren't hanging until the maximum number of retries in case the PocketIC server has been CTRL-C'ed?

nix/sources.json
```diff
@@ -169,17 +169,17 @@
         "version": "0.11.2"
     },
     "pocket-ic-x86_64-darwin": {
-        "rev": "5849c6daf2037349bd36dcb6e26ce61c2c6570d0",
-        "sha256": "06c547zhdfcf910zfz50g3aniw2b993m7c0rhsn9v1x163xfnzrm",
+        "rev": "6aceb6a35248ef2735ddd9ca99d8a1c6f4a13908",
```
mraszyk, line 172: This is too old for the PocketIC library v4.0.0. Could you please bump to the PocketIC server v5.0.0 (IC commit cec100d1607eb2258441ff292a144ba88564458e) which is [compatible](https://docs.google.com/document/d/1VYmHUTjrgbzRHtsAyRrI5cj-gWGs7ktTnutPvUMJioU/edit#heading=h.5wf28dvt742x) with the PocketIC library v4.0.0?

ericswanson-dfinity (changes requested):
Please add an entry to the changelog.  It's probably enough to say that the pocket-ic process replaces the icx-proxy process.

ericswanson-dfinity (approved):

PR #3861, opened 2024-08-05
feat: Canister snapshots

adamspofford-dfinity:
Implements the canister backup/restore feature in SDK-1403.


src/dfx/src/commands/canister/snapshot.rs
```diff
@@ -0,0 +1,187 @@
...
+                replace.as_ref().map(|x| &*x.0),
+                call_sender,
+            )
+            .await
+            .with_context(|| format!("Failed to take snapshot of canister {canister}"))?;
+            info!(
+                logger,
+                "Created a new snapshot of canister {canister}. Snapshot ID: {}",
+                SnapshotId(id.id)
+            );
```
ericswanson-dfinity, line 122: What would you think about breaking up this large method into smaller ones?

src/dfx/src/commands/canister/snapshot.rs
```diff
@@ -0,0 +1,187 @@
...
+    let canisters = env.get_canister_id_store()?;
+    let logger = env.get_logger();
+    match opts.subcmd {
+        SnapshotSubcommand::Create {
+            canister,
+            replace,
+            force,
+        } => {
+            let canister_id = canister.parse().or_else(|_| canisters.get(&canister))?;
+            if !force {
```
ericswanson-dfinity, line 100: Where is this "force" behavior coming from a design/spec/product perspective? There is a race condition here as long as dfx enforces the behavior.

src/dfx/src/commands/canister/snapshot.rs
```diff
@@ -0,0 +1,187 @@
...
+                let status = get_canister_status(env, canister_id, call_sender)
+                    .await
+                    .with_context(|| format!("Could not retrieve status of canister {canister}"))?;
+                match status.status {
+                    CanisterStatus::Stopped => {}
+                    CanisterStatus::Running => bail!("Canister {canister} is running and snapshots should not be applied to running canisters. Run `dfx canister stop` first (or override with `--force`)"),
+                    CanisterStatus::Stopping => bail!("Canister {canister} is stopping but is not yet stopped. Wait a few seconds and try again"),
+                }
```
ericswanson-dfinity, line 138: Please DRY with the other force check

ericswanson-dfinity (changes requested):

src/dfx/src/lib/operations/canister/motoko_playground.rs
```diff
@@ -65,12 +65,17 @@ impl TryFrom<InstallMode> for PlaygroundInstallMode {
         match m {
             InstallMode::Install => Ok(Self::Install),
             InstallMode::Reinstall => Ok(Self::Reinstall),
-            InstallMode::Upgrade {
-                skip_pre_upgrade: Some(false) | None,
-            } => Ok(Self::Upgrade),
-            InstallMode::Upgrade {
+            InstallMode::Upgrade(
+                Some(CanisterUpgradeOptions {
+                    skip_pre_upgrade: None | Some(false),
```
ericswanson-dfinity, line 70: Starting to wonder why skip_pre_upgrade isn't `bool` instead of `Option<bool>`

src/dfx/src/commands/canister/snapshot.rs
adamspofford-dfinity: My assumption was that someone would *eventually* find a use for it. Is it niche enough that we should not support it? And I don't think there's any way to remove the race condition unfortunately with the current protocol.

ericswanson-dfinity: I don't think we should support a `force` flag. At least an enforced order of operations `dfx canister stop && dfx canister snapshot create/apply && dfx canister start` gives a hint that it might not be good if another `dfx canister start` happens too soon.

Unfortunately, in either case, it would be reasonable to assume that the replica performs this kind of integrity check, since it's the only way to actually enforce the constraint. I would also recommend sending this back to the feature designers to find out if there is a good reason why it isn't so.

src/dfx/src/lib/operations/canister/motoko_playground.rs
adamspofford-dfinity: No breaking changes 🥲

ericswanson-dfinity (approved):

PR #3872, opened 2024-08-14
feat: embedding well known canisters at build time

NikolaMilosa:
# Description

DRE team would highly benefit if we could have a feature like this for well known canisters. Probably others as well. This PR adds a `build.rs`  for dfx-core which creates a file at build time for well known canisters.

Fixes # (issue)

# How Has This Been Tested?

TODO: will adapt if sdk team agrees to push for this. Also will add tests

# Checklist:

### DIFF
--- a/src/dfx-core/src/build.rs
+++ b/src/dfx-core/src/build.rs
@@ -1,0 +1,35 @@
+use std::{env, fs::File, io::Write, path::Path};
+
+const CANISTER_IDS_URL: &str = "https://raw.githubusercontent.com/dfinity/ic/1402bf35308ec9bd87356c26f7c430f49b49423a/rs/nns/canister_ids.json";
+fn define_well_known_canisters() {
+    let well_known_canisters = reqwest::blocking::get(CANISTER_IDS_URL)
+        .unwrap()
+        .error_for_status()
+        .unwrap()
+        .text()
+        .unwrap();
+
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let loader_path = Path::new(&out_dir).join("well_known_canisters.rs");
+    let mut f = File::create(loader_path).unwrap();
+    f.write_all(
+        format!(
+            "
+const WELL_KNOWN_CANISTERS: &str = r#\"
+{}
+\"#;
+
+pub fn map_wellknown_canisters() -> CanisterIds {{
+    serde_json::from_str(WELL_KNOWN_CANISTERS).unwrap_or(CanisterIds::new())
+}}
+",
+            well_known_canisters.replace("mainnet", "ic")
+        )
+        .as_bytes(),
+    )
+    .unwrap()
+}
+
+fn main() {
+    define_well_known_canisters();
+}
--- a/src/dfx-core/src/config/model/canister_id_store.rs
+++ b/src/dfx-core/src/config/model/canister_id_store.rs
@@ -17,6 +17,8 @@ use std::time::{Duration, SystemTime};
 use time::format_description::well_known::Rfc3339;
 use time::OffsetDateTime;
 
+include!(concat!(env!("OUT_DIR"), "/well_known_canisters.rs"));
+
 pub type CanisterName = String;
 pub type NetworkName = String;
 pub type CanisterIdString = String;
@@ -152,7 +154,7 @@ impl CanisterIdStore {
         };
         let ids = match &canister_ids_path {
             Some(path) if path.is_file() => crate::json::load_json_file(path)?,
-            _ => CanisterIds::new(),
+            _ => map_wellknown_canisters(),
         };
         let acquisition_timestamps = match &canister_timestamps_path {
             Some(path) if path.is_file() => crate::json::load_json_file(path)?,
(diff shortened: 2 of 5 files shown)


src/dfx-core/src/build.rs
```diff
@@ -0,0 +1,35 @@
+use std::{env, fs::File, io::Write, path::Path};
+
+const CANISTER_IDS_URL: &str = "https://raw.githubusercontent.com/dfinity/ic/1402bf35308ec9bd87356c26f7c430f49b49423a/rs/nns/canister_ids.json";
```
ericswanson-dfinity, line 3: It looks like we only add a well-known canister once every year or two.  Let's instead copy the file into our repo rather than downloading it on each build.

src/dfx/src/commands/canister/sign.rs
```diff
@@ -75,6 +76,9 @@ pub async fn exec(
     opts: CanisterSignOpts,
     call_sender: &CallSender,
 ) -> DfxResult {
+    let agent = env.get_agent();
+    fetch_root_key_if_needed(env).await?;
```
ericswanson-dfinity, line 80: Let's instantiate do this if/when the fetch happens.

src/dfx-core/src/build.rs
```diff
@@ -0,0 +1,35 @@
...
+        .unwrap()
+        .error_for_status()
+        .unwrap()
+        .text()
+        .unwrap();
+
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let loader_path = Path::new(&out_dir).join("well_known_canisters.rs");
+    let mut f = File::create(loader_path).unwrap();
+    f.write_all(
```
ericswanson-dfinity, line 15: The format of canister_ids.json is canister name -> network -> canister id.  For these well-known canisters, the canister id will be the same on every network.

So instead of including the json contents as a string and parsing json in the emitted code, I'd suggest instead doing that here and emitting code for a lazily instantiated HashMap<CanisterName, Principal>

src/dfx-core/src/build.rs
ericswanson-dfinity: Addition of a "well-known" canister is infrequent enough that I think it would be ok to leave it to manual update. If you feel inspired to update it in `update-replica-version.yaml` that's fine too.

NikolaMilosa: Alright, I can agree with that. Lets leave it manual for now then. If automation is needed you can always reach out and I would be happy to add it.

src/dfx-core/assets/build.rs
```diff
@@ -0,0 +1,54 @@
...
+    )
+    .unwrap();
+    let well_known_canisters = well_known_canisters.as_object().unwrap();
+    let well_known_canisters = well_known_canisters.iter().map(|(key, val)| {
+        (
+            key.as_str(),
+            val.as_object()
+                .unwrap()
+                .values()
+                .last()
```
ericswanson-dfinity, line 22: The principals are all the same for all networks in the included canister_ids.json, but what would you think about looking up entry for `mainnet` here instead of whatever's last?

src/dfx-core/assets/build.rs
```diff
@@ -0,0 +1,54 @@
...
+    WELL_KNOWN_CANISTERS.iter().map(|(key, value)| (key.to_string(), (*value).try_into().unwrap())).collect()
+}}
+",
+            well_known_canisters
+                .map(|(key, val)| format!("(\"{}\", \"{}\")", key, val))
+                .join(",\n")
+        )
+        .as_bytes(),
+    )
+    .unwrap()
```
ericswanson-dfinity, line 49: Since this is a global mapping of well-known canister ids, could we initialize this into a `LazyLock` rather than adding it as a member to CanisterIdStore?

src/dfx-core/src/config/model/canister_id_store.rs
```diff
@@ -189,6 +195,16 @@ impl CanisterIdStore {
...
+            .or_else(|| {
+                let principal = match Principal::from_str(canister_id) {
+                    Ok(p) => p,
+                    Err(_) => return None,
+                };
+                self.well_known_ids
+                    .iter()
+                    .find(|(_, id)| &&principal == id)
+                    .map(|(name, _)| name)
+            })
```
ericswanson-dfinity, line 207: If we do this here, before looking up the name in the project, then this is a potentially breaking change.  (maybe it is anyway).

ericswanson-dfinity (changes requested):
Please add a note to the changelog.  This PR should probably also include migration notes, for projects that define these as remote canisters.  Migration notes go here: https://github.com/dfinity/sdk/tree/master/docs/migration

Also, I imagine that ideally this kind of change would make it unnecessary to manually define well-known canisters as remote canisters in dfx.json.  Does this PR go that far?

src/dfx-core/assets/build.rs
NikolaMilosa: The version of rust used by `sdk` doesn't have `LazyLock` stabilized.

src/dfx-core/src/config/model/canister_id_store.rs
NikolaMilosa: Agree fully, will move it to the end so it just adds on top and does not break anything previous.

NikolaMilosa:
Not sure how I would go about migrations docs for this. Do you have some proposals?

raymondk:

PR #3873, opened 2024-08-16
feat: add fn load_settings_digest to load_settings_digest

mraszyk:
This PR adds a new function `load_settings_digest` to fetch the settings digest from the `network-id` file in the data directory if no settings digest is set in `LocalServerDescriptor`.

The fix has been tested manually against a development version of the NNS extension.

### DIFF
--- a/src/dfx-core/src/config/model/local_server_descriptor.rs
+++ b/src/dfx-core/src/config/model/local_server_descriptor.rs
@@ -10,7 +10,6 @@ use crate::error::network_config::{
     NetworkConfigError, NetworkConfigError::ParseBindAddressFailed,
 };
 use crate::error::structured_file::StructuredFileError;
-use crate::error::structured_file::StructuredFileError::DeserializeJsonFileFailed;
 use crate::json::load_json_file;
 use crate::json::structure::SerdeVec;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


src/dfx-core/src/config/model/local_server_descriptor.rs
```diff
@@ -173,10 +181,14 @@ impl LocalServerDescriptor {
...
-        self.settings_digest
-            .as_ref()
-            .expect("settings_digest must be set")
+    pub fn settings_digest(&self) -> String {
+        self.settings_digest.clone().unwrap_or_else(|| {
+            let network_id_contents = crate::fs::read(&self.data_directory.join("network-id"))
+                .expect("Could not read network-id file.");
+            let network_metadata: NetworkMetadata = serde_json::from_slice(&network_id_contents)
+                .expect("Could not parse network-id contents.");
+            network_metadata.settings_digest
```
ericswanson-dfinity, line 190: Please move this to a separate load method that returns a `Result`. This method panics if the settings digest isn't set because it should only be called if the settings digest was set earlier.

ericswanson-dfinity (changes requested):

src/dfx-core/src/config/model/local_server_descriptor.rs
mraszyk: OK. Done.

src/dfx-core/src/config/model/local_server_descriptor.rs
```diff
@@ -173,6 +182,17 @@ impl LocalServerDescriptor {
         path.exists().then(|| load_json_file(&path)).transpose()
     }
 
+    pub fn load_settings_digest(&mut self) -> Result<(), StructuredFileError> {
+        if self.settings_digest.is_none() {
+            let network_id_path = self.data_directory.join("network-id");
+            let network_id_contents = crate::fs::read(&network_id_path)?;
+            let network_metadata: NetworkMetadata = serde_json::from_slice(&network_id_contents)
+                .map_err(|e| DeserializeJsonFileFailed(Box::new(network_id_path), e))?;
```
ericswanson-dfinity, line 190:
```suggestion
            let network_metadata: NetworkMetadata = load_json_file(&network_id_path?;
```

ericswanson-dfinity (approved):

PR #3888, opened 2024-08-22
chore: Update agent for transport changes

adamspofford-dfinity:


src/dfx/src/commands/canister/send.rs
```diff
@@ -30,38 +31,64 @@ pub async fn exec(
...
-    let message: SignedMessageV1 =
-        serde_json::from_str(&json).map_err(|_| anyhow!("Invalid json message."))?;
+    let message: SignedMessageV1 = load_json_file(&file_name)?;
     message.validate()?;
 
     let network = message.network.clone();
-    let transport =
-        ReqwestTransport::create(network).context("Failed to create transport object.")?;
+    let agent = Agent::builder().with_url(&network).build()?;
+    if !message.is_ic {
```
krpeacock, line 39: The Utopia team recently asked me to make sure that the root key could be set explicitly in the JS agent for when the agent is going to call a non-IC replica with a known root key.

It doesn't necessarily mean this check is wrong for now, but something to consider for this logic in the future

krpeacock (approved):

ericswanson-dfinity (approved):

PR #3895, opened 2024-08-28
feat: enable more pocketic features

adamspofford-dfinity:
Enables artificial delay, subnet types, and canister HTTP. Since canister_http.bash is almost entirely checking the performance of the adapter process, this splits out the actual http request test and leaves the original file replica-only.


src/dfx/src/actors/pocketic.rs
```diff
@@ -323,28 +319,43 @@ fn pocketic_start_thread(
...
-                fiduciary: None,
-                bitcoin: None,
-                system: vec![],
-                application: vec![SubnetSpec::default()],
-            },
-            state_dir: Some(state_dir),
+            subnet_config_set,
+            state_dir: Some(replica_config.state_manager.state_root.clone()),
             nonmainnet_features: true,
+            log_level: None,
```
mraszyk, line 358: Can't we set this, too, from the dfx replica config?

src/dfx/src/commands/start.rs
```diff
@@ -61,12 +61,12 @@ pub struct StartOpts {
     #[arg(long, conflicts_with = "pocketic")]
     enable_bitcoin: bool,
 
-    /// enable canister http requests
-    #[arg(long, conflicts_with = "pocketic")]
+    /// enable canister http requests (on by default for --pocketic)
```
mraszyk, line 64: Is it not on by default for the replica, too?

e2e/tests-dfx/start.bash
```diff
@@ -43,7 +42,7 @@ teardown() {
 }
 
 @test "stop and start with other options does not disrupt projects" {
-  [[ "$USE_POCKETIC" ]] && skip "skipped for pocketic: artificial delay"
+  [[ "$USE_POCKETIC" ]] && skip "skipped for pocketic: clean required"
```
mraszyk, line 45: Why is clean an issue?

adamspofford-dfinity: This is a test for #3777, which doesn't (yet) apply to PocketIC. This can be addressed in a future PR, I'm just updating the comments to indicate what's the remaining required feature.

src/dfx/src/commands/start.rs
adamspofford-dfinity: No. The reason it is a flag is because the outcalls adapter is a separate process which dfx doesn't start if it doesn't need to.

src/dfx/src/actors/pocketic.rs
adamspofford-dfinity: It is already set above in RUST_LOG. Should that be moved here?

mraszyk (approved):

ericswanson-dfinity (approved):

PR #3943, opened 2024-10-07
fix: Add `idle_cycles_burned_per_day` field

berestovskyy:
The field was missed in the canister status output.

### DIFF
--- a/src/dfx/src/commands/canister/status.rs
+++ b/src/dfx/src/commands/canister/status.rs
@@ -65,12 +65,13 @@ async fn canister_status(
         }
     };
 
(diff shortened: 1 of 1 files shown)


ByronBecker:
@berestovskyy just to clarify, I believe that `idle_cycles_burned_per_day` value already includes the `compute_allocation` additions in it?

So if I update my canister to have a 1% compute allocation (from zero), the `idle_cycles_burned_per_day` seems to increase by 864 billion (per percentage point increase).

Is this correct?

berestovskyy:
Yes, this field is used internally and returned in `canister_status`, but it's not currently visible in `dfx` for some reason. By increasing the compute allocation by 1%, we should see the `idle_cycles_burned_per_day` field increase by 10M cycles multiplied by 86,4K seconds (one day), which equals 864 billion cycles.

lwshang (approved):

PR #3967, opened 2024-10-29
chore: improve dfx deploy message.

vincent-dfinity:
# Description

Improve `dfx deploy` error message to let users know what to do next.

```
$ dfx deploy
Error: Failed to fetch the root key, did you run 'dfx start' to start the local replica?
An error happened during communication with the replica: error sending request for url (http://127.0.0.1:4943/api/v2/status)
```

Fixes # (issue)

# How Has This Been Tested?

Added an e2e test.

# Checklist:

### DIFF
--- a/src/dfx/src/lib/root_key.rs
+++ b/src/dfx/src/lib/root_key.rs
@@ -1,11 +1,12 @@
 use crate::{lib::error::DfxResult, Environment};
-
+use anyhow::anyhow;
 use dfx_core::network::root_key;
 
 pub async fn fetch_root_key_if_needed(env: &dyn Environment) -> DfxResult {
     let agent = env.get_agent();
     let network = env.get_network_descriptor();
-    root_key::fetch_root_key_when_non_mainnet(agent, network).await?;
+    root_key::fetch_root_key_when_non_mainnet(agent, network).await
+        .map_err(|e| anyhow!("Failed to fetch the root key, did you run 'dfx start' to start the local replica?\n{}", e))?;
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


marydwyer:
Is information about the rootkey at all relevent to the user? I'm wondering if we could streamline the message even further to something like this:

`Error: Failed to deploy because your a replica is not running.`
`Please run 'dfx start' and try again.`

vincent-dfinity:
> Is information about the rootkey at all relevent to the user? I'm wondering if we could streamline the message even further to something like this:
> 
> `Error: Failed to deploy because your a replica is not running.` `Please run 'dfx start' and try again.`

1. It just indicates where the error comes from.
2. The reason why I use a question `did you run 'dfx start' to start the local replica?` is I'm not quite sure if it's the only reason to fail fetching the root key. @sesi200  do you know if there're any other cases?

src/dfx/src/lib/root_key.rs
```diff
@@ -1,11 +1,12 @@
...
-
+use anyhow::anyhow;
 use dfx_core::network::root_key;
 
 pub async fn fetch_root_key_if_needed(env: &dyn Environment) -> DfxResult {
     let agent = env.get_agent();
     let network = env.get_network_descriptor();
-    root_key::fetch_root_key_when_non_mainnet(agent, network).await?;
+    root_key::fetch_root_key_when_non_mainnet(agent, network).await
+        .map_err(|e| anyhow!("Failed to fetch the root key, did you run 'dfx start' to start the local replica?\n{}", e))?;
```
ericswanson-dfinity, line 9: Please take a look at DiagnosedError. It may provide even more readable output, by first displaying the error, then displaying the explanation, and finally displaying a suggested action. You'd add it here with something like `.context(DiagnosedError::new(...))`.

ericswanson-dfinity (changes requested):

PR #3970, opened 2024-10-31
chore: update command comment

sesi200:
`dfx identity --help` only shows PEM file imports for `dfx identity import` but is actually also capable of handling seed phrase files as well


manudrijvers (approved):
thanks!

adamspofford-dfinity (approved):

PR #3974, opened 2024-11-06
chore: update agent deps in frontend projects

dfx-json:
# Description

This PR updates the agent version used in frontend templates generated by `dfx new` and moves the `@dfinity/<package>` dependencies out into the top-level package.json. 

Fixes # (issue)
Some users have reported issues running builds successfully with 2.x versions of the agent, so this resolves those issues.

# How Has This Been Tested?

Manual tests

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [x] I have edited the CHANGELOG accordingly.
- [x] I have made corresponding changes to the documentation.

PR #3975, opened 2024-11-06
chore: update agent deps in frontend projects (alternate approach)

rvanasa:
Alternative solution to #3974.


dfx-json (dismissed):

chenyan-dfinity:
Do we want to remove the package-lock.json file?

rvanasa:
Good point; updated.

chenyan-dfinity (approved):

PR #3998, opened 2024-11-15
fix: only bind to webserver port if the user did not specify one explicitly

mraszyk:
This PR aims to fix spurious errors of the form
```
Error: Failed to get frontend address.
Caused by: Failed to find available socket address
Caused by: Failed to bind socket to 127.0.0.1:8080.
Caused by: Address already in use (os error 48)
```
by only binding to a webserver port if the user did not specify one explicitly (port is equal to zero) and thus a free port is grabbed by the system.


mraszyk:
The approach from this PR isn't robust against users actually trying to dfx start with the same webserver port.

PR #4013, opened 2024-11-25
feat: impersonating sender of requests to a local PocketIC instance

mraszyk:
This PR enables impersonating sender of requests to a local PocketIC instance: `dfx canister call`, `dfx canister status`, and `dfx canister update-settings` take an additional CLI argument `--impersonate` to specify a principal on behalf of which requests to a local PocketIC instance are sent.


src/dfx/src/commands/canister/request_status.rs
```diff
@@ -46,46 +48,61 @@ pub async fn exec(env: &dyn Environment, opts: RequestStatusOpts) -> DfxResult {
...
-                        request_accepted = true;
+    let blob = if let Some(pocketic) = env.get_pocketic() {
+        let msg_id = RawMessageId {
+            effective_principal: RawEffectivePrincipal::CanisterId(canister_id.as_slice().to_vec()),
+            message_id: request_id.as_slice().to_vec(),
+        };
+        let res = pocketic
+            .await_call_no_ticks(msg_id)
+            .await
+            .map_err(|err| anyhow!("Canister call failed: {}", err))?;
```
ericswanson-dfinity, line 59:
```suggestion
            .map_err(|err| anyhow!("Canister request status failed: {}", err))?;
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 # UNRELEASED
 
+### feat: impersonating sender of requests to a local PocketIC instance
+
+`dfx canister call`, `dfx canister status`, and `dfx canister update-settings` take
```
ericswanson-dfinity, line 7: Please update the docs in https://github.com/dfinity/sdk/blob/master/docs/cli-reference/dfx-canister.mdx

e2e/tests-dfx/call.bash
```diff
@@ -298,3 +298,82 @@ teardown() {
   assert_command dfx canister call inter2_mo read
   assert_match '(8 : nat)'
 }
+
+@test "impersonate sender" {
+    [[ ! "$USE_POCKETIC" ]] && skip "skipped for replica: impersonating sender is only supported for PocketIC"
```
ericswanson-dfinity, line 303: Can we have this test impersonate an identity other than (or in addition to) aaaaa-aa? It looks like these tests would pass if `--impersonate` always used the anonymous identity rather than the impersonation mechanism.

src/dfx/src/commands/canister/request_status.rs
```diff
@@ -46,46 +48,61 @@ pub async fn exec(env: &dyn Environment, opts: RequestStatusOpts) -> DfxResult {
...
-                    // The system will return Unknown until the request is accepted
-                    // and we generally cannot know how long that will take.
-                    // State transitions between Received and Processing may be
-                    // instantaneous. Therefore, once we know the request is accepted,
-                    // we restart the waiter so the request does not time out.
-                    if !request_accepted {
-                        retry_policy.reset();
-                        request_accepted = true;
+    let blob = if let Some(pocketic) = env.get_pocketic() {
+        let msg_id = RawMessageId {
```
ericswanson-dfinity, line 52: is this change related to impersonation?

src/dfx/src/lib/environment.rs
```diff
@@ -422,3 +454,9 @@ pub fn create_agent(
         .build()?;
     Ok(agent)
 }
+
+#[context("Failed to create PocketIC handle with url {}.", url)]
+pub fn create_pocketic(url: &Url) -> DfxResult<PocketIc> {
```
ericswanson-dfinity, line 459: can this fail?
```suggestion
pub fn create_pocketic(url: &Url) -> PocketIc {
```

ericswanson-dfinity (changes requested):

src/dfx/src/commands/canister/request_status.rs
mraszyk: Due to how PocketIC call response types are defined at the moment (see [forum](https://forum.dfinity.org/t/pocketic-call-response-types/38728)), canister call errors are represented by `Err(...)` - this line, e.g., canister trapped - as well as `Ok(WasmResult::Reject(...))` - below, e.g., canister call `ic_cdk::api::call::reject`.

src/dfx/src/commands/canister/request_status.rs
mraszyk: Indeed. Because request status can only be retrieved by the same caller who submitted the update call before.

e2e/tests-dfx/call.bash
mraszyk: The principal `aaaaa-aa` is the _management canister_ identity, not the anonymous identity `2vxsx-fae`, and thus `--identity anonymous` would not make the tests pass.

CHANGELOG.md
mraszyk: Done.

e2e/tests-dfx/call.bash
ericswanson-dfinity: Okay, but that being the case, this test shows that `--impersonate` can impersonate the management canister identity

mraszyk: > this test shows that --impersonate can impersonate the management canister identity

indeed and the impersonation mechanism is the only way how to make the management canister identity be the caller

src/dfx/src/commands/canister/request_status.rs
ericswanson-dfinity: This new code path is always taken when `env.get_pocketic.is_some()`, whether or not `--impersonate` is involved. Doesn't this mean that when using PocketIC, `dfx canister request_status` will work no matter which identity is used?

I'm adding a new test in https://github.com/dfinity/sdk/pull/4047 to demonstrate this.

mraszyk: > Doesn't this mean that when using PocketIC, dfx canister request_status will work no matter which identity is used?

Correct. If we aren't comfortable with this behavior, PocketIC would need to be extended with a check for matching callers (in production, the read state request handler takes care of it, but in PocketIC this was omitted since security is not a requirement and it didn't seem important for the sake of _canister_ testing).

ericswanson-dfinity: Here's how this could apply to canister testing:
- developer writes canister
- some of developer's maintenance processes include `dfx canister call --async` and `dfx canister request-status`, but they forget to pass the same `--identity` on both
- this all works locally since they're using PocketIC, but fails on mainnet

mraszyk: To unblock this PR, I've removed changes to the `request-status` command in this [commit](https://github.com/dfinity/sdk/pull/4013/commits/fc344fd3cb844e426c5d785729b03f8a09a959e6). I'll come back to this functionality once PocketIC supports checking caller when requesting ingress status.

ericswanson-dfinity (approved):

PR #4018, opened 2024-11-26
chore: improve error message for 'dfx deploy'.

vincent-dfinity:
# Description

If users run `dfx deploy` without enough cycles, show additional messages to indicate what to do next.
```
Error explanation:
Insufficient cycles balance to create the canister.
How to resolve the error:
Please top up your cycles balance by converting ICP to cycles like below:
'dfx cycles convert --amount=0.123 --ic'.
```

Fixes # (issue)

[SDK-1867](https://dfinity.atlassian.net/browse/SDK-1867)

# How Has This Been Tested?

Tested manually.

# Checklist:

[SDK-1867]: https://dfinity.atlassian.net/browse/SDK-1867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/dfx/src/lib/diagnosis.rs
+++ b/src/dfx/src/lib/diagnosis.rs
@@ -2,6 +2,7 @@ use crate::lib::cycles_ledger_types::create_canister::CreateCanisterError;
 use crate::lib::error_code;
 use anyhow::Error as AnyhowError;
 use dfx_core::error::root_key::FetchRootKeyError;
+use dfx_core::network::provider::get_network_context;
 use ic_agent::agent::{RejectCode, RejectResponse};
 use ic_agent::AgentError;
 use ic_asset::error::{GatherAssetDescriptorsError, SyncError, UploadContentError};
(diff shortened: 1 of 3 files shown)


ericswanson-dfinity (changes requested):

src/dfx/src/lib/diagnosis.rs
```diff
@@ -262,3 +269,14 @@ fn diagnose_ledger_not_found() -> Diagnosis {
 
     (Some(explanation.to_string()), Some(suggestion.to_string()))
 }
+
+fn insufficient_cycles(err: &CreateCanisterError) -> bool {
+    matches!(err, CreateCanisterError::InsufficientFunds { balance: _ })
```
ericswanson-dfinity, line 274: Can this error come up only when `--ic` is passed?  The ticket says to display this message, for example, when `dfx deploy --ic` is called, but there's no check here.

Also, it should be possible to add an e2e test for this to cycles-ledger.bash

ericswanson-dfinity (changes requested):

src/dfx/src/lib/diagnosis.rs
vincent-dfinity: Test added at https://github.com/dfinity/sdk/pull/4018/commits/b166f7fd08a9af95838d88f2ced2806850d64b63.

I didn't check the `--ic` flag because it will fail if users set up the cycles ledger locally without topping up the cycles to the current identity, as the test showed.

ericswanson-dfinity: Let's say the user is using the cycles ledger locally, hasn't topped up, and runs into the error.  dfx displays an error message telling them to run `dfx cycles convert --amount=0.123 --ic`, which won't fix their problem, since they should not pass `--ic` (and maybe should pass a different `--network` parameter). Do you think this will be confusing, or will the user figure it out?

vincent-dfinity: How about 
```
Error explanation:
Insufficient cycles balance to create the canister.
How to resolve the error:
Please top up your cycles balance by converting ICP to cycles like below:
'dfx cycles convert --amount=0.123'.
Please add '--ic' flag if you're targeting the IC mainnet.
```

ericswanson-dfinity: You can get the network name from get_network_context(), which is guaranteed to return Ok() if this error was returned

vincent-dfinity: Fixed at https://github.com/dfinity/sdk/pull/4018/commits/44a03aa727cc73d04c68d758242c09680952d189.

ericswanson-dfinity (approved):

PR #4031, opened 2024-12-06
fix: Fix json escaping in header values

adamspofford-dfinity:


e2e/tests-dfx/assetscanister.bash
```diff
@@ -1897,6 +1897,28 @@ WARN: {
   assert_match '/somedir/upload-me.txt 1/1 \(8 bytes\) sha [0-9a-z]* \(with cache and 1 header\)'
 }
 
+@test "asset configuration via .ic-assets.json5 - respects weird characters" {
+  install_asset assetscanister
+  touch src/e2e_project_frontend/assets/thing.txt
+  cat <<'EOF' >src/e2e_project_frontend/assets/.ic-assets.json5
```
raymondk, line 1903: I suppose it's not super useful to add a json ic-assets test also since it seems the json and json5 code paths are the same.

raymondk (approved):

ericswanson-dfinity (approved):

PR #4048, opened 2024-12-20
fix: add browser warning message 

adamspofford-dfinity:
This adds a hidden div to frontend templates to display a message suggesting switching browsers when the page breaks, to solve the Safari problem in a different way from #4011. When documentation is written on browser breakage this message should be updated to link to it.


krpeacock:
I don't like the implication that Safari is an unsupported browser - that would spook me as a developer

adamspofford-dfinity:
@krpeacock I mean, it's not an implication, it's straightforwardly true. An unadorned 'try switching browsers' would just lead to bug reports about it.

krpeacock:
It's not nuanced enough. There is no reason that Safari shouldn't work in production, and a developer seeing this message in their starter template would believe that the Internet Computer is incompatible with Safari, which makes the platform a non-starter

krpeacock (changes requested):
Message is too blunt - I don't want to get in trouble for this

krpeacock (approved):

PR #4069, opened 2025-01-21
fix: correctly detect hyphenated bin artifact file path

adamspofford-dfinity:


ericswanson-dfinity (changes requested):
Please add a note to the changelog

ericswanson-dfinity (approved):

PR #4003, opened 2024-11-18
chore: update frontend templates

ghost:
# Description

This PR updates the frontend templates to a design provided by the design team, updates the default CSP so that hello world projects are operational on Safari locally, and updates dependencies on Sass to remove deprecation warnings.

Resolves:
- https://dfinity.atlassian.net/browse/SDK-1898 
- https://dfinity.atlassian.net/browse/SDK-1896

PR #4105, opened 2025-02-08
chore: update frontend templates

raymondk:
# Description

New design for the frontend templates used by `dfx new`

Fixes # SDK-1896

Replaces: #4003


ericswanson-dfinity (changes requested):
This PR carries over a defect from the original PR: it appears to copy/paste, possibly with modifications, the individual headers in the standard security policy into the .ic-assets.json5 files in the various project templates.

This defeats the purpose of the security policy settings, because it means that developers who use these templates will have to manually propagate future changes to the standard security policy headers into their projects' configuration files.

If the project templates do actually need different headers than those in the standard security policy, then this PR should introduce some description of what those differences are and the reasoning behind them.

Once that rationale is better understood, we can evaluate whether and how to make those changes in a way that is better from a DRY perspective.

PR #4109, opened 2025-02-13
chore: Misc output changes

adamspofford-dfinity:
- Removes the 'Unhardened assets: all' line
- Reorders the URLs to clarify importance
- Prints a better error message when NPM is missing
SDK-1890, SDK-1893, SDK-1899

### DIFF
--- a/src/dfx/src/commands/deploy.rs
+++ b/src/dfx/src/commands/deploy.rs
@@ -238,13 +238,14 @@ fn display_urls(env: &dyn Environment) -> DfxResult {
         let green = Style::new().green();
         if !frontend_urls.is_empty() {
             info!(log, "  Frontend canister via browser:");
-            for (name, (url1, url2)) in frontend_urls {
-                if let Some(url2) = url2 {
+            for (name, (query_url, subdomain_url)) in frontend_urls {
+                if let Some(subdomain_url) = subdomain_url {
                     info!(log, "    {}:", name);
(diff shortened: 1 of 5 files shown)


src/dfx/src/commands/deploy.rs
```diff
@@ -238,13 +238,14 @@ fn display_urls(env: &dyn Environment) -> DfxResult {
...
             info!(log, "  Frontend canister via browser:");
-            for (name, (url1, url2)) in frontend_urls {
-                if let Some(url2) = url2 {
+            for (name, (query_url, subdomain_url)) in frontend_urls {
+                if let Some(subdomain_url) = subdomain_url {
                     info!(log, "    {}:", name);
-                    info!(log, "      - {}", green.apply_to(url1));
-                    info!(log, "      - {}", green.apply_to(url2));
+                    #[rustfmt::skip]
+                    info!(log, "      - {} (Recommended)", green.apply_to(subdomain_url));
```
ericswanson-dfinity, line 245: "recommended" doesn't work on Safari, yes?

adamspofford-dfinity: Yes, hence 'legacy'. I had trimmed that down from 'legacy/safari' because 'safari' implies that the website will actually work with safari, and it won't. Not sure if there's a good way to phrase it.

ericswanson-dfinity (dismissed):

lwshang (approved):

PR #4127, opened 2025-03-03
fix: cargo-deny failure due to toolchain not installed

ericswanson-dfinity:
# Description

The license check in CI started failing today. Example: https://github.com/dfinity/sdk/actions/runs/13636726972/job/38121248105?pr=4126

I tried running the command that it asked for, but it didn't help, probably because it's running in a docker container. I tried the v2 action and configuring it to use the rust version that it's used to, but rust-toolchain.toml still overrode it.

This PR removes rust-toolchain.toml when the deny workflow runs so that the workflow can run with what it needs.

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - ens/cargo-deny-fix
   pull_request:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


lwshang (approved):

adamspofford-dfinity (approved):

krpeacock:
feels hacky, but I guess it won't do any harm

PR #4139, opened 2025-03-06
feat: Record telemetry into log file

adamspofford-dfinity:


src/dfx/src/main.rs
```diff
@@ -180,15 +180,26 @@ fn inner_main(log_level: &mut Option<i64>) -> DfxResult {
...
-        std::process::exit(255);
+        255
+    } else {
+        0
+    };
+
+    if let Err(e) = Telemetry::append_current_command_timestamped(exit_code) {
+        if log_level.unwrap_or_default() > 0 {
+            eprintln!("error appending to telemetry log: {e}")
+        }
```
ericswanson-dfinity, line 199: Should only do this if the telemetry mode is On or Local

src/dfx/src/lib/telemetry.rs
```diff
@@ -1,22 +1,35 @@
...
 use std::ffi::OsString;
+use std::fs::OpenOptions;
+use std::io::Write;
 use std::path::PathBuf;
 use std::sync::{Mutex, OnceLock};
+use std::time::Instant;
 
 static TELEMETRY: OnceLock<Mutex<Telemetry>> = OnceLock::new();
+// separate var to preserve Telemetry as Eq
+static START_TIME: OnceLock<Instant> = OnceLock::new();
```
ericswanson-dfinity, line 23: This can go as an `elapsed_ms: Option<u128>` in the Telemetry struct.  Main can calculate the elapsed time and set it by calling a `Telemetry::set_elapsed_time(elapsed)` method just like any of the other fields that we're reporting. There's no need for a OnceLock, because it will all happen on the stack in main(). Tests won't set the elapsed time, so they won't have to concern themselves with differences related to it.

src/dfx/src/lib/telemetry.rs
```diff
@@ -65,6 +79,111 @@ impl Telemetry {
...
+    pub fn append_current_command_timestamped(exit_code: i32) -> DfxResult<()> {
+        let end_time = Instant::now();
+        let elapsed = end_time.duration_since(*START_TIME.get().unwrap());
+        let telemetry_data = TELEMETRY.get().unwrap().lock().unwrap();
+        let platform = if cfg!(target_os = "linux") && std::env::var_os("WSL_DISTRO_NAME").is_some()
+        {
+            "wsl"
+        } else {
+            std::env::consts::OS
+        };
```
ericswanson-dfinity, line 108: This can go in a separate set_platform() method

ericswanson-dfinity (changes requested):

PR #4149, opened 2025-03-13
feat: instrument dfx for telemetry

adamspofford-dfinity:


src/dfx/src/lib/telemetry.rs
```diff
@@ -93,12 +107,67 @@ impl Telemetry {
...
+        with_telemetry(|telemetry| {
+            telemetry.network_type = Some(match network.r#type {
+                NetworkTypeDescriptor::Ephemeral { .. } => {
+                    if let Some(local) = &network.local_server_descriptor {
+                        match &local.scope {
+                            LocalNetworkScopeDescriptor::Project => NetworkType::ProjectLocal,
+                            LocalNetworkScopeDescriptor::Shared { .. } => NetworkType::LocalShared,
+                        }
+                    } else {
+                        NetworkType::UnknownUrl
```
ericswanson-dfinity, line 156: Can we set this up so the only way `NetworkType::UnknownUrl` comes up is if something like `--network <url>` was passed (and did not turn out to be is_ic)?
`UnknownConfigured` should mean a network config entry with `providers` (as opposed to `bind`), also that didn't turn out to be `is_ic`.

Technically, `network.r#type` doesn't divide networks into some kind of taxonomy.  This field more describes how the canister ids are managed. You could have a local network that is persistent (though I don't know why you would), or a network with `providers` that is ephemeral. The playground has its own special handling.

Here's an attempt at identification:
- network.local_server_descriptor.is_some() - this is a local network. (comes from a config entry with `bind`)
- network.r#type is Playground - it's the playground (note this will have is_ic set too)
- network.is_ic - it's mainnet
- was passed as a url to `--network` - UnknownUrl (would probably have to indicate this in create_url_based_network_descriptor())
- UnknownConfigured otherwise (came from a config entry with `providers`, and did not turn out to be mainnet)

ericswanson-dfinity (changes requested):

e2e/tests-dfx/telemetry.bash
```diff
@@ -78,8 +80,85 @@ teardown() {
...
     assert_command jq -se '.[-101:-1] | all(.command == "identity get-principal") and length == 100' "$log"
 }
+
+@test "the last replica error is collected" {
+    dfx_new_assets
+    local log wallet
+    log=$(dfx info telemetry-log-path)
+    # explicit call, known canister
+    dfx_start
+    assert_command_fail dfx canister call ryjl3-tyaaa-aaaaa-aaaba-cai name
```
ericswanson-dfinity, line 95: Would you please add a comment here that this principal corresponds to one of the well-known canisters on the allowlist?

src/canisters/frontend/ic-asset/src/error/create_chunk.rs
```diff
@@ -6,13 +6,13 @@ use thiserror::Error;
 pub enum CreateChunkError {
     /// Failed in call to create_chunk, or in waiting for response.
     #[error("Failed to create chunk: {0}")]
-    CreateChunk(AgentError),
+    CreateChunk(#[source] AgentError),
```
ericswanson-dfinity, line 9: A change like this is also good to submit as a separate PR.

ericswanson-dfinity (changes requested):

src/canisters/frontend/ic-asset/src/error/compute_evidence.rs
```diff
@@ -32,5 +32,5 @@ pub enum ComputeEvidenceError {
 
     /// Failed to list assets in the asset canister.
     #[error("Failed to list assets: {0}")]
```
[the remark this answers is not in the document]
adamspofford-dfinity: I'd rather not introduce inconsistency into error messages with drive-by changes - if they should be restyled, they can be restyled all at once.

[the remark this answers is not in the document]
ericswanson-dfinity: That's okay, and typically what I do (in a separate PR, updating all at once).  The `: {0}` can't stay with `[source]` though.

ericswanson-dfinity (approved):

PR #4152, opened 2025-03-15
feat: send telemetry data to backend

ericswanson-dfinity:
# Description

This draft PR contains two commits.

The first adds `dfx _send-telemetry`, not meant to be called except by dfx itself, that sends collected telemetry data to a backend.

The second adds a mechanism for deciding when to automatically call the above, with file locking to ensure correct operation even if many commands are executed at once.

Fixes https://dfinity.atlassian.net/browse/SDK-1980

# How Has This Been Tested?

Tested locally using https://github.com/rikonor/clickhouse-playground/tree/ens/dfx-events as the backend.

# Checklist:


src/dfx/src/commands/send_telemetry.rs
```diff
@@ -0,0 +1,20 @@
+use crate::lib::error::DfxResult;
+use crate::lib::telemetry::Telemetry;
+use clap::Parser;
+use url::Url;
+
+const DEFAULT_URL: &str = "https://sdk.telemetry.dfinity.network";
```
ericswanson-dfinity, line 6: maybe this should come from ToolConfig, which would allow tests to override it.

sesi200 (approved):

PR #4213, opened 2025-04-15
fix: always require stable compatible upgrades (no skip_consent)

### LINKED ISSUE #4183: BUG: A motoko upgrade should never have the option to ignore stable compatiblity errors.
The code here:

https://github.com/dfinity/sdk/blob/5a823d1107256d1334d41dc0f696d17f111a5676/src/dfx/src/lib/operations/canister/install_canister.rs#L115

is dangerous because it permits the user ignore any stable compatibility errors which may lead to loss of data.

This is more dangerous than ignoring the candid compatibility error, because users can always fix the candid interface with an upgrade, but once data is lost, it's gone or corrupted forever.

Is it possible to skip/ask for consent on stable compatibility warnings only but never, ever ignore stable compatibility errors?

@luc-blaeser - do you agree with proposal?

crusso:
# Description

dfx currently allows you to upgrade a Motoko canister even if it could result in stable variable data loss. 
The check for stable compatibility is even skipped when consent is not required (`skip_consent`).

This PR hopes to make passing the check mandatory, and only allow skipping of consent for warnings.

Fixes #4183 (I hope).

Unresolved issue: since dfx deploy can deploy several canisters, ideally none of them should be deployed if any of the canisters is stable incompatible. I'm not sure this PR achieves that but I don't know enough about dfx code to tell.

# How Has This Been Tested?

# Checklist:


luc-blaeser (approved):
I like this stricter rule

PR #4234, opened 2025-05-06
chore: Remove niv

adamspofford-dfinity:
This replaces our niv update process with a Bash equivalent and removes our final dependency on Nix.


ericswanson-dfinity (approved):
Thank you!

PR #4274, opened 2025-06-12
docs: add note about removed pocket-ic-proxy-port file

ghost:


vincent-dfinity:
@ericswanson-dfinity Thanks for adding the docs!

PR #4280, opened 2025-06-18
fix: Add known extensions to telemetry

adamspofford-dfinity:
Records when `dfx extension run` is used with `nns` or `sns`, unless it is a help command.


raymondk:

PR #4368, opened 2025-08-28
feat!: use cbor for asset canister state serialization

ilbertt:
# Description

Using CBOR saves space and is faster.

# How Has This Been Tested?

Same tests should still pass.

# Checklist:


src/canisters/frontend/ic-frontend-canister/src/lib.rs
```diff
@@ -11,27 +15,19 @@ fn init(args: Option<AssetCanisterArgs>) {
...
-    t.encode(&mut ser)?;
-    ser.serialize(stable::StableWriter::default())
+    save_stable_state(&stable_state).expect("failed to serialize stable state");
 }
 
 #[post_upgrade]
 fn post_upgrade(args: Option<AssetCanisterArgs>) {
-    let (stable_state,): (ic_certified_assets::StableState,) =
-        ic_cdk::storage::stable_restore().expect("failed to restore stable state");
+    let stable_state = if is_candid_stable_state() {
```
adamspofford-dfinity, line 23: Must this logic go in user code? This means that you can't update the library without modifying your post-upgrade logic. Would it be a better idea for the load function to return the unversioned struct, having already done all migration logic internally?

ilbertt: Some upgrading logic was already in the user code (e.g. `stable_save_with_capacity`) and I think it makes sense to leave the user the choice of what to do with the serializable state. In addition, for new canisters being created with the new version of the `ic-certified-assets` library, the candid check logic is not needed

PR #4377, opened 2025-09-05
chore: update docs and regenerate schema

raymondk:
Fix the docs so they match @ShuoWangNSL changes here: https://github.com/dfinity/portal/pull/6003/files#diff-fac0d15935fc55b6f6bb12607d5aba4a8eec982f1d05fe34cbfd01a8eb1f62ae

In https://github.com/dfinity/portal/pull/6059 - I'm trying to make it so that the portal is using files directly from the submodule instead of keeping a different copy.


ShuoWangNSL (approved):
Thanks a lot for updating the docs.

PR #4387, opened 2025-09-23
feat!: set canister env in cookies for html files

ilbertt:
# Description

Returns the `ic_env` cookie for all the html files. The cookie includes the root key and all the canister environment variables prefixed with `PUBLIC_`.

# How Has This Been Tested?

Tests will be added in a follow-up PR.

# Checklist:


src/canisters/frontend/ic-certified-assets/src/lib.rs
```diff
@@ -1,9 +1,12 @@
 //! This module declares canister methods expected by the assets canister client.
 pub mod asset_certification;
+mod canister_env;
+mod cookies;
 pub mod evidence;
 pub mod state_machine;
 pub mod types;
 mod url_decode;
+mod url_encode;
```
sesi200, line 9:
```suggestion
mod url_encode;
```
Nit: Those are two extremely small modules, maybe we can collate them so that it's also easier to see they use en/decode logic that is compatible between the two functions?

src/canisters/frontend/ic-certified-assets/src/canister_env.rs
```diff
@@ -0,0 +1,60 @@
...
+    let env_var_count = env_var_count();
+
+    for i in 0..env_var_count {
+        let name = env_var_name(i);
+        if name.starts_with(PUBLIC_ENV_VAR_NAME_PREFIX) {
+            let value = env_var_value(&name);
+            icp_canister_ids.insert(name, value);
+        }
+    }
+    icp_canister_ids
```
sesi200, line 59:
```suggestion
    let mut public_env_vars = HashMap::new();
    let env_var_count = env_var_count();

    for i in 0..env_var_count {
        let name = env_var_name(i);
        if name.starts_with(PUBLIC_ENV_VAR_NAME_PREFIX) {
            let value = env_var_value(&name);
            public_env_vars.insert(name, value);
        }
    }
    public_env_vars
```
not guaranteed to be canister ids

src/canisters/frontend/ic-certified-assets/src/state_machine/mod.rs
```diff
@@ -685,6 +708,9 @@ impl State {
     }
 
     pub fn commit_batch(&mut self, arg: CommitBatchArguments, now: u64) -> Result<(), String> {
+        // Reload the canister env to get the latest values
+        self.encoded_canister_env = CanisterEnv::load().to_cookie_value();
```
sesi200, line 712: Missing logic to update cookie for all .html files

ilbertt:
After merging #4392, we should be able to fix tests here

ilbertt:
> Can you also add a test that shows that PUBLIC_ env vars are added to the cookie?

Adding them in a follow-up PR

sesi200 (approved):

PR #4389, opened 2025-09-25
feat!: system context for state methods

ilbertt:
# Description

After #4387, some methods already have two parameters (time and canister env) that can be condensed into a "system context". Using a system context simplifies the tests as well.

This PR introduces the `system_context` module and refactors the state methods to accept is as the last argument.

Blocked by #4387.

# How Has This Been Tested?

Same (refactored) tests should still pass.

# Checklist:


src/canisters/frontend/ic-certified-assets/src/system_context/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+///         if let Err(msg) = s.commit_batch(arg, &system_context) {
+///             trap(&msg);
+///         }
+///         certified_data_set(s.root_hash());
+///     });
+/// }
+/// ```
+pub struct SystemContext {
+    /// Do not access this field directly, call [Self::get_canister_env] instead.
+    pub canister_env: RefCell<Option<CanisterEnv>>,
```
sesi200, line 33:
```suggestion
    canister_env: RefCell<Option<CanisterEnv>>,
```
If you add a constructor function for tests only you can make the compiler enforce this

PR #4399, opened 2025-10-07
chore: Remove flaky httpbin test

adamspofford-dfinity:


lwshang:
The original purpose of that test was to verify our retry behavior. The original URL is intentionally hardcoded to return a 500 Internal Server Error so the test can reliably confirm that the retry logic won't be endless.

Changing it to example.com would make the test meaningless, as it would simply pass on the first attempt, losing the coverage for the retry logic.

Since it's highly unlikely we'll change the `download_file` implementation, I believe the simplest solution to stop the flaky CI errors is to just remove the test. That removes the flakiness without losing coverage on a feature we don't plan to modify.

lwshang (approved):

PR #4413, opened 2025-10-28
fix: use pocketic btc init

viviveevee:
Issue reported here: https://forum.dfinity.org/t/bitcoin-api-regtest-not-accepting-cycles/59338

Use pocket-ic to init BTC canisters instead of hand-rolling our own brittle way


vincent-dfinity:
Do we need a test to cover the issue reported https://forum.dfinity.org/t/bitcoin-api-regtest-not-accepting-cycles/59338?

viviveevee:
IMO it is the responsibility of pocket-ic to make sure its setup works

PR #4423, opened 2025-11-26
feat: support for canister ID migration

mraszyk:
# Description

This PR implements support for canister ID migration in dfx: Canister ID migration can be performed using `dfx canister migrate-id` and its status can be checked out using `dfx canister migration-status`.

# How Has This Been Tested?

An end-to-end test is introduced in this PR.

# Checklist:

### DIFF
--- a/docs/cli-reference/dfx-canister.mdx
+++ b/docs/cli-reference/dfx-canister.mdx
@@ -723,13 +723,14 @@ service : {
 ## dfx canister migrate-id
 
 Use the `dfx canister migrate-id` command to perform canister ID migration
-of a canister on one subnet (called the "migrated" canister)
-to another subnet replacing the canister ID of a canister on that other subnet.
+of a canister (called the "migrated" canister) on one subnet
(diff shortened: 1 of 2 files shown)


src/dfx/src/lib/operations/canister_migration.rs
```diff
@@ -0,0 +1,169 @@
...
+            ValidationError::MigratedInsufficientCycles(Reserved) => write!(
+                f,
+                "The migrated canister does not have enough cycles for canister migration. Top up the migrated canister with the required amount of cycles."
+            ),
+            ValidationError::CallFailed { reason } => {
+                write!(f, "Internal IC error: a call failed due to {reason}")
+            }
+        }
+    }
+}
```
viviveevee: We usually do this with `thiserror`, e.g. like here: https://github.com/dfinity/sdk/blob/master/src/canisters/frontend/ic-asset/src/error/create_chunk.rs

PR #4458, opened 2026-01-12
fix(cli): handle terminals with limited color support

sasa-tomic:
### Motivation
- Solve a long-standing issue we have observed many times in non-interactive environments:
```
⠁ Reinstalling code for canister 72ch2-fiaaa-aaaar-qbsvq-cai                    
⠁ Reinstalling code for canister 72ch2-fiaaa-aaaar-qbsvq-cai                    

thread 'main' panicked at src/dfx/src/main.rs:94:18:
Failed to set stderr output color.: ColorOutOfRange
```
- Color setting was non-optional. `term` crate panics in CI environments lacking full terminal capabilities.

### Solution
- Drop `term` dependency; use `std::io::IsTerminal` for detection. The term crate's "robustness" caused this bug - it queries terminfo and refuses to set colors if the terminal reports limited capabilities. This is "correct" but brittle - users hit ColorOutOfRange in some environments.
- Respect `NO_COLOR` env var: https://no-color.org/
- Inject ANSI codes directly, removing custom wrapper module. This follows the industry standard: Modern CLI tools (cargo, rustc, git, npm, ripgrep) all use ANSI escape codes directly. They check is_terminal() + NO_COLOR and just emit ANSI codes. This is the accepted pattern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and its status can be checked out using `dfx canister migration-status`.
 The optimization functionality provided by `ic_wasm::optimize" cannot handle Wasm modules that contains 64-bit table.
 Instead of blocking the build, such optimization failure will issue a warning.
 
+### fix: prevent panic on terminals with limited color support
+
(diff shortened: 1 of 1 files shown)


tojot-dfinity:
Thank you!

I remember @viviveevee and someone else commenting on my question in the dfiniy slack, but I no longer have access to check who.

viviveevee (approved):
Please add a changelog entry, otherwise good to merge. Thanks!

PR #4524, opened 2026-04-16
feat: Deprecate dfx

adamspofford-dfinity:
Needs to tweak the CI to stop combining stderr and stdout when using assert_eq. Many tests then need to be updated because they're explicitly looking for stderr.


ggreif (approved):
Looks good to me!

lwshang (approved):

PR #4545, opened 2026-09-18
feat!: length-prefix every variable-length field in asset canister evidence

lwshang:
# Description

The encoding hashed for the evidence of a proposed batch, and for the state hash, now
length-prefixes every variable-length field — asset keys, content types, content encodings,
header names and values, the declared `sha256`, and asset content — and hashes the number of
entries in a header map. This is the framing that `crates/state-hash` uses in
[dfinity/certified-assets](https://github.com/dfinity/certified-assets). Each operation in the
encoding is self-delimiting, which makes the digest an injective function of the batch it is
computed over.

The encoding is now specified under `compute_evidence` in
[docs/design/asset-canister-interface.md](docs/design/asset-canister-interface.md), so that
tooling which verifies a proposal can reimplement it, and `ic-asset` and `ic-certified-assets`
each pin the same test vector for a fixed batch — the two implementations have to agree for
`dfx deploy --compute-evidence` to be comparable with what the asset canister computes.

The evidence of a `SetAssetContent` operation now covers `last_chunk` whether or not `chunk_ids`
is empty, matching the content that `commit_batch` stores for the same operation.

The asset canister reports `api_version` 3. `dfx deploy --by-proposal` and
`dfx deploy --compute-evidence` report an error against an asset canister that reports a lower
version, instead of computing a value that cannot be compared with it.

## Upgrading

Evidence and state hash values that this change computes differ from the values earlier versions
compute over the same assets, so **the asset canister and dfx have to be upgraded together**:

- recompute and re-verify the evidence of any batch proposed before the upgrade (as before, a
  proposed batch does not survive a canister upgrade);
- expect a new state hash for unchanged assets.

Crate version bumps (`ic-certified-assets`, `ic-asset`, `icx-asset`) are left to the release.

## Unrelated CI fixes carried here

CI was red for two reasons that predate this branch, both external. They are fixed here so this
PR can go green; happy to split them into their own PR if preferred.

- **`fix: pin vite-plugin-environment to 1.1.3`** — 1.1.4, published 2026-09-09, raises its peer
  dependency from `vite >= 2.7` to `vite >= 8.0`. The frontend templates ask for `^1.1.3` against
  `vite` ^4.3.9 (^5.4.3 for svelte), so `npm install` in a new project fails with ERESOLVE,
  `node_modules` stays empty, and the build then fails with `tsc: command not found`. This broke
  every job that deploys a frontend project: `smoke`, `ui_test`, `dfx/new`, `dfx/info`,
  `dfx/telemetry`, `dfx/playground`, `dfx/frontend`, `dfx/deploy`, and one `dfx/assetscanister`
  case.
- **`ci: download bitcoin core from bitcoincore.org`** — bitcoin.org now 404s for the
  bitcoin-core binaries, failing the `dfx/bitcoin` jobs during provisioning. The existing sha256
  check on the tarball is unchanged.

Master's last e2e run was 2026-06-23, before either regression, which is why master appears green.

# How Has This Been Tested?

- `cargo test -p ic-certified-assets -p ic-asset` — 110 + 30 tests. New cases cover header-map
  partitioning, the key/content-type and key/content-encoding boundaries, a key or content region
  run on into the operations that follow it, chunk-boundary independence, and `last_chunk`
  alongside `chunk_ids`. Each new case was checked to fail against the previous encoding, rather
  than pass vacuously.
- `evidence_of_known_batch`, pinned independently in both crates, checks that `ic-asset` and
  `ic-certified-assets` hash the same batch to the same value.
- `cargo clippy -p dfx -p ic-asset -p ic-certified-assets --all-targets`, `cargo fmt --check`,
  `shellcheck e2e/**/*.*sh` — all clean.
- e2e: the asset canister tests no longer pin project-dependent evidence values; they capture the
  value the canister computed and assert `dfx deploy --compute-evidence` produces the same one.
  Not run locally — relying on CI.
- `src/distributed/assetstorage.wasm.gz` was rebuilt with
  `./scripts/update-frontend-canister.sh --release-build`. Relying on the
  `frontend-canister-up-to-date` job to confirm the reproducible build.

# Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marc0olo (commented):
The encoding change looks right and the cross-crate vector is the right way to
pin it. One thing needs fixing before merge.

The spec says each operation is a tag "followed by its fields, in the order they
are listed in the operation". The documented `CreateAssetArguments` lists
`enable_aliasing` before `allow_raw_access`, but `hash_create_asset` hashes
`allow_raw_access` first. I reimplemented the encoding from the document alone
and checked it against the vector this PR pins:

    pinned in both crates : 984350ce...3c778ca6
    spec as written       : 1f872096...b736f48d   mismatch
    code order            : 984350ce...3c778ca6   match

The pinned batch has `enable_aliasing: Some(true)` and
`allow_raw_access: Some(false)`, so the two orders diverge on this very vector.
Since verifying tooling reimplementing from the doc is the reason it exists, I'd
swap the two `hash_opt_bool` calls in both implementations rather than document
an exception — that keeps the field-order rule uniform, and the encoding is
already breaking here so it costs nothing.

Everything else in the spec reproduced first try: tags, big-endian nat64 length
prefixes, the header-map entry count, opt/bool tagging, content-length prefix,
and `SetAssetProperties` field order. Nice document.

Also verified: both implementations are structurally identical across all nine
operations; the domain separator is applied exactly once on each of the three
entry paths, including a state-hash restart; the empty-batch e2e constant is
`sha256("ic-certified-assets v2")`; the `last_chunk` gap is real and correctly
closed; the version gate is on both entry points, and an older dfx against a new
canister fails closed at commit time. 110 + 30 tests pass locally.

Two nits:

- `ENCODING_DOMAIN` says `v2` while `api_version` is 3. Two numbers that have to
  move together, currently off by one — worth a comment tying them.
- The spec calls the encoding "an injective function of the proposed
  `commit_batch` arguments" and then says chunking doesn't affect the digest.
  Both can't hold; it's injective over the content, not the arguments. Worth
  stating precisely in a document meant for reimplementers.

+1 on splitting the three CI commits out — the encoding change deserves its own
bisect point.

marc0olo (approved):
Approving. The encoding reproduces from the document alone: I reimplemented all six
operations from `docs/design/asset-canister-interface.md` in a separate language and got
the pinned `5e8a8c1c...c8b28089` first try. The extended vector now machine-checks the
`SetAssetProperties` conversion on both sides, which was the one thing previously only
traced by hand.

Two things for afterwards, neither blocking.

**The version gate is a lower bound only.** `api_version >= 3` catches a canister too old
to produce a comparable digest, but not a future one on a later encoding — that would
still satisfy `>= 3`, dfx would report a `v2` digest, and the mismatch would surface only
when `commit_proposed_batch` rejects it, after a reviewer has already compared two numbers
that were never comparable. Since the encoding version and `api_version` move
independently by design, the fix is for the canister to report the encoding version and
for ic-asset to require equality rather than a floor. Correct as shipped; it bites at the
next encoding change. Worth a follow-up issue.

**ic-asset needs a major bump at release.** `ComputeEvidenceError` and
`PrepareSyncForProposalError` both gain variants here and neither is `#[non_exhaustive]`.
Not worth changing in this PR — none of the crate's 21 error enums is marked, so doing it
for these two alone would be arbitrary and doing it crate-wide is separate scope. Just
needs catching when versions are set.

